PR #6, opened 2015-03-30
Fake Merge - Compare master to working copy

CharliePoole:
This is set up as a merge of master into working-copy in order to view what's different in master, which might be breaking the nuget package installs.


CharliePoole:
Turns out that the "reverse merge" technique (which I read about somewhere) doesn't work if you are making changes to the nominal base.

There aren't really 12 files changed, only 5 at this point. The only change in master is the actual renaming of the assembly.

I'm deleting the PR because it's misleading. Just look at the branches directly for comparison.

PR #9, opened 2015-04-02
Ensure that the framework driver unloads the test domain after discovery...

### LINKED ISSUE #8: Driver never unloads the test assembly
Because the NUnit3FrameworkDriver code was "borrowed" from the engine, the driver never actually unloads the test AppDomain. That's done at a higher level in the engine.

This makes it impossible to rebuild a test in VS while the adapter is holding the file.

Either the driver should have an Unload method or the domain should be handled at a higher level in the adapter. It needs to be unloaded in both the discoverer and the executor.

CharliePoole:
... and execution. Fixes #8 

This fixes _one_ of the things that holds onto the test assembly and prevents building. There's still another issue to fix, related to the IsNUnit3TestAssembly method in NUnit3TestDiscoverer. I tested by replacing the body of that method with 'return true' and found that i could discover tests, execute them and rebuild. I'm not combining both fixes here because... well, I don't actually have a fix for that other one yet. :-)

### DIFF
--- a/src/NUnitTestAdapter/AssemblyRunner.cs
+++ b/src/NUnitTestAdapter/AssemblyRunner.cs
@@ -23,40 +23,40 @@ namespace NUnit.VisualStudio.TestAdapter
     /// </summary>
     public class AssemblyRunner : IDisposable
     {
-        private readonly TestLogger logger;
-        private readonly string assemblyName;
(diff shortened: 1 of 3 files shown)


rprouse:
This looks good to me, I am going to merge.

PR #12, opened 2015-04-06
Prepare for release of CTP

CharliePoole:
Please review for merge and release.

### DIFF
--- a/src/NUnitTestAdapterInstall/source.extension.vsixmanifest
+++ b/src/NUnitTestAdapterInstall/source.extension.vsixmanifest
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Vsix xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" Version="1.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2010">
-  <Identifier Id="NUnit3TestAdapter..DD956069-EF7E-4E44-9934-33C4E4F3A793">
+  <Identifier Id="NUnit3TestAdapter">
     <Name>NUnit3 Test Adapter</Name>
     <Author>Charlie Poole</Author>
     <Version>3.0.0.0</Version>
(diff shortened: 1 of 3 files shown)


src/NUnitTestAdapterInstall/source.extension.vsixmanifest
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Vsix xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" Version="1.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2010">
-  <Identifier Id="NUnit3TestAdapter..DD956069-EF7E-4E44-9934-33C4E4F3A793">
+  <Identifier Id="NUnit3TestAdapter">
```
OsirisTerje: It is pretty common to have a guid as part of the Id.  Any reason for removing it ?

CharliePoole: Primarily because it's not actually required and concatenating a name with a guid is ugly. :-) Admittedly, it's a more or less hidden ugliness and does no more harm than the guids in an msi.

However, after your comment I did some research. I couldn't find anyplace where MS actually _recommends_ using a guid, although they do generate one if you create a vsix project. 

In the vsix version 1 schema description, they merely say that the id must be unique. In the 2.0 schema, they are more specific:

"This attribute must be a unique ID for the package chosen by its author. The name should be qualified the same way CLR types are namespaced: Company.Product.Feature.Name. The Id attribute is limited to 100 characters."

I think feature and name come into play when adding extensions to the product.

Possible alternatives for us could be:
   NUnit.NUnit3TestAdapter
   NUnit.VisualStudio.NUnit3TestAdapter

We should take a moment to decide this, since it's inconvenient to change it later.

Charlie

rprouse:
Looks good to me. Ship it :+1:

CharliePoole:
@rprouse Any thoughts on the form of the id?

rprouse:
I don't really have a preference for any of the IDs, so I was going to leave it up to you. I don't think it matters much as long as we are happy with it going forward. If I had to pick, either `NUnit.NUnit3TestAdapter` or maybe `NUnit.v3.TestAdapter`

CharliePoole:
@OsirisTerje what do you think?

OsirisTerje:
I agree that making it as a simple identifier, of which I would prefer the NUnit.NUnit3TestAdapter, since that is closest to the naming we use, would be nice.  And I think we should go for that.  But, Microsoft doesn't always do as they tell, since all vsix'es I have checked have a guid there, I just hope no other code they have assume there will always be a guid within the Id. 
But I think we should try just with the name, this is also "just" a ctp.

CharliePoole:
OK, I went with "NUnit.NUnit3TestAdapter"

PR #21, opened 2015-05-09
Move demo solution to demo directory

### LINKED ISSUE #18: Separate NUnit3TestDemo from NUnit3TestAdapter solution
Although the demo is not included in the adapter solution itself, the files fall under the source for the adapter. They should be split out and the demo solution should be moved to the top level.

At a later point, we may want to put this into a separate repo.

CharliePoole:
Fixes #18. Basically, a housekeeping change.

I was not able to move the solution to the root because both the demo and the installer solution use nuget packages, which would commingle the two packages directories.


CharliePoole:
Merging this on my own, since it doesn't impact functionality.

Issue #19, opened 2015-05-09
Use core engine


CharliePoole:
Use the core engine for loading and running tests. Update packaging to include it.

PR #22, opened 2015-05-10
Update to NUnit 3.0 beta 2; fixes #20

### LINKED ISSUE #20: Upgrade NUnit to beta-2
The adapter tests currently use the beta-1 release of the framework. Upgrade to beta-2.

CharliePoole:
Upgrades the tests to use NUnit 3.0.0-beta-2 and modifies the framework driver and other files to work with the new interface. These changes are mostly temporary since #19 will replace all the files in the NUnit subdirectory with use of the core engine.

### DIFF
--- a/src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
+++ b/src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
@@ -22,12 +22,8 @@
 // ***********************************************************************
 
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Collections.Generic;
-using System.Web.UI;
-using System.Xml;
-//using NUnit.Engine.Internal;
 using NUnit.Engine.Extensibility;
 
 namespace NUnit.Engine.Drivers
@@ -39,8 +35,9 @@ namespace NUnit.Engine.Drivers
     public class NUnit3FrameworkDriver : IFrameworkDriver
     {
         private const string NUNIT_FRAMEWORK = "nunit.framework";
+        private const string LOAD_MESSAGE = "Method called without calling Load first";
 
-        public static readonly string CONTROLLER_TYPE = "NUnit.Framework.Api.FrameworkController";
+        private static readonly string CONTROLLER_TYPE = "NUnit.Framework.Api.FrameworkController";
         private static readonly string LOAD_ACTION = CONTROLLER_TYPE + "+LoadTestsAction";
         private static readonly string EXPLORE_ACTION = CONTROLLER_TYPE + "+ExploreTestsAction";
         private static readonly string COUNT_ACTION = CONTROLLER_TYPE + "+CountTestsAction";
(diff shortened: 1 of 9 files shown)


CharliePoole:
@OsirisTerje Can you review this please?

src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
```diff
@@ -141,10 +140,21 @@ public string Explore(TestFilter filter)
...
+            if (_frameworkController == null)
+                throw new InvalidOperationException(LOAD_MESSAGE);
+        }
+
         private object CreateObject(string typeName, params object[] args)
         {
             return _testDomain.CreateInstanceAndUnwrap(
-                _frameworkAssemblyName, typeName, false, 0, null, args, null, null, null );
+                NUNIT_FRAMEWORK, typeName, false, 0,
+#if !NET_4_0
```
rprouse: The project targets 3.5 and doesn't define NET_4_0, so is this necessary?

rprouse:
I have reviewed. Looks good.

src/NUnitTestAdapterTests/NUnit3TestAdapterTests.csproj
```diff
@@ -46,9 +46,9 @@
     <Reference Include="Moq">
       <HintPath>..\..\packages\Moq.4.1.1309.1617\lib\net40\Moq.dll</HintPath>
```
OsirisTerje: Moq should be updated to latest version, 4.2.1502.0911

OsirisTerje:
In the nunit3-vs-adapter.build file, update the version number to ctp 2
 <property name="package.version" value="3.0.0-ctp-1"/>

src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
```diff
@@ -51,30 +48,33 @@ public class NUnit3FrameworkDriver : IFrameworkDriver
 
         AppDomain _testDomain;
         string _testAssemblyPath;
```
OsirisTerje: This is not used in the code

src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
```diff
@@ -22,12 +22,8 @@
...
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Collections.Generic;
-using System.Web.UI;
-using System.Xml;
-//using NUnit.Engine.Internal;
 using NUnit.Engine.Extensibility;
 
 namespace NUnit.Engine.Drivers
```
OsirisTerje: This namespace doesn't match folder location.  I think it would be nice if these were matching.

src/NUnitTestAdapterTests/NUnit3TestAdapterTests.csproj
CharliePoole: Moq is actually not used at all, so I'll drop it.

src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
CharliePoole: The files in NUnit are temporary copies of existing files that actually live in the engine, so their namespaces match the location in the engine project. I put them all in one directory so I can easily delete them when I bring in a reference to the actual engine.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -84,7 +84,7 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
                 }
                 catch (TypeLoadException ex)
                 {
-                    if (ex.TypeName == NUnit3FrameworkDriver.CONTROLLER_TYPE)
+                    if (ex.TypeName == "NUnit.Framework.Api.FrameworkController")
```
OsirisTerje: I am not sure about this change.  Now you have that explicit string both there and in the NUnit3FrameworkDriver class, and they need to match.

src/NUnitTestAdapter/NUnit/NUnit3FrameworkDriver.cs
CharliePoole: Used to log the name. I commented out the logging in this copy. Fixing this line online.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
CharliePoole: It's temporary, because I didn't want to make CONTROLLER_TYPE public after Rob just made it internal. When I put in the engine, this line will no longer be needed because the Engine itself figures out what kind of exception is involved.

OsirisTerje:
Been through the code and added some comments.  See you already have answered them ☺

Nothing further I can see, so ready to merge.


/Terje

CharliePoole:
@OsirisTerje I'll update the package script the next time around. Some of your comments apply only to temporary files while others actually apply to the original engine files in the nunit project. In the case of the latter I'm keeping a list to fix them there.

You're right about the merge being tricky. I had to chain some of my branches in order to work on them at all, due to prerequisites. In the case of #18, it was an error, however. Fortunately, that's a very simple one and doesn't impact the adapter, only the test demo assembly.

Issue #17, opened 2015-05-03
Can't read app.config settings within test methods


infl3x:
Hello,
I'm attempting to use the latest CTP release of the VS adapter and am running into issues reading app settings from my app.config file in the test project.

I've created a simple test project with the following test:

```
        [Test]
        public void ShouldReadAppConfigSetting()
        {
            Assert.That(ConfigurationManager.AppSettings["TestConfigSetting"], Is.Not.Null);
        }
```

...And added the following to the app.config file for the test project:

```
  <appSettings>
      <add key="TestConfigSetting" value="omg" />
  </appSettings>
```

I have been able to get around this issue using the console runner by specifying `-domain:multiple`.

If I switch Visual Studio back to the the old v2.x test adapter, the test passes.

CharliePoole:
This is confirmed. The test adapter does not set the config file location for the test domain.

marklam:
Would a fix for this also make bindingRedirect work?

CharliePoole:
Yes, so long as it's in the config file named for the assembly being loaded.

marklam:
It's in the .exe.config for an exe using NUnitLite 3, so I think so.
Thanks.

ravensorb:
This issue seems to still exist.  Did something change recently that would cause it to no longer work?

OsirisTerje:
@ravensorb There are 2 tests in the NUnit Demo project for this, and they both are working.  
Please check https://github.com/nunit/nunit3-vs-adapter-demo and the tests in class ConfigFileTests

If you still have a fail, please raise a new issue with a small repro project

ravensorb:
Hmm, I looked at that and I don't see any test that is reading from ConfigurationManager.AppSettings  -- am I missing something?

Note -- these tests also fail in VS2019

OsirisTerje:
Did you look at https://github.com/nunit/nunit3-vs-adapter-demo/blob/master/src/csharp/ConfigFileTests.cs  ?

If it still doesn't work for you, can you please create a small repro and upload it ?   It should work "as is".

ravensorb:
Ok, I was looking in the .netcore test project. 

Few things observed
- CSharpTestDemo
  - project will not execute in VS2019 (test discovery doesn't seem to work). Adding reference to test adapter for visual studio allows it to run inside VS
  - 20 test cases fail in VS2019 (some might be expected to fail though so its hard to tell if this is a correct result)
- NUnit3CoreTestDemo
[...]

PR #23, opened 2015-05-14
Re-implement adapter using nunit.core.engine assembly

CharliePoole:
This is a major change to the adapter, replacing adhoc code for running NUnit 3.0 tests with use of the core.engine. I'll try to summarize the major changes here, but it will basically require a lot of line by line review.

* All the borrowed engine code in the NUnit subdirectory is removed.
* The core engine is referenced as a nuget package.
* AssemblyRunner.cs  is removed and all execution code is put back into NUnit3TestExecutor.cs. This made it possible to simplify test execution logic.
* A great deal of refactoring has been done.


CharliePoole:
This fixes issue #19 and will probably take care of #17 as well. The latter needs to be checked.

demo/NUnitTestDemo/NUnit3TestDemo.csproj
```diff
@@ -36,9 +36,12 @@
...
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="nunit.framework, Version=3.0.5607.22015, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
+    <Reference Include="nunit.core.engine">
+      <HintPath>..\packages\NUnit.Core.Engine.3.0.0-beta-2-dbg\lib\nunit.core.engine.dll</HintPath>
+    </Reference>
+    <Reference Include="nunit.framework, Version=3.0.5608.31473, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\packages\NUnit.3.0.0-beta-2\lib\net45\nunit.framework.dll</HintPath>
+      <HintPath>..\packages\NUnit.3.0.0-beta-2-dbg\lib\net45\nunit.framework.dll</HintPath>
```
rprouse: Shouldn't you be pulling the released NuGet packages not a local debug build?

src/NUnitTestAdapter/NUnit3TestAdapter.csproj
```diff
@@ -54,28 +54,24 @@
       <HintPath>$(MSBuildProgramFiles32)\Microsoft Visual Studio 11.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="nunit.core.engine, Version=3.0.5608.17824, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\NUnit.Core.Engine.3.0.0-beta-2-dbg\lib\nunit.core.engine.dll</HintPath>
```
rprouse: Same comment about the debug version of the NuGet package

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -81,16 +56,51 @@ public static bool IsRunningUnderIDE
...
+        protected virtual void Initialize(IMessageLogger messageLogger)
+        {
+            _verbosity = 0; // In case we throw below
+            _adapterVersion = Assembly.GetExecutingAssembly().GetName().Version.ToString();
+
+            try
+            {
+                var registry = RegistryCurrentUser.OpenRegistryCurrentUser(@"Software\nunit.org\VSAdapter");
+                UseVsKeepEngineRunning = registry.Exist("UseVsKeepEngineRunning") && (registry.Read<int>("UseVsKeepEngineRunning") == 1);
+                _shadowCopy = registry.Exist("ShadowCopy") && (registry.Read<int>("ShadowCopy") == 1);
```
rprouse: I know this code hasn't changed, but as I read it, the default is _shadowCopy = false unless the registry setting is created somewhere else that I don't see. Shouldn't we default to true?

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -81,16 +56,51 @@ public static bool IsRunningUnderIDE
...
+                _verbosity = (registry.Exist("Verbosity")) ? registry.Read<int>("Verbosity") : 0;
+            }
+            catch (Exception e)
+            {
+                messageLogger.SendMessage(TestMessageLevel.Error, e.ToString());
+                // TODO: Shouldn't we rethrow here?
```
rprouse: I would think that we wouldn't rethrow and possibly disable the adapter just because we can't read some registry settings.

src/NUnitTestAdapter/NUnitTestAdapter.cs
rprouse: Once we get to a more stable point, it would be nice to move these from the registry into Visual Studio settings if possible. It would be nice if users didn't need to create or tweak registry settings. I've done it for extensions, but not sure how it works for test adapters.

rprouse:
Other than my comments, looks good. On a side note, I am wondering if a stripped down version of TestConverter should be moved to the core engine, maybe as a base class for runner specific versions? I will need to implement something similar for the Xamarin runners. I will keep that in mind as I convert the Xamarin adapters to use the core engine.

demo/NUnitTestDemo/NUnit3TestDemo.csproj
CharliePoole: Yes... I forgot that the package files were persisted.

src/NUnitTestAdapter/NUnitTestAdapter.cs
CharliePoole: I'll remove comment and add a message that says "Default settings will be used."

src/NUnitTestAdapter/NUnitTestAdapter.cs
CharliePoole: We had a discussion about this somewhere. The decision was that shadow-copy would be false by default but that individual runners would change it as necessary. AFAIK, only the NUnit gui will need to use it.

I agree about settings. Use of the registry is inconsistent with the rest of NUnit. Note, however, that there is not curently any easy-to-use gui for editing the NUnit settings - you have to edit the xml file directly. @OsirisTerje what do you think?

Further note on settings: both the registry and the NUnit xml settings are suitable for defaults. We really need to add a way for folks to change some things on a run-by-run basis.

CharliePoole:
I'll do a commit for the commented changes after @OsirisTerje has a chance to comment as well.

Regarding TestConverter in the core engine, I'm slightly - but only slightly - doubtful. The core engine is currently a subset of the full engine. This feature would be an "extra" feature in the core engine, which sounds a bit messy. We don't (at least currently) have a use for it in the full engine and if added there we would need to add interfaces to the api assembly as well. Even messier.

I can definitely see that there is a possibility of a test and result conversion "machine" that implements the basic structure, cache, etc. and allows derived classes to implement specific conversions. Not sure where to put it though.

CharliePoole:
I updated based on @rprouse 's comments. Waiting to hear from @OsirisTerje 

There is still (again?) an issue with the nuget package installing but not working. @OsirisTerje can you look at that?

src/NUnitTestAdapter/NUnitTestAdapter.cs
OsirisTerje: The shadow copy should be false by default, and yes, we discussed that earlier with that conclusion. 
I did the registry as a an temporary solution.  It should be settings controlled through the Tools/Options in VS, but where the actual storing should be is still open.  
The runsettings-file can be extended, but very few uses it, so forcing people to add it just for the settings doesn't seem right.  VS do have a lot of settings in the registry, so that is still a viable option, the adapter is anyway part of VS.

rprouse:
I added comments to the commit by mistake so they are not showing up here. To summarize,

NUnit3VisualStudioTestAdapterAndFramework.nuspec seems unneeded to me. It includes the framework incorrectly, it should be a dependency and the dependency on the core engine should be a <file> so that it is not included as a reference in the projects.

rprouse:
@CharliePoole 

At //build, they suggested deleting the ComponentModel cache directory to cause the MEF cache to be recalculated if you have trouble with changes in your extension not appearing. It may help here.

The directory is at 

`C:\Users\{username}\AppData\Local\Microsoft\VisualStudio\14.0Exp\ComponentModelCache`

Adjust username and VS version accordingly.

CharliePoole:
In this case, the extension works as a vsix but not as a nuget package, leading me to think that the packaging is somehow messed up.

CharliePoole:
I'm happy to remove the adapter and framework package. There is no dependency of the adapter on the framework and once we start doing updates you'll be able to run tests that use various versions of the framework. In the debian world, this would be a "suggests" rather than a dependency but there's no such thing in nuget.

CharliePoole:
I removed the package we aren't using. It's confusing to have it there.

Based on our email discussions, I'm merging.

PR #30, opened 2015-05-21
Runs and discovers tests in a separate domain

### LINKED ISSUE #29: Latest TestAdapter locking dll's
After updating two of my VS2013 environments to use the latest test adapter I get errors when trying to build my solution.  Where MyDLL.dll is the asp.net MVC binary which is referenced by the the test project.

`Error	46	Unable to copy file "obj\Debug\MyDLL.dll" to "bin\MyDLL.dll". The process cannot access the file 'bin\MyDLL.dll' because it is being used by another process.	`

A 3rd environment which i havent updated to the latest V3Testadapter will clean and build successfully. 

All 3 environments have the test explorer window open and docked.

rprouse:
So that test assemblie do not remain locked. Fixes #29.

Terje, this is working except for a small bug. I am out of time and have to leave in a few minutes, so I was hoping you could take a look.

This is based on your work, I just changed the wrappers around a bit to only return serializable objects. Sorry, I changed the reference paths to VS 2015 because that is the only machine I had with me.

I can discover and run all tests or individual tests and the test assemblies do not get locked. This works with or without the keep test engine running setting.

The only issue is there seems to be a discovery pass after the test run, so the tests go grey after they run. I am not sure what I did there. If you can figure that out, then I think we are ready to go. We will want to increase the version number, but I can do that and prepare a release later tonight or in the morning.

### DIFF
--- a/src/NUnitTestAdapter/EngineWrapper.cs
+++ b/src/NUnitTestAdapter/EngineWrapper.cs
@@ -1,0 +1,140 @@
+﻿// ****************************************************************
+// Copyright (c) 2011-2015 NUnit Software. All rights reserved.
+// ****************************************************************
+
+using NUnit.Engine;
+using System;
+using System.Xml;
+
+namespace NUnit.VisualStudio.TestAdapter
+{
+    [Serializable]
+    internal class EngineWrapper : MarshalByRefObject, ITestEngine
+    {
+        [NonSerialized]
+        private ITestEngine _engine = new TestEngine();
+
+        public InternalTraceLevel InternalTraceLevel
+        {
+            get { return _engine.InternalTraceLevel; }
+            set { _engine.InternalTraceLevel = value; }
+        }
+
+        public IServiceLocator Services
+        {
+            get { return _engine.Services; }
+        }
+
+        public string WorkDirectory
+        {
+            get { return _engine.WorkDirectory; }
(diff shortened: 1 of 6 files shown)


CharliePoole:
Looks good. I'd love to understand why the runner wrapper needs to be both serializable and MBR though. It doesn't make much sense to me.

rprouse:
It probably doesn't. I admit, I was rushing a bit today, so I didn't spend
any time cleaning up.

CharliePoole:
I suspect a lot of the trouble Terje had was caused by my declaring the
TestEngine property using the class rather than the interface.

OsirisTerje:
Thanks Rob !
I think this is rather good!  

I switched to the interface pretty soon, so that was no issue.  Rob has fixed the XmlNode issue, and that seems to have been the last one in the chain.   I feared there might be more after that one, but from Rob's tests it seems those were the last ones there. 
The wrappers should  not need to be serializable, as they are MBR - i'm am very sure about that.  I would guess the issue was solved when you fixed the XmlNode with the InnerLoading methods.  

I'll take a  look at the grey-issue.  That tends to be the adapter banging out before it reaches those tests, so there might be something there. Anyway, we're very close now.

OsirisTerje:
I can work on this tomorrow morning and afternoon, if you merge this, we can proceed from there.

PR #31, opened 2015-05-21
Issue 29 :  Locking dll's, code cleanup

OsirisTerje:
I have cleaned up the code, removed the Serialization attributes, they were not needed.  Further, I split the two wrappers into separate files, and I moved the XmlNode extension method to the XmlHelper class where we have more of the same.  
I also removed the interfaces from the wrappers, and the corresponding "Not implemented" methods, as these classes don't really implement those interfaces any more.
I have no locking anymore.
I have no duplicate test discoveries, and no grey tests. 

However, I have issues running the tests that uses the mockassembly.  I am running out of time now, and I guess this might just be a local issue, because the external test on a test projects works just fine. 

The only thing then left before releasing this, is to update the version numbers.  If this PR is ok, and the tests run fine by you Rob,. then either of us can release this as a ctp3  I can do this tomorrow morning (9 hours from now).

### DIFF
--- a/src/NUnitTestAdapter/EngineWrapper.cs
+++ b/src/NUnitTestAdapter/EngineWrapper.cs
@@ -4,15 +4,13 @@
 
 using NUnit.Engine;
 using System;
-using System.Xml;
 
 namespace NUnit.VisualStudio.TestAdapter
 {
-   // [Serializable]
-    internal class EngineWrapper : MarshalByRefObject, ITestEngine
+    public class EngineWrapper : MarshalByRefObject 
     {
-       // [NonSerialized]
-        private ITestEngine _engine = new TestEngine();
+      
+        private readonly ITestEngine _engine = new TestEngine();
 
         public InternalTraceLevel InternalTraceLevel
         {
@@ -36,7 +34,7 @@ namespace NUnit.VisualStudio.TestAdapter
             _engine.Dispose();
         }
 
-        public ITestRunner GetRunner(TestPackage package)
+        public RunnerWrapper GetRunner(TestPackage package)
         {
             ITestRunner runner = _engine.GetRunner(package);
             return new RunnerWrapper(runner);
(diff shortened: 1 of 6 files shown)


rprouse:
I will review tonight when I get home if I can stay awake :)

src/NUnitTestAdapter/EngineWrapper.cs
```diff
@@ -4,15 +4,13 @@
...
-    [Serializable]
-    internal class EngineWrapper : MarshalByRefObject, ITestEngine
+    public class EngineWrapper : MarshalByRefObject 
     {
-        [NonSerialized]
-        private ITestEngine _engine = new TestEngine();
+      
+        private readonly ITestEngine _engine = new TestEngine();
 
         public InternalTraceLevel InternalTraceLevel
```
rprouse: Since we are not implementing the interface anymore, I don't think many of these methods/properties are used anymore and can also be removed. Let's not worry about it now though, I think it is more important to get a fix out. Just a note for the next release.

src/NUnitTestAdapterTests/NUnit3TestAdapterTests.csproj
```diff
@@ -38,7 +38,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.VisualStudio.TestPlatform.ObjectModel">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>$(MSBuildProgramFiles32)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
+      <HintPath>$(MSBuildProgramFiles32)\Microsoft Visual Studio 12.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
```
rprouse: You changed this to 11.0 in the other project. I guess you have 2012 and 2013 installed? Once again minor for now, but I would like to see what we can do with this. In my GitHub extension, I used a NuGet package to pull the VS assemblies in making it much more version tolerant. We should see if we can do the same here.

rprouse:
This all looks good. Despite my comments, I am going to merge. Thanks for cleaning it up, I was really rushed this afternoon and barely had time to get the pull request done.

For the mockassembly, has it been updated to use the latest release of nunit? That could be the problem, I will look when I get a chance, but it is getting late here.

Issue #33, opened 2015-05-24
Inconsistent Display Behavior in Test Explorer


BgRva:
I have noticed the following behaviour when using the latest test adapter (v3.0.2.0 installed on 22 May).  Some of the test results appear greyed out when they should not be.  For example, if I clean the solution and then rebuild, all tests are shown with the blue '!' icon.  But when I execute "Run All" from the test explorer, some tests will appear greyed out in the explorer:

![nunit3_post_runall](https://cloud.githubusercontent.com/assets/3044102/7787832/58b3a22a-01ee-11e5-8c7f-d317dd46e59c.png)

If I switch to displaying all tests by Outcome and execute "Run All", the same thing occurs.  The greyed out tests correspond to the first test project as listed in the solution explorer.  If I arrange the tests by Class or Outcome, the same test results will appear greyed out following a "Run All".

![nunit3_post_runall_by_outcome](https://cloud.githubusercontent.com/assets/3044102/7787833/74b60c1a-01ee-11e5-828c-0e3e283dcf40.png)

Additionally, when running tests the visual behaviour is different when I do a "Run All" vs "Run Selected".  For example, when I do a "Run All", for a second all the tests will update with a green and white arrow icon which will then disappear when the tests are run.

![nunit3_pre_runall](https://cloud.githubusercontent.com/assets/3044102/7787834/91983632-01ee-11e5-9b26-2531b95f0682.png)

But when I execute "Run Selected Tests", the green and white arrow icon does not appear prior to the tests being executed.

OsirisTerje:
If you look in the Output window, and select to see output from Tests.  Do you see anything there, like Exception System.NullReferenceException, Exception thrown executing tests ?

OsirisTerje:
This is a bug in the Run All implementation in the adapter. Only the first assembly found will be run, the rest will throw a NullRefException, and the results in the Test Explorer will be as you show above.

rprouse:
Fixed with PR #35

PR #35, opened 2015-05-24
Removed extra Unload which caused nullref exceptions when testing mul…

OsirisTerje:
The Unload statement removes the TestEngine, and this happens after the first assembly has been tested. 

No locking issue seen after this removal.  There is an Unload at the outer level which unloads the TestEngine when all tests are done.


rprouse:
Looks good to me. I agree with your assessment of the criticality and that we should do some testing and release CTP4 early next week.

Fixes #33.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -218,7 +218,6 @@ private void RunAssembly(string assemblyName, IFrameworkHandle frameworkHandle)
                 TestLog.SendErrorMessage("Exception thrown executing tests in " + assemblyName, ex);
             }
             _testRunner.Dispose();
-            Unload();
```
CharliePoole: Rightly removed. But since it's still called elsewhere, I want to say that the name is confusing. It seems as if we are unloading the NUnitTestAdapter, which is of course impossible. UnloadEngine maybe?

We may also want to unload the engine in a dispose implementation for the adapter.

PR #42, opened 2015-07-17
Set NumberOfTestWorkers to zero when running under the debugger. Fixes #36

### LINKED ISSUE #36: Option to set Number of Worker Threads
It has been reported that debugging tests is difficult when tests are marked as Parallelizable. We should either have an option to set the number of workers as you do in the console, or possibly run with one worker if debugging.

CharliePoole:
Basically, if the debugger is attached, we won't run in parallel. This will require temporary code changes when we want to debug parallel execution itself, but makes it simpler for the user to debug tests, which is the point.

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestExecutor.cs
+++ b/src/NUnitTestAdapter/NUnit3TestExecutor.cs
@@ -47,7 +47,8 @@ namespace NUnit.VisualStudio.TestAdapter
         public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrameworkHandle frameworkHandle)
         {
 #if LAUNCHDEBUGGER
-            Debugger.Launch();
+            if (!Debugger.IsAttached)
+                Debugger.Launch();
 #endif
             Initialize(frameworkHandle);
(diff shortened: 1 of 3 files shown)


src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -47,7 +47,8 @@ public sealed class NUnit3TestExecutor : NUnitTestAdapter, ITestExecutor, IDispo
         public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrameworkHandle frameworkHandle)
         {
 #if LAUNCHDEBUGGER
-            Debugger.Launch();
+            if (!Debugger.IsAttached)
```
OsirisTerje: The debugger may not necessarily be attached when we reach this point. The VSTest.Executionengine is normally not started, so there is nothing to attach to.  For the discoveryengine, that one is normally running after the first test run, but I am in doubt whether it should check there too.

OsirisTerje:
Looks good to me

PR #44, opened 2015-07-17
Create pseudo-results for ignored and skipped test cases for which NU…

### LINKED ISSUE #26: Ignored TestCase does not show up as Ignored
Running NUnit3TestDemo, the test TestCaseIsIgnored_Attribute shows as grey rather than with the warning symbol we use for  Ignored tests. We must not be passing any result to Visual Studio. A normal Test (not a test case) with the attribute shows up correctly.

CharliePoole:
…nit doesn't send events. Fixes #26 

As a temporary measure, we go through all the loaded NUnit tests. Any suite that is skipped or ignored will not send events for it's contained tests, so we recursively descend under that suite, create results for each case and send them to Visual Studio.

This should probably be changed in NUnit and we should then remove the workaround to avoid sending duplicate results.

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -2,7 +2,7 @@
 // Copyright (c) 2011-2015 NUnit Software. All rights reserved.
 // ****************************************************************
 
-//#define LAUNCHDEBUGGER
+#define LAUNCHDEBUGGER
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -2,7 +2,7 @@
 // Copyright (c) 2011-2015 NUnit Software. All rights reserved.
 // ****************************************************************
 
-//#define LAUNCHDEBUGGER
+#define LAUNCHDEBUGGER
```
rprouse: Shouldn't this remain commented out?

rprouse:
Other than my one comment, looks good.

CharliePoole:
Fixed the one issue online.

PR #46, opened 2015-07-20
Release 3.0 ctp 4

CharliePoole:
This is the ctp-4 release of the adapter. Since we'd like to get it out, I only plan on leaving it here for review for a few hours before merging.

I found that mock-assembly and NUnit3TestDemo were actually using the beta 2 framework. So were the tests for the adapter itself. Of the three adapter projects, two were using the beta 3 core engine and one - the tests - was using the beta 2 engine.

After I straightened things out and made everything beta-3, it stopped working. Debugging, I found that the beta-3 core engine doesn't work in the environment of the adapter. I think I know why and I'll work on it after this release. I think that what happened was that the beta 2 engine kept being used until I changed what was used by the tests... probably because the tests are built last.

For this release, all framework references are to beta 3. I had to fix some tests as a result. All engine references are still using beta 2, since beta 3 is broken. Both the vsix and the nuget package are working for me.

One lesson: keep all the package references consistent or we may not know what assembly is actually being used.

Another one: separate feature updates from changes to what version of NUnit we use in future.

### DIFF
--- a/src/NUnitTestAdapter/packages.config
+++ b/src/NUnitTestAdapter/packages.config
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit.Core.Engine" version="3.0.0-beta-3" targetFramework="net35" />
+  <package id="NUnit.Core.Engine" version="3.0.0-beta-2" targetFramework="net35" />
 </packages>
--- a/src/NUnitTestAdapterInstall/packages.config
+++ b/src/NUnitTestAdapterInstall/packages.config
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit.Core.Engine" version="3.0.0-beta-3" targetFramework="net45" />
+  <package id="NUnit.Core.Engine" version="3.0.0-beta-2" targetFramework="net45" />
 </packages>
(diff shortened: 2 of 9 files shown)


src/NUnitTestAdapter/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit.Core.Engine" version="3.0.0-beta-3" targetFramework="net35" />
+  <package id="NUnit.Core.Engine" version="3.0.0-beta-2" targetFramework="net35" />
```
rprouse: Shouldn't this be beta 3?

src/NUnitTestAdapterInstall/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit.Core.Engine" version="3.0.0-beta-3" targetFramework="net45" />
+  <package id="NUnit.Core.Engine" version="3.0.0-beta-2" targetFramework="net45" />
```
rprouse: Once again, beta 3?

src/NUnitTestAdapter/packages.config
rprouse: Sorry, didn't read through all of the comments before code-reviewing. Please disregard.

rprouse:
Looks good, I will merge.

PR #48, opened 2015-07-23
Remove temp fix no longer needed. Fixes #45

### LINKED ISSUE #45: Remove workaround for tests not sending events
The workaround created by PR #44 should be removed after issue https://github.com/nunit/nunit/issues/746 is resolved.

CharliePoole:
Tested with NUnit now that nunit/nunit#746 is resolved. The underlying problem will continue to appear when running against tests built with earlier betas, but that doesn't seem consequential enough to retain the code, especially since it would require a fair amount of extra work to avoid returning duplicate results.


rprouse:
Simple enough, merging.

PR #103, opened 2015-12-27
Moved VS2012 Assemblies to a NuGet package

### LINKED ISSUE #102: Package VS2012 assemblies as a private NuGet Package
I tried to do some work on the adapter the other day, but I can't build because I don't have VS2012 installed. In the past, I have updated the references to 2013, but that is less than ideal. 

I checked what some other projects do and they have created a NuGet package of the VS assemblies they need and uploaded it to MyGet for their use only. I have already setup a MyGet feed to use, so we just need to create the NuGet package and update the references.

Sam Harwell has packaged most of the VS assemblies as NuGet packages and Microsoft made him an MVP for it and points people to his packages, so I assume they are fine with it.

rprouse:
Fixes #102 

Allows you to build and work on the project on machines without VS2012 installed.

I added NuGet feeds for the AppVeyor artifacts in case anyone needs to test with CI builds of the engine. They are not currently used. We will need to modify the AppVeyor nuget packages to include build numbers as they are currently coming up as 3.1.0. Does the new Cake build fix this?

I also added a NuGet feed to MyGet. I've added @CharliePoole and @OsirisTerje as admins for this feed. We can upload whatever NuGet packages we need.

I didn't bother checking the NuGet package into source control since it contains binaries that are not ours. I also didn't create it with a normal nuspec file. I created it directly in [NuGet Package Explorer](https://npe.codeplex.com/). If we want to modify it in the future, it is as simple as;

- open the package in Package Explorer
- making the modifications you need
- increment the version
- Save as...
- Upload the saved package to the feed.

We probably won't do this often. If we do, we can automate the process.

@OsirisTerje sent me the assembly in the NuGet package. I double checked that it is indeed the 2012 version. I have tested in VS2105, but do not have earlier versions installed on this computer.


rprouse:
This is ready for review and merge.

OsirisTerje:
Checked, seems good to me.  
About the packages, imho they should never be checked in, so I am very fine with that.

rprouse:
I am going to merge this.

Issue #104, opened 2015-12-28
Can't overload async/await methods with NUnit Test Adapter


OsirisTerje:
See https://github.com/nunit/nunit-vs-adapter/issues/87 for details.
Same issue in V2 and V3

CharliePoole:
@OsirisTerje It looks to me as if your merge fixed this, but you never closed the issue. I'm closing it now, but please confirm that I'm doing the right thing!

PR #107, opened 2015-12-31
AsyncMethodHelper handles overloaded methods.  Fixes exception #104

OsirisTerje:
Same change as for the 2.0 adapter, see https://github.com/nunit/nunit-vs-adapter/pull/95

### DIFF
--- a/src/NUnitTestAdapter/Internal/AsyncMethodHelper.cs
+++ b/src/NUnitTestAdapter/Internal/AsyncMethodHelper.cs
@@ -24,7 +24,7 @@ namespace NUnit.VisualStudio.TestAdapter.Internal
             var definingType = targetAssembly.GetType(className);
             if (definingType == null) return null;
 
-            var method = definingType.GetMethod(methodName);
+            var method = definingType.GetMethods().Where(o => o.Name == methodName).OrderBy(o => o.GetParameters().Length).FirstOrDefault();
             if (method == null) return null;
 
             var asyncAttribute = GetAsyncStateMachineAttribute(method);
--- a/src/NUnitTestAdapter/Properties/AssemblyInfo.cs
+++ b/src/NUnitTestAdapter/Properties/AssemblyInfo.cs
@@ -21,5 +21,5 @@ using System.Runtime.InteropServices;
 [assembly: ComVisible(false)]
 
 [assembly: Guid("c0aad5e4-b486-49bc-b3e8-31e01be6fefe")]
-[assembly: AssemblyVersion("3.0.8.0")]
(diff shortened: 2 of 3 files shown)


src/NUnitTestAdapter/Internal/AsyncMethodHelper.cs
```diff
@@ -24,7 +24,7 @@ public string GetClassNameForAsyncMethod(string className, string methodName)
             var definingType = targetAssembly.GetType(className);
             if (definingType == null) return null;
 
-            var method = definingType.GetMethod(methodName);
+            var method = definingType.GetMethods().Where(o => o.Name == methodName).OrderBy(o => o.GetParameters().Length).FirstOrDefault();
```
CharliePoole: What happens in this code when the targetAssembly is compiled for .NET 2.0? My assumption has been that both the discovery and execution processes run under the targeted .NET version. If so, this would be a problem if executed in such a situation. Of course, we **shouldn't** ever get here in that case, but do we need some sort of guard clause somewhere to be safe?

CharliePoole: Actually, I'd like to know this for sure because I'm tired of avoiding Linq where it would be helpful. :-)

CharliePoole:
This looks good to me, although I think we should follow up on my inline comment.

Does this really **fix** #104 though? It seems as if we are still pointing to the wrong member in some cases. Or do we have a separate issue on that?

OsirisTerje:
It is a separate issue, #108.  This one fixes the exception so that the tests become visible in the Test Explorer again.  The issue of wrong navigation data happens for any method, regardless of async or not.

I only have #108 on this adapter, and not on the 2.0 adapter.  If it turns out it can be implemented just in the adapter we can port it back, but I believe this requires changes to NUnit itself, and what it returns us.

src/NUnitTestAdapter/Internal/AsyncMethodHelper.cs
OsirisTerje: I don't think it should be an issue, as you say we should never get in here in 2.0 since it doestn have async, but agree, need to check out this.

rprouse:
As per the v2.x adapter, this looks good to me. I agree with @OsirisTerje that the async code is fine in this case. If we decide otherwise, then we need to look at the equivalent PR in the 2.x adapter that has already been merged. How do we want to make a call on this?

rprouse:
@OsirisTerje, I think this looks good and your comment on the async code is valid. Shall we merge?

OsirisTerje:
Yep, agree

Issue #109, opened 2016-01-03
NUnit 2 tests are detected as errors


OsirisTerje:
When a project has both NUnit 2 and NUnit 3 tests, the adapter give an error message instead of a informational message.  
See #57 and https://github.com/nunit/nunit-vs-adapter/issues/84

CharliePoole:
@OsirisTerje You added a commit to change the error message to a warning. Is there a reason you didn't close the issue?

OsirisTerje:
No, closed now.

PR #110, opened 2016-01-03
Changed message level for detection of nunit 2 tests  #109

OsirisTerje:
I changed the message level from Error to Informational on two places.  It might be that the semantics of NUnitLoadError has changed from V2 adapter to V3 adapter.  In the discoverer there is an explicit check, which I see in the repro that is being triggered, but in the executor there are no such ones, so I did the change both places.

### DIFF
--- a/src/NUnitTestAdapter/TestLogger.cs
+++ b/src/NUnitTestAdapter/TestLogger.cs
@@ -48,7 +48,7 @@ namespace NUnit.VisualStudio.TestAdapter
 
         public void NUnitLoadError(string sourceAssembly)
         {
-            SendErrorMessage("NUnit failed to load " + sourceAssembly);
+            SendInformationalMessage("NUnit failed to load " + sourceAssembly);
         }
 
         public void SendErrorMessage(string message)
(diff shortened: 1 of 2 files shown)


src/NUnitTestAdapter/TestLogger.cs
```diff
@@ -48,7 +48,7 @@ public void LoadingAssemblyFailedWarning(string dependentAssembly, string source
 
         public void NUnitLoadError(string sourceAssembly)
         {
-            SendErrorMessage("NUnit failed to load " + sourceAssembly);
+            SendInformationalMessage("NUnit failed to load " + sourceAssembly);
```
CharliePoole: It seems odd that "NUnitLoadError" is not an error. Do we call it from many places? Is it sometimes an error and sometimes not? Seems like we should either rename it (if it's always used informationally) or not use it in the particular case you are trying to fix.

rprouse: The other place where this is called is in the TestExecutor in the else clause for `if (loadResult.GetAttribute("runstate") == "Runnable")`, so it does look like it should be an informational message. Should the method be renamed appropriately or at least a comment placed on the method?

CharliePoole: A NonRunnable assembly seem like at least a potential error. For example, if I rename a text file to ".dll" it will be a non-runnable assembly. It sounds like we need some further distinction based on why it's not runnable. Issue nunit/nunit#1026 is supposed to address this but in the meantime, we had planned to use the message to distinguish.

rprouse:
My comment is just a minor question. Feel free to merge or change/merge, whichever you feel is best.

Issue #133, opened 2016-03-16
"Source: no source available" for tests in abstract base class


drmannion:
Test methods defined on an abstract base class are properly included under the derived class in Test Explorer and run as expected. However, you can not open the test method source code for the base class methods from Test Explorer by double clicking or by pressing F12. Test Explorer shows the following message: "Source: no source 
![untitled picture](https://cloud.githubusercontent.com/assets/12001392/13828622/b30872de-eb7f-11e5-9caa-619d1fc663ab.png)
available"

Green-Bug:
I was about to file this same issue and I already have a fix for it. It requires a change in the engine as well.

I'm now working on adding tests for the change, I expect to have them in the next few days.

CharliePoole:
Appears to require an NUnit bug fix. Once we get the fix this just needs to be verified and closed.

CharliePoole:
Closing this as a duplicate of #163. Normally, we would close the later one and keep this open but there is already a PR refering to #163. Thanks goes to the originators of both issues.

PR #134, opened 2016-03-20
Fix for #133: use declaring class for inherited methods.

Green-Bug:


Green-Bug:
The new fix doesn't require any changes here. See nunit/nunit#1365.

PR #141, opened 2016-03-24
Add comment to ignore.addins so it's not zero length. 

### LINKED ISSUE #118: Corrupted ignore.addins file in installation
This was noted in #117 at https://github.com/nunit/nunit3-vs-adapter/issues/117#issuecomment-178706413

I'm making this a separate issue for ease of fixing it.

CharliePoole:
Fixes #118.

This is really all I can see to do on the adapter side. Anybody see anything else to do ?

@rprouse has suggested some error checking in reading the addins files, but that would be an issue for the engine rather than the adapter.

### DIFF
--- a/src/NUnitTestAdapterInstall/ignore.addins
+++ b/src/NUnitTestAdapterInstall/ignore.addins
@@ -1,0 +1,1 @@
+# Empty addins file so that the engine doesn't examine every file in the directory


rprouse:
Better than nothing.

RobertK66:
I would suggest to move the Exception Handling Block in the method NUnit3TestDiscoverer.DiscoverTests() "one level higher".  (Attaching to "try {" in line 46/47 instead of the one in line 49/50)
Other than the current behaviour (silently fail) this would give some Exception indication in the VS output folder if the call to GetRunnerFor(...) does not succeed.( At Least this was how I found the problem on my side – going the hard way to compile and debug this method….)

PR #143, opened 2016-03-28
Revert to 3.0.1 test engine

CharliePoole:
The 3.2 test engine is causing errors in the adapter, which weren't detected originally. I'm reverting to 3.0.1 for the 3.0 adapter release. We may need some fixes in the 3.2 engine before we can use it for the adapter.


rprouse:
Does the adapter use a local engine, or will it load the newest installed engine like console? If so, will these issues show up if users install 3.2?

CharliePoole:
It actually references a local engine and uses `new TestEngine()` to create it. I would like to get away from that, but we have to deal with the multiple engine issue first.

PR #146, opened 2016-03-31
Issue 118b

CharliePoole:
This makes a further change to fix issue  #118 as suggested by @RobertK66.

Attempting to get a runner for the first assembly will cause the engine to auto-initialize. An exception may be thrown at this point, possibly indicating a problem in the engine itself or an extension or a corrupted file. The change makes sure this gets reported clearly.

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -43,76 +43,69 @@ namespace NUnit.VisualStudio.TestAdapter
 
                 ITestRunner runner = null;
 
+                runner = GetRunnerFor(sourceAssembly);
+
                 try
                 {
-                    runner = GetRunnerFor(sourceAssembly);
-                    try
+                    XmlNode loadResult = runner.Load();
+
+                    // Currently, this will always be the case but it might change
+                    if (loadResult.Name == "test-run")
+                        loadResult = loadResult.FirstChild;
+
+                    if (loadResult.GetAttribute("runstate") == "Runnable")
                     {
-                        XmlNode loadResult = runner.Load();
+                        XmlNode topNode = runner.Explore(TestFilter.Empty);
 
-                        // Currently, this will always be the case but it might change
-                        if (loadResult.Name == "test-run")
-                            loadResult = loadResult.FirstChild;
-
-                        if (loadResult.GetAttribute("runstate") == "Runnable")
+                        using (var testConverter = new TestConverter(TestLog, sourceAssembly))
                         {
-                            XmlNode topNode = runner.Explore(TestFilter.Empty);
-
-                            using (var testConverter = new TestConverter(TestLog, sourceAssembly))
-                            {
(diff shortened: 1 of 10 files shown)


src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -43,76 +43,69 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
...
+                }
+                catch (FileLoadException ex)
+                {
+                    // Attempts to load an invalid assembly, or an assembly with missing dependencies
+                    TestLog.LoadingAssemblyFailedWarning(ex.FileName, sourceAssembly);
+                }
+                catch (TypeLoadException ex)
+                {
+                    if (ex.TypeName == "NUnit.Framework.Api.FrameworkController")
+                        TestLog.SendWarningMessage("   Skipping NUnit 2.x test assembly");
```
OsirisTerje: Should we really send Warning if we detect 2.X?  Or is informational enough?  And, should we do that only when verbose or always?

OsirisTerje:
I'm not really sure I grasp what is going on here - the diff is rather hard to follow, so I suggest you merge this, and we take it from there.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
CharliePoole: It's not a line I really changed, though hard to tell from the formatting. In my mind it's a warning if you thought the adapter would pick it up. Since we can't read minds, I guess it's a warning only if the other adapter is not present - but we can't figure that out easily either, and probably shouldn't.

I'd say it's a warning that we may degrade to informational in the future.

CharliePoole:
@OsirisTerje Summary from 40,000 feet...

NUnitTestDiscoverer previously had two nested try blocks. Now it has just one. Not much has really changed, but lots of text in the diffs. The role of the former outer try block is now handled by a try block in NUnitTestAdapter.GetRunner method, allowing better messaging for an error in that specific situation which arises if an extension or any other part of the engine throws an exception when it is first initialized.

RobertK66:
I tested this code from commit f105e0... with my original use case (corrupted empty ignore.addins after nuget package instalation) and it worked as intended showing error indication with the underlying problems: 

_------ Discover test started ------
NUnit Adapter 3.0.8.1 discovering tests is started
Processing C:\Develop\ws\NuGetPackages\Mvvm\TestMvvmBase\bin\Debug\TestMvvmBase.dll
NUnit Adapter 3.0.8.1 discovering tests is started
Processing C:\Develop\ws\NuGetPackages\Mvvm\MvvmBase\bin\Debug\MvvmBase.dll
Error: Unable to get runner for this assembly. Check installation, including any extensions.
ArgumentException: Illegales Zeichen im Pfad.
Error: Unable to get runner for this assembly. Check installation, including any extensions.
ArgumentException: Illegales Zeichen im Pfad.
========== Discover test finished: 0 found (0:00:25,927883) ==========_

PR #147, opened 2016-03-31
Correct URL for docs

### LINKED ISSUE #94: More Information Link in Adapter Broken
The `More Information` link in the adapter goes to the old docs site and does not work.

http://nunit.org/index.php?p=vsTestAdapter&r=3.0

It should go to either the extension page on Visual Studio Gallery, or to this repository.

CharliePoole:
This fixes #94


src/NUnitTestAdapter/NUnit3TestAdapter.csproj
```diff
@@ -59,10 +59,6 @@
       <HintPath>..\..\packages\Microsoft.VisualStudio.TestPlatform.ObjectModel.0.0.3\lib\net35\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Mono.Cecil, Version=0.9.6.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
```
rprouse: Why was this removed? Seems unrelated.

CharliePoole: It's a horrible hack, without which we can't build!

* We need mono.cecil in the same directory as the engine. We make VS put it there by way of a reference.
* BUT that reference introduces the famous error about a missing member on ExtensionAttribute, because mono.cecil defines it's own attribute. VS __sometimes__ picks the wrong one and tries to use it.
* SO after I install nuget package for the engine, I remove the reference to mono.cecil from the test adapter project. It's still copied into the directory due to the reference in the install project, so everything works. Well, I said it was horrible! :-)
* As a side issue, I also remove the references to the two agent exes, but that's because we don't need them (and the x86 exe causes a warning)

We clearly need a long term solution to this problem if other folks are going to use the engine packages but for the adapter, this is a way to get it a release out at least. The users of the adapter only get a package with what they need in tools - no references required.

CharliePoole:
BTW, I'm pretty sure that the reference is put back in when AppVeyor restores the package. And yet, if you look at the last two builds, you'll see that the first failed and the second passed. I wish I could tell you why. :-(

CharliePoole:
Ah... maybe restore does less than install does. In fact I'm sure it does.

ChrisMaddock:
Mostly unrelated - but as this link points to the docs repo...the docs are currently duplicated between nunit/docs and this repo - should the latter be removed?

CharliePoole:
The old wiki is just waiting for me to double-check that we won't lose anything when I delete it.

OsirisTerje:
Also look good enough to merge, link works.

CharliePoole:
@ChrisMaddock Old wiki is gone now, btw.

ChrisMaddock:
@CharliePoole Great - just thought I'd mention it in case it had been missed! =)

PR #151, opened 2016-04-05
Support settings in a .runsettings file

### LINKED ISSUE #49: Need a way to specify test settings
When running tests with the console runner, a number of settings can be specified, either through command line parameters, or through settings in a nunit project file. For most of these, there isn't an equivalent way to control the setting when running tests through the plugin.

One way to do this would be through the runsettings file, introduced with VS 2012. It looks like that file is intended for 3rd party extensions to be able to use. That file is at a solution level, so there may be multiple test assemblies present. In that case, it may be useful to specify some settings at either the solution level or the test assembly level (which would override a solution-level setting). Here are the settings that seem appropriate to me:
- application base
- private bin path
- process isolation
- domain
- x86
- timeout
- seed
- workers
- trace level
- shadow copy
- verbosity
- UseVsKeepEngineRunning

In particular, I'd like to enable scenarios where the assemblies containing unit tests and the assemblies being tested (which need to be referenced by the assemblies containing the tests) are output into separate parallel folders. Normally this would present a problem in loading the assemblies being tested, without copying them to over to the folder containing the tests, or some other hack. It would be much cleaner to set the application base to a higher level folder, and use the private bin path to specify the probing path (which needs to be inside the application base folder) for assembly resolution. It should be possible to specify the application base relative to the solution directory, or maybe the directory containing the runsettings file. That way, it'll work correctly regardless of the full path to the source tree.

It would also be good if settings specified in the runsettings file for the solution could be automatically used when passing the solution to the console runner. That way, settings only need to be changed in one place, which makes it easier to get consistent test behavior in both the adapter and the console runner.

CharliePoole:
This fixes #49 

The following Settings are supported under the NUnit element:

* BasePath (string)
* PrivateBinPath (string)
* InternalTraceLevel (string)
* WorkDirectory (string)
* NumberOfTestWorkers (int)
* DefaultTimeout (int)
* ShadowCopyFiles (bool)
* Verbosity (int)
* RandomSeed (int)
* UseVsKeepEngineRunning (bool)

The following requested items are __not Implemented__
* ProcessModel (currently fixed as InProcess)
* DomainUsage (currently fixed as a separate domain per test)
* TargetFramework (currently determined by VS)
* RunAsX86 (currently determined by VS)

In the current build, any registry settings are ignored. You must use the corresponding settings in the `.runsettings` file. There is code, currently conditioned out, which applies the registry settings for Verbosity, UseVsKeepEngineRunning and ShadowCopyFiles when those are not set in the file. My inclination, however, is to drop use of the registry entirely for the final 3.0 release.

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -1,0 +1,115 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+
+namespace NUnit.VisualStudio.TestAdapter
+{
+    public class AdapterSettings
+    {
+        private XmlNode _runConfiguration;
+        private XmlNode _testRunParameters;
+        private XmlNode _nunitAdapter;
+
+        #region Properties
+
+        public string MaxCpuCount
+        {
+            get { return GetInnerText(_runConfiguration, "MaxCpuCount"); }
+        }
+
+        public string ResultsDirectory
+        {
+            get { return GetInnerText(_runConfiguration, "ResultsDirectory"); }
+        }
+
+        public string TargetPlatform
+        {
+            get { return GetInnerText(_runConfiguration, "TargetPlatform"); }
+        }
+
+        public string TargetFrameworkVersion
+        {
+            get { return GetInnerText(_runConfiguration, "TargetFrameworkVersion"); }
+        }
+
+        public string TestAdapterPaths
+        {
+            get { return GetInnerText(_runConfiguration, "TestAdapterPaths"); }
+        }
+
+        private Dictionary<string, string> _testProperties;
+        public IDictionary<string, string> TestProperties
+        {
+            get
+            {
+                if (_testProperties == null)
+                {
+                    _testProperties = new Dictionary<string, string>();
+
+                    if (_testRunParameters != null)
+                    {
+                        foreach (XmlNode node in _testRunParameters.SelectNodes("Property"))
+                        {
+                            var key = node.GetAttribute("name");
+                            var value = node.GetAttribute("value");
+                            if (key != null && value != null)
+                                _testProperties.Add(key, value);
+                        }
+                    }
+                }
+
+                return _testProperties;
+            }
+        }
+
+        #endregion
+
+        #region Public Methods
+
+        public void Load(IDiscoveryContext context)
+        {
+            if (context == null)
+                throw new ArgumentNullException("context", "Load called with null context");
+
+            Load(context.RunSettings.SettingsXml);
+        }
+
+        public void Load(string settingsXml)
+        {
+            if (settingsXml == null)
+                throw new ArgumentNullException("settingsXml", "Load called with null XML string");
+            if (settingsXml == string.Empty)
+                throw new ArgumentException("settingsXml", "Load called with empty XML string");
+
+            var doc = new XmlDocument();
+            doc.LoadXml(settingsXml);
+            _runConfiguration = doc.SelectSingleNode("RunSettings/RunConfiguration");
+            _testRunParameters = doc.SelectSingleNode("RunSettings/TestRunParameters");
+            _nunitAdapter = doc.SelectSingleNode("RunSettings/NUnit");
+        }
+
+        #endregion
+
+        #region Helper Methods
+
+        private string GetInnerText(XmlNode startNode, string xpath)
+        {
+            if (startNode != null)
+            {
+                var targetNode = startNode.SelectSingleNode(xpath);
+                if (targetNode != null)
+                    return targetNode.InnerText;
+            }
+
+            return null;
+        }
+
+        #endregion
+    }
+}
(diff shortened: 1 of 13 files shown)


rprouse:
This looks good to me. Since these are the using the run settings built into Visual Studio, I assume they need to be manually setup before using them by the developer? I haven't used them, are they auto-discovered?

Anyway, just questions for me, I think we can go ahead with this. In addition, I might add a default run settings file to the Visual Studio templates so developers have an example they can fill in without having to search on the web for the format.

CharliePoole:
In VS you have to select the particular .runsettings file to use.

I'll comment further on the template issue.

OsirisTerje:
@rprouse I have a set of default runsetting templates here:  https://visualstudiogallery.msdn.microsoft.com/1cc3863b-f15f-4107-bb05-3586fd65540b  
Note that the settings have evolved, so there are differences between the different VS versions

I see InternalTraceLevel is a string, shouldn't this be an int ?

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public string MaxCpuCount
+        {
+            get { return GetInnerText(_runConfiguration, "MaxCpuCount"); }
+        }
+
+        public string ResultsDirectory
+        {
+            get { return GetInnerText(_runConfiguration, "ResultsDirectory"); }
+        }
```
OsirisTerje: This is the setting used by VSTest to tell how many cores to use for running assemblies in parallel.  Do the adapter need this?   And, how do we separate out how many cores are available for NUnit parallel runs in itself ?

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -100,37 +92,41 @@ public static bool IsRunningUnderIDE
...
-                Verbosity = 1;
-#else
-                Verbosity = (registry.Exist("Verbosity")) ? registry.Read<int>("Verbosity") : 0;
-#endif
+                Settings.Load(context);
             }
             catch (Exception e)
             {
-                messageLogger.SendMessage(TestMessageLevel.Error, "Unable to access registry. Default settings will be used");
+                messageLogger.SendMessage(TestMessageLevel.Error, "Error initializing RunSettings. Default settings will be used");
```
OsirisTerje: We should try to give some more specific information here.  Just "Error initializing" doesn't tell the user what is possibly wrong.

CharliePoole:
InternalTraceLevel is an exam in the framework passed as a string from the command line. The console runner checks the values of the strings and then passes them on to the framework. Do you think we should validate them here?

OsirisTerje:
If the console runner validates them, then it also makes sense to let the adapter validate, doesn't it ?

CharliePoole:
@OsirisTerje Yes, you're probably right. It's one less obscure NUnit exception to deal with. I'll do that.

src/NUnitTestAdapter/NUnitTestAdapter.cs
CharliePoole: I'll make this more precise, but I wanted to wait till we decide about whether to scrap the registry settings (which I prefer) If we keep them, there will be more to go wrong. Probably, this catch can be eliminated and the messaging pushed back into the AdapterSettings.Load method.

src/NUnitTestAdapter/AdapterSettings.cs
CharliePoole: So if this is set to 2, for example, will VS start two different execution processes and feed different assemblies to them? If that's so, then maybe we never have to implement parallel assembly execution in the adapter itself.

The setting we currently support is NumberOfTestWorkers, which controls the threads used by the framework. The user would need to adjust that based on the number of NUnit assemblies being run.

If you have any suggestions for how to set up a test bed for this that would be helpful.

src/NUnitTestAdapter/NUnitTestAdapter.cs
CharliePoole: I ran some tests and found it's already quite specific because the e.ToString() displays the exception that actually caused the problem.

CharliePoole:
Latest commit does more validation and creates workdirectory if necessary. MaxCpuCount is now an int.

CharliePoole:
@OsirisTerje Waiting for an OK on this.

CharliePoole:
I'm closing this on my own. :-(

anirudhsanthiar:
@CharliePoole I was wondering how many of the options documented here are supported in the adapter: https://github.com/nunit/docs/wiki/Console-Command-Line, and whether the mapping from the settings described in your first comment is documented somewhere.

OsirisTerje:
The list specified here is the most recent on what the adapter sends down to NUnit:  https://github.com/nunit/docs/wiki/Tips-And-Tricks

CharliePoole:
See code starting here: https://github.com/nunit/nunit3-vs-adapter/blob/master/src/NUnitTestAdapter/NUnitTestAdapter.cs#L147 where you can see that the following settings are used:

 * ShadowCopyFIles
 * NumberOfTestWorkers
 * DefaultTimeout
 * InternalTraceLevel
 * BasePath
 * PrivateBinPath
 * RandomSeed
 * TestProperties
 * DomainUsage (see Note below)
 * DefaultTestNamePattern
 * WorkDirectory

Someone has made domain usage usable, as part of a Microsoft change that uses domain None when running an inproc data collector. This is probably an error, since the setting is not really __generally__ usable under the adapter. It's probably harmless, however.

Othewr settings are not honored either because they don't work under the adapter or because VS is already handling the equivalent feature. As an example of the latter, VS takes care of running assemblies in parallel in multiple processes and each adapter instance only runs one assembly at a time.

The list pointed to by @OsirisTerje includes general settings that are not specific to NUnit.

anirudhsanthiar:
@OsirisTerje, @CharliePoole Thanks a lot for the links, they are really useful.

Issue #144, opened 2016-03-29
Test adapter ignores fixtures that failed to be built


yaakov-h:
If an `ITestBuilder` for a given test method throws an exception, all tests in the same fixture disappear from the Visual Studio Test Explorer. There is no warning or error shown to the user.

See nunit/nunit/issues/1381 for further details.

CharliePoole:
Thanks @yaakov-h We'll add an Error message in the output window for invalid fixtures for which tests can't be generated. I think that can be done without waiting for any improvements in how NUnit works.

CharliePoole:
We are already giving an error message in the log but only when the test is run. That seems to be the best we can do without a complete reorgainization of test discovery, which currently _only_ looks at test cases returned by NUnit, not fixtures. We should re-examine this when the NUnit fix is in.

CharliePoole:
Added a test (currently #ifed out) to the demo class in order to see this problem.

CharliePoole:
If the corresponding fix to #1381 is in NUnit 3.4, then we can get this into the 3.4 adapter release.

PR #157, opened 2016-04-14
Test of ITestBuilder that throws

CharliePoole:
A test to demonstrate issue #144, which is being postponed.


rprouse:
Looks good.

PR #173, opened 2016-06-01
Fix navigation data retrieval for methods defined in a base class

### LINKED ISSUE #163: Inheriting test methods from abstract base test fixutures emits "No source data found for..." warning.
A couple of weeks ago I ran into a scenario that can be stripped down to something like this

    abstract class BaseFixture
    {
        [NUnit.Framework.Test]
        public void RunTest()
        {
        }
    }

    class DerivedTester : BaseFixture
    {
    }
So in essence, I'm deriving a class from an abstract test fixture that has a test method. The test adapter successfully recognizes a test method in the derived classed `DerivedTester`, but I get a message in the output window, Tests channel reading

    No source data found for InheritedTests.DerivedTester.RunTest
The above code was inside a namespace called `InheritedTests`, hence the namespace prefix. This symptom has been touched upon in nunit/nunit-vs-adapter#6 and nunit/nunit#1080, albeit while using generics somehow. My code here doesn't use generics; the message appeared all the same.

CharliePoole:
Fixes #163 

This fix can be left in even after the framework itself is fixed to return the correct declaring class name for tests defined in a base class. The new code only comes into play if the navigation data can't be retrieved in the normal way.

### DIFF
--- a/src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
+++ b/src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
@@ -8,13 +8,27 @@ using System.Reflection;
 
 namespace NUnit.VisualStudio.TestAdapter.Internal
 {
-    public class AsyncMethodHelper : MarshalByRefObject
+    public class NavigationDataHelper : MarshalByRefObject
     {
         Assembly targetAssembly;
 
         public void LoadAssembly(string assemblyName)
         {
             targetAssembly = Assembly.LoadFrom(assemblyName);
+        }
+
+        public string GetDefiningClassName(string className, string methodName)
+        {
+            var type = targetAssembly.GetType(className);
+
+            if (type != null)
+            {
+                var method = type.GetMethod(methodName, BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+                if (method != null)
+                    return method.DeclaringType.FullName;
+            }
+
+            return className;
         }
 
         public string GetClassNameForAsyncMethod(string className, string methodName)
--- a/src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
+++ b/src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
@@ -23,7 +23,7 @@ namespace NUnit.VisualStudio.TestAdapter.Internal
 
             if (type != null)
             {
-                var method = type.GetMethod(methodName, BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+                var method = type.GetMethods().Where(o => o.Name == methodName).OrderBy(o => o.GetParameters().Length).FirstOrDefault();
                 if (method != null)
                     return method.DeclaringType.FullName;
             }
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -143,7 +143,7 @@ namespace NUnit.VisualStudio.TestAdapter
             var className = testNode.GetAttribute("classname");
             var methodName = testNode.GetAttribute("methodname");
             var navData = GetNavigationData(className, methodName);
-            if (navData != null)
+            if (NavigationDataIsValid(navData))
             {
                 testCase.CodeFilePath = navData.FileName;
                 testCase.LineNumber = navData.MinLineNumber;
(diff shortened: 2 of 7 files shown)


CharliePoole:
@MSK61 This is inspired by your code but somewhat simplified. Feel free to review the PR.

src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
```diff
@@ -17,6 +17,20 @@ public void LoadAssembly(string assemblyName)
...
+            if (type != null)
+            {
+                var method = type.GetMethod(methodName, BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+                if (method != null)
+                    return method.DeclaringType.FullName;
+            }
+
+            return className;
+        }
+
```
MSK61: I see no problem in duplicating the lookup code here(in GetDefiningClassName) with the old one GetClassNameForAsyncMethod as they have different enough matching criteria.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -159,14 +159,24 @@ public DiaNavigationData GetNavigationData(string className, string methodName)
         {
             if (this.DiaSession == null) return null;
 
+            // First try using the class and method names provided directly
             var navData = DiaSession.GetNavigationData(className, methodName);
```
MSK61: If `NavigationDataHelper` is not `null` and the method is defined in a base class this statement will always return `null` navigation data and later flow will proceed through if `(NavigationDataHelper != null)` anyway; I would suggest that you try with the provided class name only when `NavigationDataHelper` is `null`. I'm assuming that `NavigationDataHelper.GetDefiningClassName` call isn't so expensive anyway(unless you know otherwise).

CharliePoole: This is true so long as you are not creating the AppDomain unnecessarily and it's already created if NavigationDataHelper is non null. I thought about doing what you suggest but i was afraid the code would be hard to read. Let me see if I can change the names of some variables and properties so it is not confusing.

src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
CharliePoole: We need both methods called in this order because you could have an async method defined in a base class.

src/NUnitTestAdapterTests/NavigationDataTests.cs
```diff
@@ -47,6 +47,7 @@ public void TearDown()
         [TestCase("+GenericFixture`2", "Matches", 116, 117, 117, 118)]
         [TestCase("+GenericFixture`2+DoublyNested", "WriteBoth", 132, 133, 134, 134)]
         [TestCase("+GenericFixture`2+DoublyNested`1", "WriteAllThree", 151, 152, 153, 153)]
+        [TestCase("+DerivedClass", "EmptyMethod_ThreeLines", 160, 161, 161, 161)]
```
rprouse: What am I missing here? In the code in `NavigationTestData`, `DerivedClass` does not derive from `BaseClass`. How is `EmptyMethod_ThreeLines` being found, especially since it is named differently?

CharliePoole: Well that explains things!  

It's not being caught in the CI build or my local batch runs because we exclude the navigation tests. I've been using the test in the demo app to verify that it works.

I'll update this.

src/NUnitTestAdapter/Internal/NavigationDataHelper.cs
CharliePoole: I duplicated the code to call GetMethods and apply a selection to it - it's actually needed if there is more than one method.

src/NUnitTestAdapter/TestConverter.cs
CharliePoole: I decided not to make this efficiency improvement. If we end up replacing the use of DiaSession with Mono.Cecil, no AppDomain will be needed. if we don't we can come back to this.

src/NUnitTestAdapterTests/NavigationDataTests.cs
CharliePoole: Fixed this.

rprouse: @CharliePoole I see you updated `DerivedClass` but the method is still `EmptyMethod_ThreeLines` here and `EmptyMethodThreeLines` in `BaseClass`. That is incorrect isn't it?

CharliePoole: I have a mental block for this test I guess. Fixed.

CharliePoole:
I'm going ahead and merging at this point. There are some cleanup and efficiency steps I would take except that we may be phasing out this code in short order - if the use of Mono.Cecil to replace it works out. This will go into the 3.2 release now and I'll replace or revisit it for the 3.4 release.

PR #182, opened 2016-06-05
Cecil spike

CharliePoole:
**This is a Spike - Do Not Merge**

This branch is a demonstration of replacing the use of Diasession to find source code data for the adapter with use of Mono.Cecil. As a result of doing this, I'm able to now run adapter tests fully using nunit3-console. Up to now, we have not been able to run the navigation tests. This is a big plus.

I turned on running all tests in Appveyor for the most recent commit and they pass. I have to turn this off or we will get failures on other branches, since the adapter doesn't use an appveyor.yml file.

I found the hint of how to do this in a comment on the fixie repository and used the fixie implementation as my starting point. It's complicated by the fact that the Engine uses mono.cecil and runs in the same AppDomain as the adapter. Also, the engine doesn't use the full mono.cecil package - only the basic mono.cecil assembly. I had to add  the package to the adapter in order to have the assemblies that read symbol files.

The code here should get a review, but should not be merged. Please also give this a general design review... that is, should we go this way in the adapters. If the answer is yes, then I'll create issues and do a cleaner implementation as a separate PR.


CharliePoole:
Everybody is welcome to review but I'm specially poking @rprouse and  @OsirisTerje :-)

CharliePoole:
Offline comment from @seesharper who first got me thinking about this approach...

> The Mono.Cecil approach works against the compiled test assemblies and uses the information 
> from the latest build to locate the source for a given test.
> This means that if you move a test method around in the test file and then double click the test in 
> the test explorer without compiling, it will hit the line where the test method used to be located 
> during the last build. In my experience this has not proven to be a big problem, but it is something 
> to be aware of :)

I don't think this is a special problem. It happens in other circumstances as well if you haven't compiled.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -142,11 +139,11 @@ private TestCase MakeTestCaseFromXmlNode(XmlNode testNode)
 
             var className = testNode.GetAttribute("classname");
             var methodName = testNode.GetAttribute("methodname");
-            var navData = GetNavigationData(className, methodName);
-            if (NavigationDataIsValid(navData))
+            var navData = _navigationDataProvider.GetNavigationData(className, methodName);
```
rprouse: The code is a lot cleaner now that the getting of navigation data is pulled out of this class. :+1:

src/NUnitTestAdapter/NavigationDataProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Cecil.Rocks;
+
+namespace NUnit.VisualStudio.TestAdapter
+{
+    public class NavigationDataProvider : MarshalByRefObject
```
rprouse: It looks like you removed all of the AppDomains. Is there still one in the unchanged code that makes this need to be marshalable? I thought the advantage of Cecil for us in the Engine was that we didn't need AppDomains.

src/NUnitTestAdapterTests/NavigationDataTests.cs
```diff
@@ -41,19 +41,20 @@ public void TearDown()
...
-        [TestCase("+GenericFixture`2", "Matches", 116, 117, 117, 118)]
-        [TestCase("+GenericFixture`2+DoublyNested", "WriteBoth", 132, 133, 134, 134)]
-        [TestCase("+GenericFixture`2+DoublyNested`1", "WriteAllThree", 151, 152, 153, 153)]
-        [TestCase("+DerivedClass", "EmptyMethod_ThreeLines", 160, 161, 161, 161)]
+        [TestCase("/GenericFixture`2", "Matches", 116, 117, 117, 118)]
+        [TestCase("/GenericFixture`2/DoublyNested", "WriteBoth", 132, 133, 134, 134)]
+        [TestCase("/GenericFixture`2/DoublyNested`1", "WriteAllThree", 151, 152, 153, 153)]
+        [TestCase("/DerivedClass", "EmptyMethod_ThreeLines", 160, 161, 161, 161)]
         public void VerifyNavigationData(string suffix, string methodName, int minLineDebug, int minLineRelease, int maxLineRelease, int maxLineDebug)
         {
```
rprouse: If we go with this, we can dropped the unused maxLine params.

src/NUnitTestAdapter/NavigationDataProvider.cs
CharliePoole: It shouldn't be needed any longer.

rprouse:
I like the approach and the architecture of it. Minor comments inline.

As for the disadvantage, I don't think that is a major one.

src/NUnitTestAdapter/NavigationData.cs
```diff
@@ -0,0 +1,31 @@
+﻿// ****************************************************************
+// Copyright (c) 2016 NUnit Software. All rights reserved.
+// ****************************************************************
```
rprouse: I prefer this copyright header. Should we start using it elsewhere? I don't see the point in reproducing the license everywhere.

On an aside, thanks for this PR. The .NET Core runner needs to provide line numbers too. The .NET Core runner  has a class that does what Mono.Cecil does, but I like this architecture, so I am reusing a few of your classes and names.

CharliePoole: The copyright header is the one I used originally for the VS Adapter and has never been changed.

Note that it allows releasing the software under multiple licenses, which was my original intent. I planned to have both GPL and proprietary licenses available. As far as I remember, I changed my mind and used MIT before the initial release.

We may need to have an issue somewhere (where?) to discuss this globally. Although many projects do include the text of short licenses like BSD and MIT at the top of each file, not all of them do. So I would be OK with a shorter header. However, we need to have a clear understanding of what goes into the header (as well as the license) for each separate project we manage.

Should this perhaps be a docs issue?

seesharper:
Me and @CharliePoole discussed this approach last week and there might be a few more things to consider here. 

1. What if there is no PDB? Although not that common it is possible to compile without any debug info at all and that might break this approach. Probably not an issue.

2. Mono.Cecil might become a conflicting dependency. I just briefly saw some comments regarding not relying on AppDomains and if the TestEngine runs in the same AppDomain as the target test assembly, there is a possibility that a different version of Mono.Cecil is actually referenced from that test assembly and this might in turn break the test engine. One possible solution here is to "internalize" Mono.Cecil so that Mono.Cecil is removed as an external dependency. ILMerge is the tool for the job if this even is an issue. 

Sorry if neither of these are relevant for the test engine, but I just wanted to let you know :)

CharliePoole:
@seesharper I think the PDB requirement is a matter of documentation but we should test this out with the PDB missing.

The engine is referenced by the adapter and runs in the same AppDomain. It creates the domains or processes in which tests are run. It's the framework that could potentially conflict with the application under test, which is why we keep it free of any external libraries.

rprouse:
> What if there is no PDB? Although not that common it is possible to compile without any debug info at all and that might break this approach. Probably not an issue.

Interestingly, I am working on the same functionality right now for the .NET Core runner. For that, Microsoft supplies a class called `Microsoft.Extensions.Testing.Abstractions.SourcInformationProvider` that gets the filename and line number for a method. It is constructed with the path to the PDB file. To me, that means that the Visual Studio team expects there to be a PDB file.

> Mono.Cecil might become a conflicting dependency

I think this is our biggest concern and the reason we use so few external packages within NUnit itself. I think we could get around it by loading Mono.Cecil in a different AppDomain, but it is backwards to the way I've done that in the past. @CharliePoole what do you think about that?

CharliePoole:
@rprouse See my prior reply to @seesharper 

Current structure looks like this...

| Original AppDomain | Test AppDomain |
| --------------------------|--------------------- |
| Adapter ====> Engine ==|==> Test Assembly ====> SUT | 

So we have to worry about conflicts between the adapter and engine but not between  the engine and the tests. If the adapter-engine conflicts become a problem, we could run the adapter in a separate AppDomain as we once did, but I would be more inclined to iron-out the conflicts.

The biggest issue at the moment is that the adapter targets .NET 3.5 and the engine targets .NET 3.0. I was afraid that this might cause a conflict of versions with the cecil package, but so far it looks as if there isn't any difference between the cecil assemblies for 2.0 and 3.5.

If a problem comes up with the two .NET targets, my inclination would be to roll back the adapter to target 2.0 and add in our compatibility classes for System.Linq. We might consider making them into a separate assembly if we are going to do that.

PR #186, opened 2016-06-07
Replace DiaSession with use of Mono.Cecil for source data retrieval

### LINKED ISSUE #183: Use Mono.Cecil to retrieve source code locations
There are various complications with use of DiaSession to retrieve navigation data in the adapter. One of the greatest is that we cannot easily use the console to run our navigation data tests.

I have created a branch `cecil-spike` that demonstrates how we could use cecil for this purpose.

We need to make some decisions...

1. Whether to go this route. I think we should.

2. Whether to implement it in the adapter or as an engine service the adapter can use. There are pros and cons...

If we do it in the adapter, it wil work with the current engine and will work even better with the 3.4 engine, which will return the correct declaring classname in the xml. No further engine changes will be needed.

If we do it in the engine itself, we will need to expose a service that works like the existing NavigationDataProvider and will probably want to do some caching in the engine as well. This would make the functionality available to all runners.

CharliePoole:
This fixes #183 and is substantially based on the earlier spike.


rprouse:
Looks good to me. @OsirisTerje can you take a second look and merge if you think it is okay?

OsirisTerje:
Surely will do :-)

PR #189, opened 2016-06-09
Set default test name pattern to old default value

CharliePoole:
This PR deals with change in the default NUnit test name pattern by explicitly setting its own default.

The 19 Appveyor errors on the branch are due to the fact that this branch doesn't yet have some unrelated changes that exist in master.


CharliePoole:
The failures seem to be confusing @nunit/core-team  so I'll rebase it to get the already-merged change that makes those 19 tests pass. :-)

rprouse:
Ping when you merge and I will review.

CharliePoole:
Turns out the fix for #183 had not yet been merged... I did it, rebased and pushed -f.

Issue #180, opened 2016-06-04
Upgrade to NUnit 3.4.1


CharliePoole:
Both the engine and the framework packages should be upgraded.

CharliePoole:
NUnit 3.4 is out so we can proceed. If NUnit 3.4.1 comes out in the next few days, we can use that instead but the fix going in doesn't really impact the adapter so it doesn't matter.

PR #195, opened 2016-07-02
Upgrade to NUnit 3.4.1

CharliePoole:
This fixes issue #180

### DIFF
--- a/demo/NUnitTestDemo/NUnit3TestDemo.csproj
+++ b/demo/NUnitTestDemo/NUnit3TestDemo.csproj
@@ -36,8 +36,8 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="nunit.framework, Version=3.2.1.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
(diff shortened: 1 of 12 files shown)


bbi-yggy:
We recently upgraded to the 3.4.1 version of NUnit and the Visual Studio 2015 test adapter, and now we are periodically getting test discovery failures with an error about not being able to load the Mono.Cecil.Rocks assembly. Has anyone else seen this? I will try to track down more details on the repro steps, as it has been frustratingly inconsistent.

CharliePoole:
Would that be the NUnit 3 VS Adapter 3.4? That was the last release of the adapter. It uses the NUnit 3.4.1 engine. In terms of repro, it could make a difference whether you use the vsix or the nuget package.

bbi-yggy:
Yup, we're using the NUnit 3 Test Adapter 3.4.0. I _think_ I installed from the vsix package, but I remember messing around with it for a while and that may not be correct. Don't spend any time on our issue until I can get you solid info, mostly I wondered if there was a known problem along the lines of what we're seeing.

mariuszjamro:
I'm having the same problem when using adapter 3.4.0.0 and nunit 3.2.1 on .net 3.5. Adapter is complaining about not found Mono.Cecil.Rocks.

I noticed that it works correctly when project has target framework version set to .net 4.5.

rprouse:
@bbi-yggy and @mariuszjamro I entered issue #205 to track the issues you are seeing.

PR #197, opened 2016-07-02
Process TestParameters from .runsettings file

### LINKED ISSUE #145: Implement TestRunParameters inside .testsettings for runtime parameters
[Reference #49 ](https://github.com/nunit/nunit3-vs-adapter/issues/49#issuecomment-201333589)
[Blocked by #720](https://github.com/nunit/nunit/issues/720)

Scenario:
I am using this with Selenium tests to override which browser to use and the URL of the web application at runtime for multiple environments.  I am looking to switch to NUnit because VSTest runs tests in parallel at the assembly not the class/test(I know test hasn't been implemented) like NUnit to speed up execution time.  

It is desired to implement TestRunParameters inside the .runsettings file into the VS Adapter for runtime changes.  VSTest with Visual Studio (Update 1?) and VSTS allow for these parameters to be used.  

[MSDN](https://msdn.microsoft.com/en-us/library/jj635153.aspx).  They state inside the ForcedLegacyMode "We recommend that you consider refactoring your tests to allow you to use the newer adapter."  Here is the original blog post about the TestRunParameters: [Blog](https://blogs.msdn.microsoft.com/visualstudioalm/2015/09/04/supplying-run-time-parameters-to-tests/) .  It gives a good example on how to implement it and override the parameter with VSTest at runtime with the VSTS builds.

Example XML for .runsettings file:
```xml
<TestRunParameters>
	<Parameter name="webAppUrl" value="https://google.com" />
	<Parameter name="Username" value="username" />
	<Parameter name="Password" value="password" />
</TestRunParameters>
```

Example code for MS Unit Test Framework:
```csharp
public static string GetString(TestContext context, string property)
{
        if (context.Properties[property] == null)
        {
            throw new ArgumentException("Property does not exist or does not have a value");
        }
        else
        {
            return context.Properties[property].ToString();
        }
}
```

CharliePoole:
Fixes #145

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -2,7 +2,7 @@
 // Copyright (c) 2011-2015 NUnit Software. All rights reserved.
 // ****************************************************************
 
-//#define LAUNCHDEBUGGER
+#define LAUNCHDEBUGGER
 
 using System;
 using System.Collections.Generic;
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -2,7 +2,7 @@
 // Copyright (c) 2011-2015 NUnit Software. All rights reserved.
 // ****************************************************************
 
-#define LAUNCHDEBUGGER
+//#define LAUNCHDEBUGGER
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -2,7 +2,7 @@
 // Copyright (c) 2011-2015 NUnit Software. All rights reserved.
 // ****************************************************************
 
-//#define LAUNCHDEBUGGER
+#define LAUNCHDEBUGGER
```
rprouse: Is leaving this uncommented a mistake?

rprouse:
Other than my one comment, looks good.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
CharliePoole: Yes... and AppVeyor definitely doesn't like it!

PR #201, opened 2016-07-05
Create Cake script for  build

### LINKED ISSUE #198: Create Cake script for build

CharliePoole:
Fixes #198


CharliePoole:
Still working on this - I want to add tests using vstest.console.exe.

CharliePoole:
Ready to review... Here's what's included...

* We can build test and package on Appveyor using the script.
* Build includes both the adapter and the demo solution
* Test runs the adapter tests under both nunit3-console and vstest.console. It runs the demo (which fails by design) under vstest-console.
* Packaging produces zip, vsix and nuget packages as artifacts.

This is the first time we have had a ci build that actually ran the adapter itself.

I haven't done anything to change versioning or the  suffixes we use. This will be changed after we resolve the outstanding questions using nunit/NUnit.System.Linq#8

Once this is merged, I plan to port the hold thing immediately to the V2 adapter.

build.cake
```diff
@@ -0,0 +1,268 @@
...
+    });
+
+//////////////////////////////////////////////////////////////////////
+// TEST
+//////////////////////////////////////////////////////////////////////
+
+Task("TestAdapterUsingConsole")
+	.IsDependentOn("Build")
+	.Does(() =>
+	{
```
ChrisMaddock: I may be wrong, but from memory if we run the console as a process, we need to throw an exception based on the return code. I don't think this will break the build on a failed test as it's currently written - although not setup to pull it and try right now I'm afraid! 

Would perhaps be nice to log all the failed tests in a list like the main nunit/nunit script, given there's multiple tests suites being run. I think this will currently stop on the first failed suite?

CharliePoole: You're right. I'll put in a throw for this one. For the VSTest call, it does throw on failure.

As far as logging goes, it's less useful here because the VSTest tests don't return a meaningful return code as NUnit does - it's either 0 or 1 - and the last "test" of the three is an expected failure.

build.cake
```diff
@@ -0,0 +1,268 @@
...
+		{
+			VSTest(DEMO_BIN_DIR + "NUnit3TestDemo.dll", VSTestCustomSettings);
+		}
+		catch(Exception ex)
+		{
+			Information("\nNOTE: Demo tests failed as expected.");
+			Information("This is normally not an error.\n");
+		}
+	});
+
```
ChrisMaddock: I don't know what the purpose of the demo project is - but should there be a way that this task can fail? Can it be tied down to an expected number of failures or exception type or something, or does `VSTest()` not pass any of that information back?

build.cake
```diff
@@ -0,0 +1,268 @@
+#tool nuget:?package=NUnit.ConsoleRunner&version=3.4.1
```
ChrisMaddock: If this directive doesn't specify a version, does it default to latest? That would be one less thing to update if we could avoid it!

CharliePoole: It may do that. I'll do some reading.

ChrisMaddock:
Looks good Charlie!  Not sure if any of the things I've commented on can actually be solved, but thought they were worth mentioning.

Noticed you've put `package` dependent on `build` in this one. Is that a new plan for everything?

.gitignore
```diff
@@ -27,7 +27,12 @@ tools/Cake
 packages/
 # !packages/repositories.config
 
+# NUnit results
+TestResult.xml
 
+# Tools installed by Cake
+tools/
+!tools/packages.config
```
ChrisMaddock: Not sure this line is required - looks like you're using the bootstrapping scripts that download packages.config from Cake?

build.cake
CharliePoole: I took a look at the options here. The trouble is that I don't really know anything about VSTest - this is the first time I ever used it. There doesn't seem to be an option to capture the output so I can scan for the expected result. Anyway, the actual number of failures will change with time anyway, so the only thing I could check is that it ran to completion and produced a summary line.

I also determined experimentally that an actual error, like putting in an invalid option to the call, produces the same exception, return code and message as a test failure.

So, all in all, it seemed easiest to put in a special message at the end, allowing folks to look up a few lines if they are looking at the console. The alternative, of course, would be to take this out of the standard AppVeyor tests entirely, only running it locally by command.

build.cake
```diff
@@ -0,0 +1,268 @@
...
+
+// Test Runners
+var NUNIT3_CONSOLE = TOOLS_DIR + "NUnit.ConsoleRunner/tools/nunit3-console.exe";
+
+// Test Assemblies
+var ADAPTER_TESTS = TEST_BIN_DIR + "NUnit.VisualStudio.TestAdapter.Tests.dll";
+var DEMO_TESTS = DEMO_BIN_DIR + "NUnit3TestDemo.dll";
+
+// Packages
+var ZIP_PACKAGE = PACKAGE_DIR + "NUnit3TestAdapter-" + packageVersion + ".zip";
```
ChrisMaddock: The ZIP and VSIX are both being build as 3.5.0, I presume they should instead be taking the packageVersion string calculated in `InitialiseBuild` like the nupkg?

build.cake
ChrisMaddock: >I also determined experimentally that an actual error, like putting in an invalid option to the call, produces the same exception, return code and message as a test failure.

That was the sort of thing I was wondering, that's annoying...

.gitignore
CharliePoole: That only happens if it's not present. In the past, we have edited this file to get a new version locally. The download of packages.config is a new feature in the bootstrap - at least new since 0.6.4 - so I'm not sure what it downloads. However, it seems to me that we have to control the version of our tools that get used.

ChrisMaddock: Agreed it's good to control the version - in that case do you perhaps need to force add the packages.config to git? It doesn't seem to have got into the PR. :-)

build.cake
CharliePoole: Good point. This will all be revised once we settle some of the issues of using GitVersion, but it should be consistent for now. The other package names are set after `InitializeBuild` takes effect, so packageVersion has changed. This one, being pre-initialized gets the original default setting.

CharliePoole:
I'll go through and make some of the changes - others can wait till we resolve what we are doing with GitVersion. Are you following the discussion on nunit/NUnit.System.Linq#8?

Regarding Package and Build... yes, we had some discussion about that. It should be possible to separate packaging from building - as when we bundle several things that are built separtely - but the best default seems to be to make sure we are packaging a fresh build. Then we don't have to warn people to be sure to build first. If somebody is working on the packaging script itself, this is inconvenient, but you can always comment out the dependency locally - so long as you don't commit it by mistake.

ChrisMaddock:
Sounds good.

> Are you following the discussion on nunit/NUnit.System.Linq#8?

Yeah - but as of yet have nothing really to add. Not my area of expertise to be honest, we start version numbers at 1, and just keep going up. ;-)

build.cake
CharliePoole: I moved all the AppVeyor stuff to the start of the script, outside any task, so that the variable get set once and for all.

PR #211, opened 2016-08-01
Ensure package works on .NET 2.0 and 3.5

### LINKED ISSUE #205: Adapter fails to find Mono.Cecil when targeting .NET 3.5
Entering an issue to investigate issues reported by @bbi-yggy and @mariuszjamro in #195 

> I'm having the same problem when using adapter 3.4.0.0 and nunit 3.2.1 on .net 3.5. Adapter is complaining about not found Mono.Cecil.Rocks.

> I noticed that it works correctly when project has target framework version set to .net 4.5 - seems like a workaround.

and 

>@CharliePoole, we recently upgraded to the 3.4.1 version of NUnit and the Visual Studio 2015 test adapter, and now we are periodically getting test discovery failures with an error about not being able to load the Mono.Cecil.Rocks assembly. Has anyone else seen this? I will try to track down more details on the repro steps, as it has been frustratingly inconsistent.

CharliePoole:
This fixes #205 by ensuring that we use .NET 2.0 packages for all dependencies.


CharliePoole:
Build already tested by user - fixes #205.

PR #214, opened 2016-08-03
Release 3.4.1

CharliePoole:
**Do Not Merge**

After review, I will release version 3.4.1 from this branch and then selectively merge changes to the build script into master. Changes due to resolved issues are already in the master branch.

This release resolves two critical issues and introduces a Cake build.

This release ran into problems because the '.' character is not allowed in a nuget pre-release string. At first I tried replacing with an underscore, but that's not allowed either. Only upper and lower case letters, digits and the dash are actually permitted. We should probably aim for a future build improvement where we clean up the branch name entirely if we are going to use it. A simpler alternative would be to stop using it, since the ci prefix identifies the version uniquely anyway.


CharliePoole:
Note that PR can't be merged automatically but is irrelevant anyway since this won't be merged.

src/NUnitTestAdapterInstall/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Mono.Cecil" version="0.9.6.1" targetFramework="net45" />
+  <package id="Mono.Cecil" version="0.9.6.1" targetFramework="net35" />
```
rprouse: We'll need to be careful that this doesn't get toggled back if we ever update the version.

rprouse:
Looks good to me

CharliePoole:
Released

PR #216, opened 2016-08-07
Save and restore random seed

### LINKED ISSUE #97: Tests with dynamic/random parameters are never run
Consider the following test:

	namespace ExampleNamespace
	{
		using System;
		using System.Collections.Generic;
		using NUnit.Framework;

		[TestFixture]
		public class RandomParameterTest
		{
			static readonly Random random = new Random();

			public static IEnumerable<int> Numbers => new[] { random.Next(), random.Next() };

			[Test]
			public void TestMethod([ValueSource(nameof(Numbers))] int number)
			{
				Assert.Fail();
			}
		}
	}

This test appears in the adapter as:

Example of Test Explorer window showing tests with explicit random numbers as parameters

When I then attempt to run these tests (right-click, Run Selected Tests), neither are run at all and become grayed out.  I assume this is because it generates test names twice; once when lists the tests, and then again when attempting to run the tests.  Since a different random number is generated each time, the adapter can't find the test that was selected and so skips it.

The console runner does not have this problem:

CharliePoole:
Fixes #97 

This fix saves the initial random seed to a file upon discovery and reads it back on execution. Since there is no known writable directory other than the output directory for each assembly, it repeats the action for each assembly.  This has the benefit of simplifying the code and also allowing for variation between assemblies in the future.

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -52,7 +52,7 @@ namespace NUnit.VisualStudio.TestAdapter
 
         public string PrivateBinPath { get; private set; }
 
-        public int RandomSeed { get; private set; }
+        public int? RandomSeed { get; private set; } 
 
         #endregion
 
@@ -104,7 +104,9 @@ namespace NUnit.VisualStudio.TestAdapter
             UseVsKeepEngineRunning = GetInnerTextAsBool(nunitNode, "UseVsKeepEngineRunning");
             BasePath = GetInnerText(nunitNode, "BasePath");
             PrivateBinPath = GetInnerText(nunitNode, "PrivateBinPath");
-            RandomSeed = GetInnerTextAsInt(nunitNode, "RandomSeed", -1);
+            RandomSeed = GetInnerTextAsNullableInt(nunitNode, "RandomSeed");
+            if (RandomSeed == null)
+                RandomSeed = new Random().Next();
             
 #if SUPPORT_REGISTRY_SETTINGS
             // Legacy (CTP) registry settings override defaults
@@ -121,6 +123,18 @@ namespace NUnit.VisualStudio.TestAdapter
             // Force Verbosity to 1 under Debug
             Verbosity = 1;
 #endif
+        }
+
+        public void SaveRandomSeed(string dirname)
+        {
+            using (var writer = new StreamWriter(Path.Combine(dirname, "$RANDOM_SEED$")))
+                writer.Write(RandomSeed.Value);
+        }
+
+        public void RestoreRandomSeed(string dirname)
+        {
+            using (var reader = new StreamReader(Path.Combine(dirname, "$RANDOM_SEED$")))
+                RandomSeed = int.Parse(reader.ReadLine());
         }
 
         #endregion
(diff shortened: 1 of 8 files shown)


src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -123,6 +125,18 @@ public void Load(string settingsXml)
 #endif
         }
 
+        public void SaveRandomSeed(string dirname)
+        {
+            using (var writer = new StreamWriter(Path.Combine(dirname, "$RANDOM_SEED$")))
```
rprouse: This and the read below should probably be wrapped in a try/catch. Too much can go wrong accessing the filesystem and we don't want to crash the adapter or fail the entire test run. Better to output a log message.

CharliePoole: Good catch. I intended to take another look for error handling after coding the happy pathg but didn't... working too fast!

CharliePoole:
Added try/catch block when reading or writing file that holds the random seed. To allow easier error recovery, we don't try reading or writing the file if the runsettings file has a random seed specified. Should be good to go now.

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -123,6 +138,45 @@ public void Load(string settingsXml)
...
+            try
+            {
+                writer = new StreamWriter(Path.Combine(dirname, "$RANDOM_SEED$"));
+                writer.Write(RandomSeed.Value);
+            }
+            catch(Exception ex)
+            {
+                _logger.Error("Failed to save random seed.", ex);
+            }
+            finally
```
rprouse: A using statement probably would have been cleaner than disposing in a finally block, but good enough.

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -123,6 +138,45 @@ public void Load(string settingsXml)
...
         }
 
+        public void SaveRandomSeed(string dirname)
+        {
+            TextWriter writer = null;
+
+            try
+            {
+                writer = new StreamWriter(Path.Combine(dirname, "$RANDOM_SEED$"));
+                writer.Write(RandomSeed.Value);
```
rprouse: Do we need to worry about non-english cultures here? Should we be writing this out as invariant culture and parsing it below as invariant? Probably not since it is the same machine right?

src/NUnitTestAdapter/AdapterSettings.cs
CharliePoole: We had a using statement but we needed the catch so I made the try explicit. Using inside a try generates two try blocks.

src/NUnitTestAdapter/AdapterSettings.cs
CharliePoole: It's only for local use, so I think it's fine. We can revisit if we end up saving any more stuff to pass between discovery and execution.

src/NUnitTestAdapter/AdapterSettings.cs
rprouse: I didn't know that.

CharliePoole: Unless it's optimized away by the latest compiler... :-)

PR #217, opened 2016-08-09
Run V2 Tests under the V3 Adapter

### LINKED ISSUE #207: Run NUnit V2 tests using the NUnit 3 adapter
This will require either separate installation of the V2 driver extension or bundling it with the NUnit 3 adapter. I'd prefer to keep them separate but its a little uncertain how the extension will be found when the adapter is installed globally.

CharliePoole:
**Do Not Merge - Proof of Concept to be Discussed**

I'm creating this PR as a place to discuss issues with including the V2 framework driver with the adapter in order to allow running of V2 tests. See following comments for specific issues and status.

If merged, this will fix #207.


CharliePoole:
This doesn't currently build in CI because there are additional steps needed to make it work. I perform them manually in my local build. Specifically, three assemblies from the V2 framework driver extension have to be referenced with copy local set to true. This doesn't happen automatically because the extension uses the tools directory and not a lib directory. We either need a special script that does this or a private engine package that includes the extension, which we use exclusively for the adapter.

CharliePoole:
If we decide to go with this approach, we need to figure out how to phase it in. Would we stop supporting the V2 adapter with the 3.5 release? Should people have a choice of installing a V3-only versus a V3+V2 adapter? Should there be a setting to enable V2 tests? Or to disable them? Some messages will probably have to be changed.

One problem is that there is no clear way right now to run extensions that are installed in the solution. We could do that if we knew how to find the solution but TestExplorer doesn't seem to give us that information.

OsirisTerje:
Would be nice to pick this up again.  What is the status right now on this?  What are the blockers to proceed?

CharliePoole:
My own blocker has been twofold:

1) Waiting to know what the future of this feature is. I'm a bit dubious of having an adapter that understands both .NET and .NET Core as well as V2 and V3  of NUnit. What do you think?

2) I haven't been able to build the adapters for quite a while, since the .NET Core stuff was added. I've been limiting myself to advice and simple changes, where I can let the CI tell me if it works. This is tedious, however.

jnm2:
> I haven't been able to build the adapters for quite a while, since the .NET Core stuff was added. I've been limiting myself to advice and simple changes, where I can let the CI tell me if it works. This is tedious, however.

Why do you have this limitation? Is there something we can do better?

rprouse:
Building .NET Core requires a specific version of .NET Core to be installed. I should document that and provide links to the versions that need to be installed. I hope that .NET Core becomes easier in the future and that just installing latest solves these problems.

ChrisMaddock:
I'd like to see this happen too. 😄 

> Waiting to know what the future of this feature is. I'm a bit dubious of having an adapter that understands both .NET and .NET Core as well as V2 and V3 of NUnit. What do you think?

Sounds great to me. As a user - one adapter which runs all my NUnit tests sounds ideal. Which format to test under seems fairly clear cut too.

It would be good to have the latest fixes here available for NUnit 2 users as well - e.g. new csproj.

CharliePoole:
@OsirisTerje I just noticed this PR is still here. Any reason I shouldn't delete it?

OsirisTerje:
No, this one can go.

PR #235, opened 2016-10-11
Respect SynchronousEvents and DomainUsage in runsettings

### LINKED ISSUE #234: [MS] Allow to change SynchronousEvents and DomainUsage settings
At the moment there is no way of changing the SynchronousEvents and DomainUsage settings passed to NUnit as they are hardcoded into the adapter.
Perhaps they could be specified using the DiscoveryContext's RunSettings xml together with other settings.

drognanar:
Fixes #234


rprouse:
@drognanar, before I review, when you submit pull requests, could you add a reference to the original issue and a short description on how your changes resolve that issue? We like to have cross-references between the pull request and the issue and adding a Fixes #234 to your description will auto-close the issue once it is reviewed and merged.

src/NUnitTestAdapterTests/AdapterSettingsTests.cs
```diff
@@ -171,5 +173,22 @@ public void RandomSeedSetting()
...
+
+        [Test]
+        public void SynchronousEventsSetting()
+        {
+            _settings.Load("<RunSettings><NUnit><SynchronousEvents>True</SynchronousEvents></NUnit></RunSettings>");
+            Assert.True(_settings.SynchronousEvents);
+        }
+
+        [Test]
+        public void DomainUsageSetting()
```
rprouse: Minor point, but this would probably be better as two tests created using [TestCase(...)] attributes.

rprouse: You should probably also expand on the `[TestCase]`'s to test the other allowed values of `Default` and `Multiple`.

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -177,7 +179,15 @@ private TestPackage CreateTestPackage(string assemblyName)
...
             package.Settings[PackageSettings.ProcessModel] = "InProcess";
-            package.Settings[PackageSettings.DomainUsage] = "Single";
+
+            if (Settings.DomainUsage != null)
+            {
+                package.Settings[PackageSettings.DomainUsage] = Settings.DomainUsage;
+            }
+            else
+            {
+                package.Settings[PackageSettings.DomainUsage] = "Single";
```
rprouse: The normal default for NUnit is "Multiple". @CharliePoole should that be updated here, or was it made "Single" for a reason?

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -177,7 +179,15 @@ private TestPackage CreateTestPackage(string assemblyName)
 
             // Always run one assembly at a time in process in it's own domain
             package.Settings[PackageSettings.ProcessModel] = "InProcess";
-            package.Settings[PackageSettings.DomainUsage] = "Single";
+
+            if (Settings.DomainUsage != null)
+            {
+                package.Settings[PackageSettings.DomainUsage] = Settings.DomainUsage;
```
rprouse: Our `ConsoleOptions` in the NUnit Console project constrains this to "None", "Single" and "Multiple". We should probably constrain it to valid values here to prevent our parsing in the engine from failing. Either that, or we should wrap the `Enum.Parse` in the engine in a try/catch and default to "Default" on failure.

CharliePoole: I believe he checks in AdapterSettings but a test of a bad value would be good if we don't have one already.

src/NUnitTestAdapterTests/AdapterSettingsTests.cs
CharliePoole: Other values should not be allowed if we don't want to support them. Of course it's academic if we don't want to support None either.

CharliePoole:
@rprouse I've only been commenting on code details here and reserving my general comments to the issue itself, where I raised some fairly large issues about whether we should be doing this at all. This has all now been discussed in our meeting today and I'll follow up in my next comment.

@drognanar I edited your empty description so that it has "Fixes #234" in it. Having a reference from the issue to the PR (as you do) doesn't do anything WRT GitHub processing. Having a reference in that form in the PR causes the two to be linked so that merging the PR closes the issue.

CharliePoole:
Based on our discussion earlier this morning, @drognanar will change this to use a single setting in the .runsettings to indicate that we are running under the control of the new feature. The code to handle that should turn on synchronous events and set the domain to none. It should also set number of test workers to zero to eliminate all parallelism. Note that setting it to 1 worker would leave open the possibility that one test can run in the MTA while another runs in the STA.

In this way, all the details of setting up the environment as necessary to support the feature will be localized in the adapter. We should also add some checks in the adapter that ensure only one test assembly has been provided and that it can be loaded in the current AppDomain. The latter may not require a special check, since simply trying to complete the discovery would throw an exception.

drognanar:
I have updated the PR, such that `DomainUsage` and `SynchronousEvents` are no longer processed by `AdapterSettings`.

Instead the adapter checks for the presence of the `InProcDataCollectionRunSettings` node, which is a `TestPlatform` specific setting, that is passed to `TestPlatform` to instantiate `InProcDataCollectors` during a test run. The new commit makes the adapter run tests sequentially and disable the `AppDomain` when such mode is detected.

Furthermore. I added checks that only one assembly is passed to the adapter at a time when the `TestPlatform` expects to be running with `InProcDataCollectors`.

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -146,6 +146,8 @@ private TestPackage CreateTestPackage(string assemblyName)
                     package.Settings[PackageSettings.NumberOfTestWorkers] = workers;
             }
 
+            package.Settings[PackageSettings.SynchronousEvents] = Settings.SynchronousEvents;
```
CharliePoole: I'd test Settings.InProcDataCollectorsAvailable here and set all three package values, giving a warning if the user has provided any value for NumberOfTestWorkers other than 0 - i.e. it must be either zero or not set.

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -136,6 +142,16 @@ public void Load(string settingsXml)
             // Force Verbosity to 1 under Debug
             Verbosity = 1;
 #endif
+
+            // If any in proc data collector will be instantiated by the TestPlatform run tests sequentially.
+            var inProcDataCollectorNode = doc.SelectSingleNode("RunSettings/InProcDataCollectionRunSettings/InProcDataCollectors");
+            InProcDataCollectorsAvailable = inProcDataCollectorNode != null && inProcDataCollectorNode.SelectNodes("InProcDataCollector").Count > 0;
+            if (InProcDataCollectorsAvailable)
```
CharliePoole: If we remove the two settings, that leaves NumberOfTestWorkers, which may have been set in conflict with InProcDataCollectorsAvailable. I wouldn't override that conflict here but would lean toward overriding it in the package settings with a warning message.

ManishJayaswal:
@CharliePoole can you point us to the myget feed with dev suffix that would have this change?

CharliePoole:
@ManishJayaswal This PR has not yet been merged, so it's not in our myget feed at all.

We were waiting for our 3.5 release to drop first. It has dropped but a critical error was discovered that calls for a 3.5.1 release. If it doesn't cause you a problem, I'd prefer to wait a few days for that so as to avoid having to cherry-pick changes for it.

CharliePoole:
Note that the successful ci build does give you the artifacts as well. See https://ci.appveyor.com/project/CharliePoole/nunit3-vs-adapter/build/3.5.0-ci-00261-pr-235/artifacts

ManishJayaswal:
Thanks @CharliePoole !
We can wait a couple of days.

rprouse:
@ManishJayaswal AppVeyor is producing a Nuget, so I updated out AppVeyor feed to use the URI https://ci.appveyor.com/nuget/nunit3-vs-adapter. I haven't tested it yet and it may require another push/build.

Artifacts are also available for every PR build if you walk out to AppVeyor from the PR checks. The latest build artifacts are at https://ci.appveyor.com/project/CharliePoole/nunit3-vs-adapter/build/3.5.0-ci-00261-pr-235/artifacts

CharliePoole:
Merged manually to master

PR #238, opened 2016-10-14
Just warn upon trying to restore random seed.

dwellna:
I had an issue with failures to restore the random seed where the test run always failed although all tests passed.

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -168,7 +168,7 @@ namespace NUnit.VisualStudio.TestAdapter
             }
             catch(Exception ex)
             {
-                _logger.Error("Unable to restore random seed.", ex);
+                _logger.Warning("Unable to restore random seed.", ex);
             }
             finally
             {


CharliePoole:
What happens as a result of using Error here? Other than displaying "Error" in the message that is.

CharliePoole:
Ah... I re-read your description... under what environment do you run the test and see a failure?

Why all the questions? We're trying to figure out something about how VS handles error messages. Since it's undocumented info, experimentation is all we have. :-(

dwellna:
Hi Charlie, we are using TFS 2015 and for running the tests I use the Visual Studio Test runner.
Hope, I could help. If you got any additional questions, don't hesitate to ask.
Best regards,
David
Am 14.10.2016 17:49 schrieb CharliePoole <notifications@github.com>:Ah... I re-read your description... under what environment do you run the test and see a failure?

Why all the questions? We're trying to figure out something about how VS handles error messages. Since it's undocumented info, experimentation is all we have. :-(

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/nunit/nunit3-vs-adapter","title":"nunit/nunit3-vs-adapter","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/nunit/nunit3-vs-adapter"}},"updates":{"snippets":[{"icon":"PERSON","message":"@CharliePoole in #238: Ah... I re-read your description... under what environment do you run the test and see a failure?\r\n\r\nWhy all the questions? We're trying to figure out something about how VS handles error messages. Since it's undocumented info, experimentation is all we have. :-("}],"action":{"name":"View Pull Request","url":"https://github.com/nunit/nunit3-vs-adapter/pull/238#issuecomment-253839976"}}}

CharliePoole:
This seems to confirm that simply issuing a message at the Error level causes the TFS client runner to fail the run! We have run into that in some other areas and it has been confusing because the error reporters didn't always say whether they were using TFS or the IDE. 

I think we'll fix this one and simultaneously scan for any other use of Error level messages so we can change any from which we successfully recover to warnings.

dwellna:
Ok, sounds great.
I just encountered another one when unloading the AppDomain times out.
There are several places where this is logged as error, especially in the test adapter catching the NunitEngineException which as I read shall not lead to a test run failure, right?

Although I am still trying to figure out why the AppDomain unload times out at all as I am pretty sure I stopped all threads started (in managed code) during the test run...

PR #246, opened 2016-10-22
Modify nuget package to be a tool

### LINKED ISSUE #221: Change adapter package to use tools directory
Use of the lib directory creates unneeded references in the user project. At times, they cause problems, particularly in the case of the Mono.Cecil reference. For example, the user may want to reference a different version of that assembly than we use. In addition, this reference may give rise to duplicate definitions of System.Runtime.CompilerServices.ExtensionAttribute.

When the adapter was originally developed, use of the lib directory was required. However, it appears that this has not been true since VS2012. Too bad this stuff isn't documented anywhere!

### LINKED ISSUE #243: Nuget package and copy local
The Nuget package of version 3.4.0 did set the copylocal property of the testadapter to false.
This is no longer the case with version 3.4.0, although the nuget/install.ps1 isn't changed. Is this something that was missed when implementing Cake?

CharliePoole:
Fixes #221 

This changes the nuget package so that it uses the tools directory rather than the lib directory. This means that no references are created in the project and the install.ps1 is no longer needed to prevent copying the adapter assemblies to the output directory.

This should also fix #243


rprouse:
Install.ps1 was removed from the nuspec, should it be removed from the repo too? Otherwise good.

ChrisMaddock:
This solves the issue, but do you changing moving the dll location it might be a breaking change for anyone manually copying the adapter from the nuget package/referencing its location?

This is undoubtedly the more 'correct' way to package, but as an existing package, I'm not sure how much we should think about backwards compatibility. Another option would be to overcome the references problem with the nuspec <references> tag, which would allow us to keep the current dll location.

Not sure which is the best solution, just wanted to raise the point! 🙂

CharliePoole:
@ChrisMaddock good point. Let me think about it. What do you mean by using the nuspec references tag to overcome the problem?

ChrisMaddock:
https://docs.nuget.org/ndocs/schema/nuspec#specifying-explicit-assembly-references

Nuspec's can have a `<references>` tag, which allows you to specify explicitly which items in your `/lib/` directory should be added as project references. If it's not specified (like our nuspec), everything in `/lib/` is added as a reference.

I haven't tested it, but I presume adding an empty `<references />` would work as expected - and mean we could keep everything in`lib`, but not have it add project references.

CharliePoole:
I was assuming that that the internal structure of the package is not something that people should be relying on. However, let me try out this trick and see if it will work.

CharliePoole:
It turns out that an empty `<references/>` element works the same as no element at all: all the assemblies in `lib` are referenced. :frowning_face: 

That gets us back to the original plan. Merging.

ChrisMaddock:
Shame...sorry for the false lead!

rprouse:
In which scenarios would people need paths to the adapter? There is no
executable to run and VS finds it. Is it needed for some CI builds?

On Oct 22, 2016 7:14 AM, "Chris Maddock" <notifications@github.com> wrote:

This solves the issue, but do you changing moving the dll location it might
be a breaking change for anyone manually copying the adapter from the nuget
package/referencing its location?

This is undoubtedly the more 'correct' way to package, but as an existing
package, I'm not sure how much we should think about backwards
compatibility. Another option would be to overcome the references problem
with the nuspec tag, which would allow us to keep the current dll location.

Not sure which is the best solution, just wanted to raise the point! 🙂

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/nunit/nunit3-vs-adapter/pull/246#issuecomment-255522351>,
or mute the thread

CharliePoole:
I think it may impact scenarios where you run under vstest-console.exe with the nuget package pointed to by a command-line option. We can document this in breaking changes.

ChrisMaddock:
> Is it needed for some CI builds?

Yep, that was my thought, VSTS seems to require a path to the test adapter. I don't use VSTS, but it looks like that just takes a path to the package root - I don't know if other CI systems require anything more specific. Although I guess this is only used for VS and VSTS, right?

PR #249, opened 2016-10-26
Fix for criical issue 248

### LINKED ISSUE #248: Adapter fails to discover tests
@brunoss commented on [Mon Oct 24 2016](https://github.com/nunit/nunit-vs-adapter/issues/131)

The following messages appears on `output` Window after trying to rebuild my `Nunit` test project

```
------ Discover test started ------
NUnit Adapter 3.5.0.0: Test discovery starting
Error: Unable to get runner for this assembly. Check installation, including any extensions.
RemotingException: Found two different objects associated with the same URI, '/ad6409a5_eed6_4df5_b143_6aea6249e17e/TestAgency'.
An exception occurred while test discoverer 'NUnit3TestDiscoverer' was loading tests. Exception: Found two different objects associated with the same URI, '/ad6409a5_eed6_4df5_b143_6aea6249e17e/TestAgency'.
========== Discover test finished: 0 found (0:00:00.4420455) ==========
```

Interestingly the last change that I made to my visual studio is that I put a product key on it. Before that the tests were able to be found.

This happened for me with Visual Studio Professional 2015
Version 14.0.25431.01 Update 3
Running as administrator

.Net Framework 4.5.2

NUnit version 3.5.0

I am using the visual studio extension test adapter 3.5.0

I already tried to reinstall the adapter after putting my product key.

The following message also appears at the very end of test discovery, if I rebuild my solution (which has tests with other testing frameworks)

`An exception occurred while test discoverer 'NUnit3TestDiscoverer' was loading tests. Exception: Format of the executable (.exe) or library (.dll) is invalid.`

---

@brunoss commented on [Mon Oct 24 2016](https://github.com/nunit/nunit-vs-adapter/issues/131#issuecomment-255699574)

The test discovery was successful after I changed the test to prefer 64 bits

Test -> Test Settings -> Default processor architecture -> x64

I'm going to leave this open if you want to pursue why it didn't work for x86

---

@CharliePoole commented on [Mon Oct 24 2016](https://github.com/nunit/nunit-vs-adapter/issues/131#issuecomment-255743373)

Since this is about the NUnit 3 adapter, I'm moving the issue to that project.

CharliePoole:
Fixes #248 

I haven't added a test for his, since it would be a fairly heavy-duty integration test. However, moving the call to GetRunner inside the try block seems sensible. In fact, I can't figure out why it ever worked!

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -49,10 +49,10 @@ namespace NUnit.VisualStudio.TestAdapter
 
                 ITestRunner runner = null;
 
-                runner = GetRunnerFor(sourceAssembly);
-
                 try
                 {
+                    runner = GetRunnerFor(sourceAssembly);
+
                     XmlNode topNode = runner.Explore(TestFilter.Empty);
 
                     // Currently, this will always be the case but it might change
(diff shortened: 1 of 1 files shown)


rprouse:
Looks good to me.

PR #252, opened 2016-10-27
Issue 1

OsirisTerje:
Hi! 

Please look over.  After the latest info from Charlie I think this is ready to be merged now.   The version number is already updated, I understand Charlie has another procedure for this, but I was not aware of that. 
The vsix is compatible with VS versions 11 through 14.  It does not support VS 15.


CharliePoole:
Why are you renaming the directory and creating entirely new files rather than changing in place? It makes it very difficult to review since the diff just says one file was removed and an entirely new one was created. 

It would probably help to have a general explanation of the change since it is being reviewed by two guys without your VS background. For example, why is there a .css file now. What's it for?

Basically, I'm just surprised that we have 14 files changed when I imagined there would be just one!

rprouse:
My recollection working with the v2 format myself is that there are more files so @OsirisTerje just created a new VSIX project and then copied the values across.

Looks good. I spent the most time looking through the manifest though.

src/NUnit3TestAdapterInstallV2/source.extension.vsixmanifest
```diff
@@ -0,0 +1,22 @@
...
+    <DisplayName>NUnit 3 Test Adapter</DisplayName>
+    <Description xml:space="preserve">NUnit 3 adapter for running tests in Visual Studio. Works with NUnit 3.x, use the NUnit 2 adapter for 2.x tests.</Description>
+    <MoreInfo>https://github.com/nunit/docs/wiki/VS-Adapter</MoreInfo>
+    <License>license.rtf</License>
+    <Icon>nunit3_32x32.png</Icon>
+    <PreviewImage>preview.png</PreviewImage>
+    <Tags>test, Testing, adapter, NUNit, TDD, NUnit3</Tags>
+  </Metadata>
+  <Installation>
+    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[11.0,15.0)" />
```
CharliePoole: Does this mean it will only install on Community edition? Don't we need entries for Pro, etc.?

CharliePoole: Also, we don't support 10.0.

src/NUnit3TestAdapterInstallV2/stylesheet.css
```diff
@@ -0,0 +1,129 @@
+body {
```
CharliePoole: What's the purpose of this file?

src/NUnitTestAdapter/Properties/AssemblyInfo.cs
```diff
@@ -21,5 +21,5 @@
 [assembly: ComVisible(false)]
 
 [assembly: Guid("c0aad5e4-b486-49bc-b3e8-31e01be6fefe")]
-[assembly: AssemblyVersion("3.5.0")]
-[assembly: AssemblyFileVersion("3.5.0")]
+[assembly: AssemblyVersion("3.5.1")]
```
CharliePoole: It may confuse things to change the Assembly version without changing the package versions to match. However, I can fix this up when doing the release.

src/NUnit3TestAdapterInstallV2/source.extension.vsixmanifest
rprouse: I am pretty sure this is a lowest SKU and above entry, so this will work on Community and every other edition.

src/NUnitTestAdapter/Properties/AssemblyInfo.cs
CharliePoole: I asked because the v2 format docs do not say that and because it never worked that way before. We had to make an entry for each sku in the past. Note also, that it has to work for people who __only__ have dev11 or dev12 installed, so that they are using the installer from that release.

OsirisTerje:
- Rob is right, I just created a new VSIX project to be sure I got all the stuff in there, and copied relevant things over.  Sorry that it messes up a few things, but it was much faster to do it this way, I also needed to compare with the other, and had both projects in there while I got it to work.
 I removed the index.html, but kept the css file, since Rob mentioned he wanted to add more stuff  in later.  Didn't give it that much thought - went for speed rather than details.

CharliePoole:
@OsirisTerje I'm confused. Looking at master branch, there's no index.html or css file at all. Where did they come from? Were they created as a part of your new project? If so, what are they for?

CharliePoole:
This seemed like a priority for 3.5.1 when I thought we could target dev15 with V2. But we need V3 format for dev15, which makes this less useful and less pressing. I'll run some tests. If it won't work on VS2012-13, that's a bit more than a detail. :grimacing:

OsirisTerje:
@CharliePoole   Yes, as said, I created a new vsix project.  With that came these files too.  Note that the project type id has changed, it does that for a V2, and I didnt want to try to copy things over and trying to get that to match.  I'll delete the css file.  And, the files changes are also things that were not correct otherwise, like the  csproj for the TestAdapter itself - causing the MS test object dll for VS2012  to be shipped along with the adapter.    I would also think it is better to go through vsix V2 before going to vsix V3, but....  whatever.....

OsirisTerje:
And, you must anyway remove the object model.  Frankly, I dont understand what is your issue here now

CharliePoole:
Regarding the removal of the object model, I agree. But I guess it has been there for many months without an issue being raised by anybody.

Questions do not imply criticism, but lack of understanding. It takes two people to merge this and I can't agree with what I can't understand. This is exactly how we have done hundreds of merges for the framework and console, even if it's not a familiar process for you. 

Actually, I assumed you were asleep so my last comment was trying to figure out how to do the release without your feedback. Are you available for a brief chat on Skype? That might be quicker.

CharliePoole:
My main remaining issue is that you said it may not work for people who don't have VS2015. That would not be acceptable as a release. Did I misunderstand you?

OsirisTerje:
I am online.    1)  It will work for VS1012-VS2015 on runtime, 2) To modify it, you need to use VS2015

Issue #253, opened 2016-10-28
Warnings about $RANDOM_SEED$ during build


CharliePoole:
@waldimen commented on [Fri Oct 28 2016](https://github.com/nunit/docs/issues/134)

I am using Nunit adapter with during TFS build.
In the build log file I got warnings for each assembly:

Exception System.IO.FileNotFoundException, Unable to restore random seed.
Could not find file  [DLLDIR]\$RANDOM_SEED$
[DLLDIR] is path to directory with TestAssembly.

Is it possible to disable the warnings ?
---

@CharliePoole commented on [Fri Oct 28 2016](https://github.com/nunit/docs/issues/134#issuecomment-256986343)

We don't want to disable that, but fix it! The $RANDOM_SEED$ file should be created by the adapter in discovery. If it isn't, then test cases using random parameter values won't work correctly. Is it possible that the test assembly directory is write-protected?

This is a potential adapter bug, so I'm moving the issue to that repo.

OsirisTerje:
The directory it tries to write to is the binary output directory, so we *know* it is writeable.  I see this:
![image](https://cloud.githubusercontent.com/assets/203432/19832903/a5ba95ee-9e30-11e6-9a68-88e606eeaa76.png)
which is the same directory as all other nbianries are written on the TFS build machine.

CharliePoole:
The fact that it isn't there is clear. The issue is why didn't it get written? It should be written in at the start of discovery, unless we are somehow bypassing that initialization on TFS.

OsirisTerje:
I can't see any traces of discovery being run in the logs.  It seems to go directly to execution, which sort of makes sense, since it is just running assemblies.  Waiting for confirmation from the softies.

OsirisTerje:
Just got it confirmed, running vstest.console for a bunch of assemblies, like TFS/VSTS build does, will not run the discoveryphase.  So it that case we should not react to the missing file, or create it if it missing the first time.

CharliePoole:
We should be able to observe that in the debugger as well.

I see two courses:
1. Detect tfs and act differently.
2. Give an error in discovery if we fail to write it and ignore it if it's missing.

Number 2 seems better to me. If we are only run once, in a single process, then the problem that the file is designed to solve doesn't exist.

CharliePoole:
Seems like a low priority to me.

OsirisTerje:
[...]

PR #254, opened 2016-10-31
Executor now doesnt warn on missing RANDOM_SEED file  #253

OsirisTerje:
Added a check for non-existing RANDOM_SEED file to the executor.  For a vstest.console test, only the executor is run, so the file is not needed to be present.  

I see my editor did some inadvertent changes to spaces etc.......


CharliePoole:
Looks good to me.

CharliePoole:
@OsirisTerje I closed the issue manually. Issue # in the title of the PR doesn't help. Only putting "fixes #253" or "fix #253" in a comment or a commit message does it. We usually put it as the first line of our descriptions.

PR #263, opened 2016-11-11
Updating vsix for vsix3 format for VS 15

### LINKED ISSUE #262: Support for VS 15
The adapter needs to be updated to support VS 15.  That requires changes to the VSIX format, and the specified verson range.

OsirisTerje:
The vsix manifest file has been updated to support the new VSIX3 format.  The version ranges have been updated to include the VS 15.  A build package for the VS 15 SDK build tools have been included. 

Fixes #262


src/NUnit3TestAdapterInstallV2/source.extension.vsixmanifest
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="NUnit.NUnit3TestAdapter" Version="4.0.0.0" Language="en-US" Publisher="Terje Sandstrom" />
+    <Identity Id="NUnit.NUnit3TestAdapter" Version="3.6.0.1" Language="en-US" Publisher="Terje Sandstrom" />
```
CharliePoole: Why are we updating the version now rather than at release?

rprouse:
The files in .vs folder need to be deleted. We should add a gitignore for the directory. 

We tested 2015. I will retest. My work computer has 2013.I can test that Monday. 

We changed the version because it was already at 4.0.

rprouse:
I have testing this and the NUnit 2 adapter in the latest release of VS "15" and in VS 2015. I will also test in 2013. Does anyone have 2012 for testing? I can provide binaries.

We should update the assembly versions to match the VSIX version. As for the NuGet package, I don't think an update is needed for VS "15" since it isn't a VSIX. Is there any reason to release a new NuGet package? Are there issues fixed that we want to include?

src/NUnit3TestAdapterInstallV2/source.extension.vsixmanifest
OsirisTerje: I just need to do this to test the extension.  It is much easier than anything else.  I did the same earlier, you remember, with the 2.X adapter.  I could have dropped commiting this change, but then there are other changes in the same file.

OsirisTerje: * I need to update the assemblies too. 
* I have tested it now on VS 2015, and it works there.  I don't have the earlier versions on my machine here, but it working on 2015 is a good sign.  
* I see one file in the commit here I overlooked, and I dont recognize, the .vs/slnx.sqlite file, I don't think that should be included. 
* Where do you see a new install folder?  If there is a new one, that is not intentional.

OsirisTerje: About the nuget package - this change is really only relevant to the vsix, but there are 2 other changes we get along, 253 and 254, and that means the nuget should be updated too. 
 I have tried using the nuget 3.5.1 package on VS 2017, and it works as it should, so there is nothing more needed to be done there.   We need to update the text for it however.

OsirisTerje:
See comments above (I added them to Charlies), but #253 and #254 is fixed, so we should update nuget too.

src/NUnit3TestAdapterInstallV2/NUnit3TestAdapterInstall.csproj
```diff
@@ -1,8 +1,31 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\Microsoft.VSSDK.BuildTools.15.0.25907-RC2\build\Microsoft.VSSDK.BuildTools.props" Condition="Exists('..\..\packages\Microsoft.VSSDK.BuildTools.15.0.25907-RC2\build\Microsoft.VSSDK.BuildTools.props')" />
   <PropertyGroup>
-    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+    <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
```
CharliePoole: Requires updating docs

src/NUnit3TestAdapterInstallV2/source.extension.vsixmanifest
CharliePoole: We have a general approach to doing releases that involves not mixing the actual code fixes with the release changes. Permanent code changes are done on the master branch. Release changes, that should not go to master, on the release branch. This is all pretty well documented, with perhaps a few things that need updating.

The master branch needs to stay at 4.0, which we decided would be our next major/minor release.

The last release we did was 3.5. So making this new release 3.6.1 is going to be a bit confusing to users. I think this number choice is a side effect of that fact that you are making the version change on master rather than the release branch. We should have decided in advance whether this was to be release 3.5.1 or release 3.6. I guess there are arguments either way - it depends whether you consider "Works on dev 15" to be a bug fix or a feature.

rprouse:
We have two options. The easiest would be to treat this as a release branch that won't get merged in directly. I will treat it like the release branch in other projects (albeit with a few more changes) and create a merge branch/pull request off of this where I switch back to 4.0 before we merge.

The second option is to revert back to 4.0.0 in this branch, merge this PR, then create a release branch. That would be technically more correct, but we are a bit under the gun for time, so I would prefer to go with the first option.

Based on that, **Do not merge this PR**. When everyone approved the changes and it is tested, we can create the release from this PR, tag it accordingly and I will create a branch to merge everything except the version changes into master. I will also rename the branch to a more meaningful release branch name.

Since this is sort of a feature and we have the random seed bug fixed too, let's call this 3.6. I will update versions and push accordingly.

rprouse:
The documentation that needs updating is at the very end of the release notes,  https://github.com/nunit/docs/wiki/Packaging-the-V3-Adapter.

We should hold off on updating them until the name is announced.

rprouse:
Well, there is a new one. I checked out the remote branch `issue-262` on Windows and it pulled in this PR, but when I push, it creates a new branch. This PR is on `Issue-262` (first letter capitalized). I was wondering why my changes weren't showing up here!

rprouse:
Changes are pushed. Do you have any other remaining changes you would like to see @CharliePoole?

CharliePoole:
No, and I agree with your approach to get this out as a release branch. Will you rename it to release/3.6? - it's probably academic but would be handy to keep around if we have any errors.

rprouse:
I have renamed the branch, but cannot attach it to this PR, so I am going to close this PR and open a new one for the release. I will do the release Monday after checking the build on my work computer which has more versions of Visual Studio on it.

CharliePoole:
Yes, I've run into that dilemma before. Anyway, the two PRs with their comments are linked for future reference.

PR #264, opened 2016-11-13
Release 3.6 - Support for Visual Studio vNext

### LINKED ISSUE #262: Support for VS 15

rprouse:
The vsix manifest file has been updated to support the new VSIX3 format. The version ranges have been updated to include the VS 15. A build package for the VS 15 SDK build tools have been included.

This code has been reviewed in #263 and there are no additional changes here. I will create the release from this branch this week after testing with more versions of Visual Studio.

Fixes #262

**Do not merge** I will create a release merge branch off this PR once it is released.

@CharliePoole I named this branch `release-3.6` not `release/3.6` because I don't think the Cake build in this repo has been updated to support the slash.


CharliePoole:
Actually, I suggested release/3.6 because I know it's your preference. 😄 I'm a dash guy.

.gitignore
```diff
@@ -179,3 +179,134 @@ $RECYCLE.BIN/
...
+# JetBrains Rider
+.idea/
+*.sln.iml
+# CodeRush
+.cr/
+# Python Tools for Visual Studio (PTVS)
+__pycache__/
+*.pyc
+# Cake - Uncomment if you are using it
+tools/
```
CharliePoole: This works because the latest cake bootstraps will download a packages.config if one is not found. However, we have been following the practice of controlling the version of Cake used through the packages.config. Without that, I think we would be giving up our control of what Cake version we are using in each project.

Also, I have found that nunit console needs to be specified here rather than in a #tools comment in order for the NUnit3 command to work on Linux.

The upshot is that I think we need to add !tools/packages.config

CharliePoole: Maybe we should fork the github/gitignore repo and add to it.

src/NUnit3TestAdapterInstallV2/Properties/AssemblyInfo.cs
```diff
@@ -10,7 +10,7 @@
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("NUnit3TestAdapterInstall")]
-[assembly: AssemblyCopyright("")]
+[assembly: AssemblyCopyright("Copyright © 2011-2016, NUnit Software")]
```
CharliePoole: Both the NUnit 3 and NUnit 2 adapters say that they are copyright Charlie Poole in some places and NUnit Software in other places. Let's not further complicate things by adding another thing we may have to change.

I understand you're taking this from the discussion we had about the test generator, but that was a special case because of the Microsoft involvement and I wasn't intending it to be a global decision. In fact, I don't think a global decision is possible wrt copyright - it has to reflect the actual facts in each case. I think we can resolve this after the release and I'll create an issue for it.

rprouse: Actually, I just copied it from the NUnit 2 adapter, but I will update to Charlie Poole for this release.

rprouse:
On a whim, I downloaded the VSIX built on AppVeyor using VS2105 and installed it in vNext and it works fine, so it appears that we don't need the latest version of Visual Studio to build. It is probably because the build tools for the vNext SDK are pulled into the project as a NuGet package now?

rprouse:
All requested changes have been made and I have tested this on the supported versions of VS. We need to release by Wed, so I plan on doing the release tomorrow (Tuesday) so we are sure we are ready.

rprouse:
This has been released. I will create a new pull request to merge everything except version changes back into master.

PR #282, opened 2016-12-15
Resolve differences in copyright headers

### LINKED ISSUE #265: Resolve confilicting Copyright statements
For historical reasons, we state that the copyright for the adapter is held by Charlie Poole in some places and by NUnit Software in others. It's not too terrible, since NUnit Software is a DBA of mine, but it's potentially confusing and should be resolved.

There is no good category for this, so I'm just calling it a build issue. Priority is high because I want the current copyright to be completely clear in order to be able to easily transfer it elsewhere at some point.

CharliePoole:
Fixes #265

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -1,6 +1,25 @@
-﻿// ****************************************************************
-// Copyright (c) 2011-2015 NUnit Software. All rights reserved.
-// ****************************************************************
+﻿// ***********************************************************************
+// Copyright (c) 2011-2015 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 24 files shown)


src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -1,6 +1,25 @@
-﻿// ****************************************************************
-// Copyright (c) 2011-2015 NUnit Software. All rights reserved.
-// ****************************************************************
+﻿// ***********************************************************************
+// Copyright (c) 2011-2015 Charlie Poole
```
rprouse: Outside the scope of this pull request, but I was reading Clean Code and the recommendation is to not put these copyright headers in every file since they are just noise that forces you to scroll and they are redundant with the LICENSE file in the root. Makes sense to me. Maybe something for future discussion?

CharliePoole: It's something that has been disputed over the years, for sure. Most people using MIT seem to put the entire thing in each header. The shorter notice is a holdover from when this was licensed differently. I think most people writing about copyright and license say to put a notice in the files, while those merely concerned with coding issues will discount it. We can discuss more - or more likely, you guys can after I'm gone. 😄

PR #283, opened 2016-12-15
Display warnings in the adapter

### LINKED ISSUE #273: Report Warnings in VS adapter
Warning results, which are now supported by the framework should be reported as such to VS.

CharliePoole:
Fixes #273 

As of now, this PR shows warnings as "Skipped" tests, in order to get the warning icon. We can easily change to Passed or Failed, but there is no Warning status per se in VS.

If desired, it's possible to add a run setting that will determine how we display warnings.


rprouse:
Just passing the tests seems wrong, the warnings will never show up, and since there is no alternative, I think skipped is the best choice. I don't think we should add a runnsetting unless we get feedback that it is needed, ie. nobody can agree if warnings should be passed or skipped.

CharliePoole:
Ready for review and merge

Issue #218, opened 2016-08-11
No tests found to run when TestFixture is a nested class


rprouse:
@pitlyck commented on [Thu Aug 11 2016](https://github.com/nunit/nunit-vs-adapter/issues/126)

Hi,

I am unable to run tests by selecting "Run Tests" from the context menu in VS2015 when the test fixture is a nested class.

Right-click in the window with code below, then select "Run Tests". In output window I can see output from tests: **No tests found to run.** 

Example:

```
public class NestedClasses
    {
        [TestFixture]
        public class NestedClass1
        {
            [Test]
            public void Test1()
            {
                Assert.Pass();
            }
        }
    }
```

I can see the test in list of tests in Test Explorer window and I am able to run it from there, but this is not what I want to do when I need to run/debug specific test/fixture that I am currently working on.

I am sure it worked before switching from NUnit 2.6.4 to NUnit 3.2.0.

I tried similar scenario with Microsoft VisualStudio UnitTesting framework and the test was run properly, so I expect this is nunit-vs-adapter issue.

VS 2015 Update 3 + NUnit3 Test Adapter v3.4.1 + NUnit 3.2.0

---

@rprouse commented on [Thu Aug 11 2016](https://github.com/nunit/nunit-vs-adapter/issues/126#issuecomment-239179975)

Since this is an issue with the NUnit 3 adapter, I am moving this issue to that repository.

CharliePoole:
I'm guessing that the bottom panel of the test window shows "No source found" for these tests. Is that correct?

pitlyck:
Yes, when I select the test in Test Explorer window the following message is displayed at the bottom panel: "Source: no source available"

CharliePoole:
I think this has to do with the difference between how nested classes are represented using Reflection versus Mono.Cecil. 

Reflection uses a '+' between the outer class and the inner one. Cecil uses the ECMA-235 representation, which specifies a '/' to separate them.

We are getting the class in reflection format from NUnit and then using it to ask the engine and locate the source code. The engine uses Cecil.

We need to change the format somewhere but it will take some thought to decide whether this should be done in the framework, the engine or the adapter!

CharliePoole:
When working on this, we should probably look at other differences caused by Cecil's use of ECMA-235 and deal with them as well. I believe that the repreentation of generics may differ.

jblazek:
[...]

PR #285, opened 2016-12-19
Fix no source available for nested classes

jblazek:
Updated adapter to locate source file and line for tests in nested
classes.  Previously the message 'No source available' was displayed for
tests in sub classes and some Test Explorer functionality was unavailable.

Issue #218


CharliePoole:
Looks good to me. Needs one more committer review. @nunit/vs-extensions-team

rprouse:
Looks good to me too, thanks for the fix. 👍

Issue #276, opened 2016-12-02
Adapter requires test project pbd's to be generated for tests to be executed


ppejovic:
See #229 

Disabling pbd generation in the project properties of a test project prevents the tests being discovered in Visual Studio. This was for a very basic test project (class library type) I created to test NUnit 3.0. The test project has no dependencies aside from NUnit and NUnit3TestAdapter. 

I believe the problem code is [here](https://github.com/nunit/nunit3-vs-adapter/blob/8a9b8a38b7f808a4a78598542ddaf557950c6790/src/NUnitTestAdapter/NavigationDataProvider.cs#L79):

**NavigationDataProvider.cs#L79**

`var readerParameters = new ReaderParameters() { ReadSymbols = true };`

The [docs ](https://github.com/jbevain/cecil/wiki/Debug-symbols) from Cecil state:

> From an API standpoint, there are two ways of interacting with debug symbols. The easiest is to simply pass `True` to ReaderParameters.ReadSymbols or WriterParameters.WriteSymbols. Cecil will try to load the debug store support assembly depending on the runtime (pdb on .net, mdb on Mono) and read and write the appropriate symbols.

CharliePoole:
@ppejovic In my earlier experiments with this, not having the pdb available simply prevented the ide from displaying line numbers, which seemed like a valid tradeoff. In order to reproduce this, can you let me know exactly what versions of VS and the NUnit VS adapter you are using?

ppejovic:
@CharliePoole I've added a repro here: https://github.com/ppejovic/nunit3-pbd-repro

Microsoft Visual Studio Professional 2015
Version 14.0.25431.01 Update 3

OsirisTerje:
I can confirm this.   In NUnit2 it behaves as @CharliePoole  says, without a pdb, you can't click and go to source, but in NUnit3 it doesn't show the test at all.

CharliePoole:
I made this high priority rather than critical. As a reminder, critical would cause us to do an immediate hotfix release. This doesn't seem in that category, since there's a workaround.

OsirisTerje:
@CharliePoole  Just to be clear on what you want:  Should this be closed now or after the PR are merged?  I would assume it is best to close it after the PR merge, right?

CharliePoole:
[...]

PR #286, opened 2016-12-25
Checks if pdb file exist before trying to read symbols.  If it doesnt…

OsirisTerje:
…, the test will now appear but you can't jump to the source code for test  #276


CharliePoole:
Please close the issue and delete the branch if you're done with this. You can keep the branch if you prefer, but it doesn't actually have any information you can't get by looking at the PR and the commits in master.

Issue #268, opened 2016-11-16
Make the Icon Larger


rprouse:
I think it is because we are using the recommended icon size from the v1 VSIX format, but our icon on the new Visual Studio Marketplace is much smaller than most other extensions. Minor from a code perspective, but I think important from a professionalism perspective.

https://marketplace.visualstudio.com/items?itemName=NUnitDevelopers.NUnit3TestAdapter

![image](https://cloud.githubusercontent.com/assets/493828/20329498/9d847dc0-ab66-11e6-9b13-a652cca8ff25.png)

![image](https://cloud.githubusercontent.com/assets/493828/20329509/ad65a304-ab66-11e6-8b7e-9159500d4b4c.png)
====================

Affected repos:

- [x] Framework
- [x] Console
- [x] nunit.xamarin (https://github.com/nunit/nunit.xamarin/issues/70)
- [x] nunit-vs-templates
- [ ] nunit.portable.agent
- [ ] dotnet-test-nunit
- [ ] nunit3-vs-adapter (In progress?)
- [ ] nunit-vs-adapter (Blocked: https://github.com/nunit/resources/issues/4)
- [ ] nunit-v2-framework-driver
- [ ] nunit-project-loader
- [ ] nunit-vs-testgenerator
- [ ] nunit-v2-result-writer
- [ ] vs-project-loader
- [ ] NUnit.System.Linq

External projects:
- [ ] teamcity-event-listener
- [ ] nunit3-tdnet-adapter

CharliePoole:
We may need to keep this in multiple sizes.

ChrisMaddock:
We actually have the same issue on NuGet gallery (for _almost_ all packages), and of course for all the other extensions. 

New icons for everyone! 😄

rprouse:
>We may need to keep this in multiple sizes.

True, we will need to go back to VS 2012 and ensure whatever we use there continues to work. The screenshot above is for the 2012 version of Web Essentials, so I expect the larger icons are fine there too. The small icons are probably a hold back to the VSIX v1 format.

CharliePoole:
What are we using for file explorer with the console?

rprouse:
The console uses an ICO with large size icons embedded in it. I think the NuGet and VSIX require large PNG files, but I need to check.

rprouse:
[...]

PR #287, opened 2016-12-26
Rename Random seed file

### LINKED ISSUE #256: Rename $RANDOM_SEED$
I noticed a bit of confusion over the purpose of this file at http://stackoverflow.com/q/40364171/4026748

Perhaps we should rename it to $NUNIT_RANDOM_SEED$ so people will know who creates it.

CharliePoole:
Fixes #256

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -12,6 +12,7 @@ namespace NUnit.VisualStudio.TestAdapter
 {
     public class AdapterSettings
     {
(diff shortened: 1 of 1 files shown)


OsirisTerje:
The icon size issue is #268 , but should we increase it to 90x90 as MS recommends or stay at 64x64 ?
And, should two issues be in the same PR - was that unintentional?

Random seed file change looks fine :-)

CharliePoole:
It looks like I created the changes for #268 on top of this branch and not in a new one. I'll back them out.

CharliePoole:
Merging with just the file name change.

PR #289, opened 2016-12-28
Test parameter parsing out of framework

### LINKED ISSUE #288: Test parameters containing semicolons are truncated
@CharliePoole in https://github.com/nunit/nunit/issues/1885#issuecomment-268872044:

> @jnm2 I'm assigning this issue to you. Is there already an nunit 3 adapter issue? If not there should be.

jnm2:
Umbrella: https://github.com/nunit/nunit/issues/1941

Closes #288.

### DIFF
--- a/src/NUnitTestAdapter/NUnitTestAdapter.cs
+++ b/src/NUnitTestAdapter/NUnitTestAdapter.cs
@@ -30,6 +30,7 @@ extern alias ENG;
 using TestEngineClass = ENG::NUnit.Engine.TestEngine;
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
@@ -184,19 +185,9 @@ namespace NUnit.VisualStudio.TestAdapter
                 package.Settings[PackageSettings.RandomSeed] = Settings.RandomSeed;
 
             if (Settings.TestProperties.Count > 0)
-            {
-                var sb = new StringBuilder();
-                var index = 0;
-                foreach(string name in Settings.TestProperties.Keys)
-                {
-                    if (index++ > 0) sb.Append(";");
-                    sb.AppendFormat("{0}={1}", name, Settings.TestProperties[name]);
-                }
+                SetTestParameters(package.Settings, Settings.TestProperties);
 
-                package.Settings[PackageSettings.TestParameters] = sb.ToString();
-            }
-
-            // Always run one assembly at a time in process in it's own domain
+            // Always run one assembly at a time in process in its own domain
             package.Settings[PackageSettings.ProcessModel] = "InProcess";
 
             if (Settings.DomainUsage != null)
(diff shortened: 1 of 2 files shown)


CharliePoole:
Isn't this a duplicate of #288?

jnm2:
This is a PR, #288 is an issue.

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -224,6 +215,27 @@ private TestPackage CreateTestPackage(string assemblyName)
...
+            if (testParameters.Count != 0)
+            {
+                // Kept for backwards compatibility with old frameworks.
+                // Reserializes the way old frameworks understand, even if the parsing above is changed.
+                // This reserialization cannot be changed without breaking compatibility with old frameworks.
+
+                var oldFrameworkSerializedParameters = new StringBuilder();
+                foreach (var parameter in testParameters)
+                    oldFrameworkSerializedParameters.Append(parameter.Key).Append('=').Append(parameter.Value).Append(';');
+
```
CharliePoole: Extra semicolon at end is not backward compatible.

jnm2: It gets removed, see [this line](https://github.com/nunit/nunit3-vs-adapter/pull/289/files/f7a4d9a978fc4537b5be7ce922f9d950ec736ea6#diff-730f6bad2666ef06114645aaaa78868eR235).

jnm2: Purely for simplicity.

CharliePoole: OK then... it's one of those matters of individual style I keep preaching about. :smile:

jnm2: Yep, it's a tradeoff 😆 do you want a complex `ToString` or do you want another local and complex logic in the loop? I honestly couldn't care which it is but this is my go-to.

jnm2: I'm the one that keeps preaching about coupling though. 😆 This style breaks if the loop iterates zero times, so it's coupled to the containing `if` check. That could be less desirable simply because it's a distant side effect. Maybe I'll switch back to the complex loop in the future.

Or just wait until `StringBuilder.AppendJoin` makes it into .NET Standard. =)

CharliePoole:
Is that in .Net 2.0?

oznetmaster:
```
var oldFrameworkSerializedParameters = String.Join (";", testParameters.Select (tp => tp.Key + "=" + tp.Value).ToArray());
```

(from my phone so may have typos)

jnm2:
> Is that in .Net 2.0?

Right, never mind. StringBuilder.AppendJoin is currently in the API review stage for .NET Core with a tag to consider porting to .NET Framework.  dotnet/corefx issue 3419.

If that's the entire string you're building, `string.Join` is even better except for that `ToArray`.

jnm2:
The most performant .NET 2.0 code is likely what I have, with the semicolon removed rather than a condition inside the loop. If not, then the fastest is probably the same with the loop condition on sb.Length or something. Could be put into an `AppendJoin(IEnumerable<KeyValuePair<,>>, char pairSeparator, char keyValueSeparator)` extension method if we hypothetically ran into this a lot.

oznetmaster:
```
2.0 does not have String.Join (string, IEnumerable<string>), so the ToArray is necessary.
```

CharliePoole:
Anyway, this is merged. :smile:

jnm2:
@oznetmaster exactly. Otherwise this PR would have looked different. :-(

@CharliePoole I'll try to be less chatty then =D

PR #294, opened 2017-01-14
Handle multiple assert in adapter

### LINKED ISSUE #258: Modify adapter to display multiple assert information
@CharliePoole commented on [Sun Nov 06 2016](https://github.com/nunit/nunit-console/issues/127)

Currently, the adapter uses the single failure message created for compatibility in order to display the result of a test using multiple asserts. It should display separate results for each assert, provided that VS is able to handle it.

CharliePoole:
Fixes #258


OsirisTerje:
It looks ok to me, and the unit tests seems to cover the cases.   I haven't built it yet and tried, but can do so in a couple of days  (on travel).  Otherwise seems good to merge to me.

CharliePoole:
I'll merge it but if you come up with any other changes needed, they can be done as a part of the release... which is the next thing I want to get on with.

PR #309, opened 2017-03-17
Proposed solution to Issue-231

pi3k14:
Added more information if TestDiscovery went ok, but TestExecuter failed (maybe because of different NUnit versjons in dll's)

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestExecutor.cs
+++ b/src/NUnitTestAdapter/NUnit3TestExecutor.cs
@@ -266,7 +266,13 @@ namespace NUnit.VisualStudio.TestAdapter
                     }
                 }
                 else
-                    TestLog.Info("NUnit failed to load " + assemblyName);
+                {
+                    var msgNode = loadResult.SelectSingleNode("properties/property[@name='_SKIPREASON']");
+                    if (msgNode != null && (new[] { "contains no tests", "Has no TestFixtures" }).Any(msgNode.GetAttribute("value").Contains))
+                        TestLog.Info("Assembly " + assemblyName + " contains no tests supported by this runner: " + _activeRunner.GetType().AssemblyQualifiedName);
+                    else
+                        TestLog.Info("NUnit failed to load " + assemblyName);
+                }
             }
             catch (BadImageFormatException)
             {
--- a/src/NUnitTestAdapter/NUnit3TestExecutor.cs
+++ b/src/NUnitTestAdapter/NUnit3TestExecutor.cs
@@ -269,7 +269,7 @@ namespace NUnit.VisualStudio.TestAdapter
                 {
                     var msgNode = loadResult.SelectSingleNode("properties/property[@name='_SKIPREASON']");
                     if (msgNode != null && (new[] { "contains no tests", "Has no TestFixtures" }).Any(msgNode.GetAttribute("value").Contains))
-                        TestLog.Info("NUnit couldn't find any tests in " + assemblyName + ". (Probably an assembly dependency problem in your projects.");
+                        TestLog.Info("NUnit couldn't find any tests in " + assemblyName);
                     else
                         TestLog.Info("NUnit failed to load " + assemblyName);
                 }
--- a/src/NUnitTestAdapter/NUnit3TestExecutor.cs
+++ b/src/NUnitTestAdapter/NUnit3TestExecutor.cs
@@ -269,7 +269,7 @@ namespace NUnit.VisualStudio.TestAdapter
                 {
                     var msgNode = loadResult.SelectSingleNode("properties/property[@name='_SKIPREASON']");
                     if (msgNode != null && (new[] { "contains no tests", "Has no TestFixtures" }).Any(msgNode.GetAttribute("value").Contains))
-                        TestLog.Info("Assembly " + assemblyName + " contains no tests supported by this runner: " + _activeRunner.GetType().AssemblyQualifiedName);
+                        TestLog.Info("NUnit couldn't find any tests in " + assemblyName + ". (Probably an assembly dependency problem in your projects.");
                     else
                         TestLog.Info("NUnit failed to load " + assemblyName);
                 }


src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -266,7 +266,13 @@ private void RunAssembly(string assemblyName, TestFilter filter)
                     }
                 }
                 else
-                    TestLog.Info("NUnit failed to load " + assemblyName);
+                {
+                    var msgNode = loadResult.SelectSingleNode("properties/property[@name='_SKIPREASON']");
+                    if (msgNode != null && (new[] { "contains no tests", "Has no TestFixtures" }).Any(msgNode.GetAttribute("value").Contains))
+                        TestLog.Info("Assembly " + assemblyName + " contains no tests supported by this runner: " + _activeRunner.GetType().AssemblyQualifiedName);
```
CharliePoole: I'm not sure what the user will understand by "no tests supported by this runner." Is there another runner that the user should use? What do we mean by a runner? etc.

I think the fix is good if we can come up with a meaningful message.

pi3k14: Yeah, well - I don't know what that should be. I stumbled across this issue when one of the projects in my solution was linked with an older version of NUnit, I have no idea what other situations could cause this.
Of course I can change the text runner with TestRunner, but that doesn't help much.

OsirisTerje: What about:  "contains no tests supported by NUnit3" .    It is NUnit3 that gives this message, so the adapter should only report it.

pi3k14: Maybe, but it is not quite correct. In my case it was a mismatch between 3.4 and 3.5 that caused the fault (thats why I returned the full AssemblyQualifiedName in the message)

OsirisTerje: It might not be absolutely correct, but I think the message should be directed toward the general user, and inform about possibly user mistakes. If there is a NUnitAdapter/NUnit mismatch issue that is possibly a bug in the adapter, as any adapter version should ideally be able to run any NUnit version (and I know it is not really true this will happen).   But I am not sure we need to surface that in this message.  Alternatively, replace "this" with the explicit name and version we're running.  "contains no NUnit3 tests that can be run by NUnitAdapter version xxx"

CharliePoole:
That's the sort of confusion I was worried about. The user may think that "runner" is the adapter. Or maybe the engine and one of its internal runners. But we know it's the nunit 3 framework that was used and failed to find the test, so let's say that. The actual reason it failed to find the test is another matter but at least the problem should be reported to the right project.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
pi3k14: Okay, so the message should be along the lines of "NUnit3 was able to discover some tests in {assembly}, but could not run thoose tests. Please inspect your projects assembly dependencies. (TestRunner was: {AssemblyQualifiedName})" ?

OsirisTerje: Yeah, that looks good to me :-)

CharliePoole:
I still have question. How do we know it originally discovered tests but now fails to find them? Do we just assume it's so

OsirisTerje:
The code is in the test executor, so I would assume it is safe to assume that they were discovered first, and then attempted to be run, but failed.

CharliePoole:
@OsirisTerje I agree that that sounds logical. However, how and when VS calls us is sufficiently obscure and undocumented, that it might be worthy running some experiments to figure out what the rules are. For example, are we ever called for execution when we returned no tests from discovery? Does this vary between the IDE, vsconsole, TFS and (eventually) dotnet test?

We have never been able to get solid info about this from Microsoft. In part this is probably due to the fact that various groups there work on the code and they are somewhat compartmentalized. We, on the other hand, have a need to see the bigger picture in order to integrate with their software. 

I'm thinking we should do the best we can to come up with a reasonable message, merge this PR and close the issue. Following that, perhaps we should create a new issue to experimentally determine for ourselves exactly how and when we are called in each environment.

I'm also concerned that whatever we do here may become broken when we upgrade to use the newest engine. So as yet another issue, maybe we need to set up a build of the adapter that uses the latest development version of the engine. what do you think?

OsirisTerje:
@CharliePoole Yes, good points.     With the possibility for being broken later, I assume you mean the check against a certain string value., which of cause is brittle.   But since this is XML parsing, are there any other good options?   Agree with the message though, so I would just suggest it states what it know, and don't assume anything.  What we know is that we're trying to execute NUnit3 tests and that failed.   If discovery found them, that is visible in the output log too. 

I fully agree we need to set up some builds to check the different issues and possible scenarios, but also keeping these builds for later checks (like what I also experienced now with the 2.X adapter and the 3.5 issue.).  There are too many variations now, so the risk of breaking something is just getting bigger.

CharliePoole:
@OsirisTerje Feel free to merge when you think the message is right and we can deal separately with this other stuff.

@pi3k14 Including the name of the runner in your message doesn't make sense as it is constant. I think we just want to say that NUnit (i.e. the nunit framework, of whatever version you are using) didn't find any tests. I realize it's only a small improvement on the existing message, but that's all we actually know for sure at that point in the code.

CharliePoole:
@OsirisTerje I was thinking of the check for a string but I was also thinking we could change how we deal with assemblies with no tests completely in another engine release!

pi3k14:
Hm, I pushed this change because this message would have helped me a lot while debug'ing a problem I had.
The problem was running tests for a solution with five test assemblies, where one was built with NUnit 3.4 the others built with 3.5. The only help I got was "NUnit failed to load xxx". The TFS log showed
Information: Running all tests in C:\VsoBuildAgent\Agent3\_work\6\a\xxx
Information: NUnit failed to load C:\VsoBuildAgent\Agent3\_work\6\a\xxx

The suggested fix now became a bit bland.
Anyway, I'm pushing an update.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -266,7 +266,13 @@ private void RunAssembly(string assemblyName, TestFilter filter)
                     }
                 }
                 else
-                    TestLog.Info("NUnit failed to load " + assemblyName);
+                {
+                    var msgNode = loadResult.SelectSingleNode("properties/property[@name='_SKIPREASON']");
+                    if (msgNode != null && (new[] { "contains no tests", "Has no TestFixtures" }).Any(msgNode.GetAttribute("value").Contains))
+                        TestLog.Info("NUnit couldn't find any tests in " + assemblyName + ". (Probably an assembly dependency problem in your projects.");
```
rprouse: The `(Probably an assembly dependency problem in your projects.` bothers me as there are several reasons that this could be the case. I would prefer to leave that part out or at least change it to Possibly. At the very least, the trailing `)` is missing in the message.

CharliePoole: That's why I had not approved this earlier, although I can't see my comment here!

This is a perennial problem! From the perspective of the user encountering the problem, it seems that most cases of a particular error probably arise from a certain cause, so they want the message to explain the cause. However, such a message would be misleading in many other cases.

Idea: Can we have a link in the message to an extended explanation? Would a link work in the output window?

pi3k14: As a developer I understand your concerns, but as a user I would prefer some information instead of none. Some information will help some, no information will help none.

Your idea might work, but what about at tag before such "extra" information, and a disclaimer in the system documentation about what that tag implies? Our just a reference ID, and a documentation page collecting such information? (could even reference back to the commit/PR's where the issue was handled)

PR #315, opened 2017-04-10
Support base-class tests from external assemblies

### LINKED ISSUE #314: Any TestFixture deriving from a base-class which defines unit-tests will fail when the base-class is from another class-library
We use some base-classes (from a common util-assembly) defining common tests (like correct implementation of comparison/equality/etc.), and use derived tests for specific types (which extend on the base tests).

This worked until 3.0.10 of the test-adapter, but got broke.

To illustrate this I've added a minimal sample below.

**Expected behavior** is that the test-adapter finds and runs two test-cases for `CommonTests`, namely `TestCaseApplicableToAllTypes` and `SpecificTest`.

**Actual behavior** is that only `SpecificTest` is discovered, but **no** test in the assembly containing class `SpecificTypeTests` is executed.

3.0.10 found the test-cases from `CommonTests`, but added them to a group/project `External`. Not perfect, but at least it executed the tests.

**Code Sample**
Needs two projects, both referencing NUnit (I'm still at 3.2.1, but should happen with newer version as well).

*Project 1 a.k.a. "TestUtils"*
```cs
namespace TestUtils {
    using NUnit.Framework;

    public abstract class CommonTests {
        [Test]
        public void TestCaseApplicableToAllTypes() { }
    }
}
```

*Project 2 a.k.a. "Tests"*
```cs
namespace Nunit3TestAdapterProblems {
    using NUnit.Framework;
    using TestUtils;

    [TestFixture]
    public class SpecificTypeTests : CommonTests {
        [Test]
        public void SpecificTest() { }
    }
}
```

gimpf:
Fixes #314 

The problem was caused because the Mono.Cecil failed to resolve types from external assemblies, even if they are in the same directory.

This is fixed by explicitly adding the directory of the base-library to the search-path for Mono's assembly-resolver.

This fix works for me, but I'd have to lie when I'd say I've tested it extensively. Please see [gimpf/nunit3-vs-adapter-issue314](https://github.com/gimpf/nunit3-vs-adapter-issue314) for a test-case. It fails using the regular adapter, but works using the fixes from this PR.


CharliePoole:
The PR looks like an improvement. I'm vaguely aware that we recently made almost the same change, however, so I want to look more closely when I'm back home with my development system in a few hours.

This doesn't explain that you are seeing the same problem with the console runner, since your fix would not have anything to do with that.

CharliePoole:
I approved but with a question for @OsirisTerje. We need his second review to merge. Thanks for your work!

gimpf:
@CharliePoole Sorry that I didn't find the existing work. It didn't occur to me that searching for "navigation" would turn up some work in progress, though that's rather obvious in hindsight... Thanks for integrating it ;-)

CharliePoole:
Well it's in a different repo and so hard to find unless you know it's there.

gimpf:
I've now integrated the changes from the nunit2 adapter, and cleaned up the code a bit for readability.

Since this navigation problem is strictly dependent on having multiple assemblies, should we extend the NavigationDataTests (that is, introduce another test assembly), or do you use separate integration tests for such issues?

OsirisTerje:
@gimpf   Good that you integrated those changes, and yes, @CharliePoole , I think that is necessary, this was what solved the 2.X adapter issue.  I am not sure exactly how this works out with the different Mono.Cecil libs, but I would better be safe than sorry here. 
@gimpf   About the integration tests, we have discussed them earlier, but currently we dont have any other than what is in the repo.  If you want to add another test assembly, it will not harm in any way.,  If they start to slow down things we can always move them later, but then they are done - so, yes please, if you want to.

gimpf:
@OsirisTerje I've now added two additional tests for the NavigationDataProvider for types that reference other assemblies. Also, since the first time I opened the PR, I've refactored some things, you might want to revisit the code.

Those two new tests brought to light another issue regarding the navigation-data: The way we resolved the base-types caused that no navigation-data for the types was loaded. I remember that you mentioned an issue regarding "cannot jump to source" in the linked issue. That might be resolved now as well (though I haven't checked).

Regarding the PR itself, I from my side consider it complete, but I'll remain available for fixing issues from code reviews or tests.

Thanks for your support!

CharliePoole:
@OsirisTerje I was just trying to figure out whether the use case (base assembly in a different directory) can arise in normal use. I guess it would require the user to define two different output directories and turn off local copy... odd but possible.

@gimpf Re-reviewing!

CharliePoole:
This still looks good to me.

rprouse:
This looks good to me too. @OsirisTerje shall we merge? I will need to merge this into my PR.

OsirisTerje:
@rprouse   Yep, looks fine to me too.

kantnisha:
Is this change released?

rprouse:
@kantnisha yes, this is in the current released version.

kantnisha:
thanks

kantnisha:
Does this work on VS2017? Anyone tried?

OsirisTerje:
See comments on #314 and also try the repro here on the NUnit3 branch, https://github.com/OsirisTerje/NUnitTestAdapterIssue143  
Ref @CharliePoole comments on #314 you might need to tweak the repro to check your exact case.  
And yes, this should work on VS2017 and earlier. 
Do you have any issues related to this?

PR #318, opened 2017-04-27
TestExecutor: Fix exception handling, refactor a little

### LINKED ISSUE #259: An exception occurred while invoking executor 'executor://nunit3testexecutor/'
When I try to run my tests from Visual Studio, since a few days, I get the error message:

```
------ Run test started ------
NUnit Adapter 3.0.10.0: Test execution started
Running selected tests in D:\PROJ_DIR\bin\Debug\Testing\Core.UnitTest.dll
An exception occurred while invoking executor 'executor://nunit3testexecutor/': Failed to resolve assembly: 'Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=11.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
========== Run test finished: 0 run (0:00:00.453) ==========
```
This issue suddenly appeared. At least I cannot remember that I updated something on my system.

The used VS version is 2015 Update 3. I can see the same behavior in VS15 Preview 5.

Any help to diagnose / fix / work around the issue would be appreciated.

Note: This issue doesn't happen in the 3.5.1 version of the test adapter, but as my test currently crash with app domain unload exceptions in that version, I cannot upgrade yet.

gimpf:
Sometimes my test-execution would be interrupted, that is, only 500 of 1000 test-cases would run. The only note would be something about failing to unload an app-domain. Turns out, running _all_ tests was fine, but selecting a few and then only running those caused the problem.

The commit in this PR unifies the behavior w.r.t. exceptions for both kinds of test-execution. Please see the commit message for details.

I can provide a test-case, but it would cause the VS test executor to hang and keep the assembly locked (so without killing the VS executor task, later builds would fail). Since VS does kill the executor process when "Cancel" is triggered, there might be a way out of that problem, but I think that's a separate issue, and I haven't investigated yet.

In case you want to see it for yourself, add a test-case (that needs to be run _first_) in a separate test assembly which does nothing more but

```cs
new Thread(() => { try { } finally { Thread.Sleep(TimeSpan.FromHours(1)); } }).Start()
```

Then execute all tests, but not via "Run All", but by selecting all test cases (so that the `RunTests()` overload given an `IEnumerable<TestCase>` is invoked). Around 10-30 seconds after this bad-kitty test-case finished, the CLR will give up unloading the app-domain, throw the exception, and the remaining tests will not be excuted.

Fixes #259

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestExecutor.cs
+++ b/src/NUnitTestAdapter/NUnit3TestExecutor.cs
@@ -42,7 +42,6 @@ namespace NUnit.VisualStudio.TestAdapter
     [ExtensionUri(ExecutorUri)]
     public sealed class NUnit3TestExecutor : NUnitTestAdapter, ITestExecutor, IDisposable
     {
-
         // Fields related to the currently executing assembly
         private ITestRunner _activeRunner;
 
@@ -95,21 +94,17 @@ namespace NUnit.VisualStudio.TestAdapter
                     if (!Path.IsPathRooted(assemblyName))
                         assemblyName = Path.Combine(Environment.CurrentDirectory, assemblyName);
 
-                    TestLog.Info("Running all tests in " + assemblyName);
-
                     RunAssembly(assemblyName, TestFilter.Empty);
                 }
                 catch (Exception ex)
                 {
-                    if (ex is TargetInvocationException)
-                        ex = ex.InnerException;
+                    if (ex is TargetInvocationException) { ex = ex.InnerException; }
                     TestLog.Warning("Exception thrown executing tests", ex);
                 }
             }
 
             TestLog.Info(string.Format("NUnit Adapter {0}: Test execution complete", AdapterVersion));
             Unload();
-
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -95,21 +94,17 @@ public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrame
                     if (!Path.IsPathRooted(assemblyName))
                         assemblyName = Path.Combine(Environment.CurrentDirectory, assemblyName);
 
-                    TestLog.Info("Running all tests in " + assemblyName);
```
CharliePoole: As it happens, I agree with this change but, in general, please don't make unrelated changes when fixing a problem. If you feel that there are too many messages, that's really a separate issue and mixing different issues in one PR means it only gets merged if the reviewers agree with __all__ of them.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -95,21 +94,17 @@ public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrame
...
-
                     RunAssembly(assemblyName, TestFilter.Empty);
                 }
                 catch (Exception ex)
                 {
-                    if (ex is TargetInvocationException)
-                        ex = ex.InnerException;
+                    if (ex is TargetInvocationException) { ex = ex.InnerException; }
```
CharliePoole: Please keep to our layout rather than making changes purely for your own style preferences.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -209,14 +207,18 @@ private void RunAssembly(string assemblyName, TestFilter filter)
...
+
             // No need to restore if the seed was in runsettings file
             if (!Settings.RandomSeedSpecified)
                 Settings.RestoreRandomSeed(Path.GetDirectoryName(assemblyName));
 
-            _activeRunner = GetRunnerFor(assemblyName);
-
             try
             {
+                _activeRunner = GetRunnerFor(assemblyName);
```
rprouse: This and your finally clause below appear to be the fix for the issue, but as @CharliePoole says, you have included numerous unrelated changes. This code looks fine and the unrelated `TestLog` changes are probably fine, but they are unrelated, so it makes it harder to accept this PR.

rprouse:
@gimpf I also added a Fixes #259 to your description so it will auto-close the issue. I noticed that you linked to this PR from the issue, but it is best to link to the issue from the PR so that it will be automatically closed when the PR is merged.

This is your first PR with us, so please don't be discouraged by our comments. We appreciate the work you have done, but we have to review so many issues and PRs every day that anything you can do to make our job easier helps to get your work merged quicker. We're all volunteers doing this on limited spare time 😄

CharliePoole:
I'm going to merge as the unrelated changes are minimal and more good than bad. I just wanted @gimpf to know that we prefer to see the code in a PR address a single point. Among other things, it makes them go through more smoothly.

Thanks for the work!

gimpf:
@CharliePoole Thanks for merging, but I would have split it up, and reformatted, no problem. As a clarification: Do you want such changes as separate _commits_, or as separate _PRs_?

rprouse:
@gimpf no worries this time. In the future, please create separate PRs for each fix/enhancement.

Thanks

Issue #338, opened 2017-06-09
Unhandled Exception when running through vstest.console.exe /listtests


JakenVeina:
`vstest.console.exe [AssemblyName] /ListTests /UseVsixExtensions`

produces

> Information: NUnit Adapter 3.8.0.0: Test discovery starting
> 
> Warning: Exception System.ArgumentException, Exception thrown discovering tests
> in CommonUtilitiesTests.dll
> 
> Warning: Path cannot be the empty string or all whitespace.
> 
> Information: NUnit Adapter 3.8.0.0: Test discovery complete

I traced the exception to NUnitTestAdapter.cs, line 217.

```C#
var workDir = Settings.WorkDirectory;
if (workDir == null)
    workDir = Path.GetDirectoryName(assemblyName);
else if (!Path.IsPathRooted(workDir))
    workDir = Path.Combine(Path.GetDirectoryName(assemblyName), workDir);
if (!Directory.Exists(workDir))
    Directory.CreateDirectory(workDir);
package.Settings[PackageSettings.WorkDirectory] = workDir;
```

In this case, Settings.WorkDirectory returns null, which takes us into the first if statement, where Path.GetDirectoryName(assemblyName); is called for assemblyName "MyTestAssembly.dll", and returns an empty string, which then throws an exception when passed to Directory.CreateDirectory(workDir);.

Since PackageSettings.cs line 57 specifies that the WorkDirectory setting should be a full path, I implemented a fix in PR #339 that ensures workDir is always a full path by the time the Directory.Exists() check occurs.

CharliePoole:
I don't understand why/how you have a value of "MyTestAssembly.dll" for assemblyName. I've only ever seen - and the code is written assuming - a complete rooted path passed to the adapter.

JakenVeina:
I'd guess that value ultimately comes from the [AssemblyName] command line argument to vstest.console.exe. In this case, I'm running the executable directly from the directory containing the assembly.

CharliePoole:
Fair enough. The code was originally written with the IDE in mind, which always passes in full paths.

OsirisTerje:
PR merged, so closing the issue

PR #339, opened 2017-06-09
Ensure PackageSettings.WorkDirectory is always a full path

JakenVeina:
Fix for Issue #338.

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -68,19 +68,21 @@ namespace NUnit.VisualStudio.TestAdapter
 
             foreach (string sourceAssembly in sources)
             {
+                var sourceAssemblyPath = Path.IsPathRooted(sourceAssembly) ? sourceAssembly : Path.Combine(Directory.GetCurrentDirectory(), sourceAssembly);
+
                 TestLog.Debug("Processing " + sourceAssembly);
 
                 // Only save if seed is not specified in runsettings
                 // This allows workaround in case there is no valid
                 // location in which the seed may be saved.
                 if (!Settings.RandomSeedSpecified)
-                    Settings.SaveRandomSeed(Path.GetDirectoryName(sourceAssembly));
+                    Settings.SaveRandomSeed(Path.GetDirectoryName(sourceAssemblyPath));
 
                 ITestRunner runner = null;
 
                 try
                 {
-                    runner = GetRunnerFor(sourceAssembly);
+                    runner = GetRunnerFor(sourceAssemblyPath);
 
                     XmlNode topNode = runner.Explore(TestFilter.Empty);
 
@@ -90,7 +92,7 @@ namespace NUnit.VisualStudio.TestAdapter
 
                     if (topNode.GetAttribute("runstate") == "Runnable")
                     {
-                        var testConverter = new TestConverter(TestLog, sourceAssembly);
+                        var testConverter = new TestConverter(TestLog, sourceAssemblyPath);
 
                         int cases = ProcessTestCases(topNode, discoverySink, testConverter);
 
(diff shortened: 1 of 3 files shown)


src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -210,9 +210,9 @@ private TestPackage CreateTestPackage(string assemblyName)
             // Set the work directory to the assembly location unless a setting is provided
             var workDir = Settings.WorkDirectory;
             if (workDir == null)
-                workDir = Path.GetDirectoryName(assemblyName);
+                workDir = Path.GetDirectoryName(Path.GetFullPath(assemblyName));
             else if (!Path.IsPathRooted(workDir))
-                workDir = Path.Combine(Path.GetDirectoryName(assemblyName), workDir);
+                workDir = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(assemblyName), workDir));
```
CharliePoole: Shouldn't this be

```C#
workdir = Path.Combine(Path.GetDirectoryName(Path.GetFullPath(assemblyame)), workDir);
```

or refactoring the whole thing

```C#
var assemblyDir = Path.GetDirectoryName(Path.GetFullName(assemblyName));
var workDir = Settings.WorkDirectory ?? assemblyDir;
if (!Path.IsPathRooted(workDir))
    workDir = Path.Combine(assemblyDir, workDir);
```

JakenVeina:
Given the conversation over in #338, it's probably better to bump up to NUnit3TestDiscoverer and NUnit3TestExecutor and wrap all source references in Path.GetFullPath() there, immediately after receiving them from an external entity. Alternatively, in NUnitTestAdapter.GetRunnerFor() and NUnitTestAdapter.CreateTestPackage.()

CharliePoole:
I agree. Do you want to modify your PR to do that?

I'd do it at the earliest possible moment, so that all methods called with an assembly name get the full path.

JakenVeina:
I think that covers all the entry points.

For clarity, the NUnit.VisualStudio.TestAdapter.Tests namespace is the only one that contains actual tests, yes? The rest are just fakes/mocks?

CharliePoole:
Correct... Are you running under the adapter? You can't test the adapter under the adapter in the IDE because VS tries to run all tests in the mock-assembly.

In fact we don't use the adapter for testing any of the NUnit modules - at least I don't.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -68,19 +68,21 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
 
             foreach (string sourceAssembly in sources)
             {
+                var sourceAssemblyPath = Path.IsPathRooted(sourceAssembly) ? sourceAssembly : Path.Combine(Directory.GetCurrentDirectory(), sourceAssembly);
```
CharliePoole: This will work so long as we are sure that any relative path we get is relative to the current directory at the time of running the test. We never change it, and hopefully VS doesn't do that either.

CharliePoole:
OK... hang on... I just noticed that the second commit of two has the description "Reverts last commit". However it doesn't, since there are changes still in the file. Can you explain please?

JakenVeina:
Bad choice of phrasing on my part. I didn't actually revert the commit itself in git, I just changed the code back to what it used to be. 

If you'd rather I actually revert the commit, I can go learn how to do that. I've been told it's cumbersome.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
JakenVeina: I was originally using Path.GetFullPath(), but that introduces new error catching concerns, as it requires that the given file actually exists, which hasn't been verified yet. I spotted the "combine with current directory" method in a couple spots already in the code and it seemed like a better way to do it, since it leaves validation to the error-handling code that's already in place.

CharliePoole:
Thanks... that explains it. We'll continue to wait for the second review.

OsirisTerje:
Have been through the code, and it looks good to me too.

Issue #298, opened 2017-01-15
$RANDOM_SEED$ is appearing in non-test project build output


CharliePoole:
@jnm2 commented on [Sat Jan 14 2017](https://github.com/nunit/nunit/issues/1982)

This is messing up the installer generated by CI. Now I _can_ update my build script to find and exclude those files, but they should not be appearing at all because there are no test assemblies in those folders.

---

@CharliePoole commented on [Sat Jan 14 2017](https://github.com/nunit/nunit/issues/1982#issuecomment-272667441)

What installer, for what project?

This file has nothing to do with NUnit but is created by the NUnit3 adapter. There is an adapter issue that renamed this file so you must be using the last adapter release to get that name. Anyway, the file will appear in the output directory for any nunit3 test assembly. It should not be created for other assemblies that don't share the same output directory. If it does, that's an adapter bug.

Do I need to move this or just close it?

---

@jnm2 commented on [Sat Jan 14 2017](https://github.com/nunit/nunit/issues/1982#issuecomment-272667854)

Oh, I didn't know this was an adapter thing. Can I turn it off? If the console runner doesn't need it, why do I want it?

> Do I need to move this or just close it?

Depends- is the fact that it's showing up in the build output of a non-test project a  bug?

---

@CharliePoole commented on [Sat Jan 14 2017](https://github.com/nunit/nunit/issues/1982#issuecomment-272671755)

Erase it and figure out what you do that creates it. It should only be created by running the NUnit3 adapter on that project. Of course, if you have it installed as a vsix, it is run every time you open the project.

---

@CharliePoole commented on [Sat Jan 14 2017](https://github.com/nunit/nunit/issues/1982#issuecomment-272671808)

Does the output directory contain any test projects, in addition to the non-test project?

CharliePoole:
Examining the code, I see that the random seed file is saved when discovery starts. Discovery is actually run on every assembly, as directed by VS, so we don't know if it's a test assembly until discovery ends. That's when we should save it.

jnm2:
The output directory does not contain any DLLs that reference a testing framework.

jnm2:
I have files from 2 pm yesterday in both Debug and Release, so I know those are from IDE. For some reason my Cake script which also uses the VS adapter is not generating new ones.

[...]

PR #346, opened 2017-06-27
Moved code for creating the random seed file 

OsirisTerje:
… Moved code for creating the random seed file to after we have checked… if  it is a test project, and that there are any tests at all.  #298

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -72,12 +72,6 @@ namespace NUnit.VisualStudio.TestAdapter
 
                 TestLog.Debug("Processing " + sourceAssembly);
 
-                // Only save if seed is not specified in runsettings
-                // This allows workaround in case there is no valid
-                // location in which the seed may be saved.
-                if (!Settings.RandomSeedSpecified)
-                    Settings.SaveRandomSeed(Path.GetDirectoryName(sourceAssemblyPath));
-
                 ITestRunner runner = null;
 
                 try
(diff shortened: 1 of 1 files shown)


OsirisTerje:
@jnm2 Good if you can check if this now works on your code.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -97,6 +91,11 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
                         int cases = ProcessTestCases(topNode, discoverySink, testConverter);
 
                         TestLog.Debug(string.Format("Discovered {0} test cases", cases));
+                        // Only save if seed is not specified in runsettings
+                        // This allows workaround in case there is no valid
+                        // location in which the seed may be saved.
+                        if (cases>0 && !Settings.RandomSeedSpecified)
+                            Settings.SaveRandomSeed(Path.GetDirectoryName(sourceAssemblyPath));
```
CharliePoole: Have you verified that the random seed saved at this point is the same one saved initially? I think that's the case, but just want to be sure.  Also, do we have random tests?

OsirisTerje: The code was copied over, no changes otherwise.   It will generate different values for each run, but that's the same as it always has been.  
There is a test for the random seed, and for the reading of that from the runconfig, but not any for running with these.   It would be good to wrap the discoverer and the executor with tests around them, but that would require mocking of the connections or using some fake assemblies.  The older tests we had for the AssemblyRunner are commented out.  There is a TODO there to convert "as needed".    This could be improved, but I think that is beside this fix, although it would have benefited from it.

CharliePoole:
I see the code is identical. What I am asking is whether the value returned remains the same when the code is moved. If not, we would have to fetch it, save it in a member and then later write out the file. If it's not the same, then it would break any random tests when run under VS. That's why I wondered whether we have any.

Looking at the code, it seems like the property we return should be the same at both points. I'm just not in a position to run it right now.

OsirisTerje:
I've added a test in  my issue repo (https://github.com/OsirisTerje/NUnit.Issues/tree/NUnit3)   to check that random still works, and it does. 
So all is well :-)  I'll merge this then.

Issue #332, opened 2017-05-21
Add Test Attachment functionality


ChrisMaddock:
See https://github.com/nunit/nunit/issues/1670

kdubau:
Is this available in any preview yet?

ChrisMaddock:
@kdubau - if, like in https://github.com/nunit/nunit/issues/1670#issuecomment-303217309, you're just interested in the entry being in the XML, that was just released in NUnit Framework 3.7. 🙂 I'm afraid I haven't got round to writing any docs yet, but just use `TestContext.AddTestAttachment(string filepath, string description = null)`.

Integration with the VS adapter I'm afraid I haven't got round to yet.

ChrisMaddock:
In terms of timescales, it's probably going to be a few weeks before I can get on to it - I can only really get stuff like this done at weekends, and have a busy few coming up. More than happy for someone else to take over this part, and do the PR though. 😄 I'll unassign myself for now to show that, although will come back to it in time, if nobody else wants to do it!

kdubau:
Hey, @ChrisMaddock - yep I found the source for adding the attachment to XML. I was just being greedy by looking for it in VS test explorer 😛. Perhaps I can take a look at this adapter integration if i'm feeling ambitious, but it'd be my first experience in the realm of VS extension development.

CharliePoole:
@kdubau You would even be helping us quite a bit if you outlined what you want to see in the adapter!

kdubau:
I was really just looking for something like the screenshot here: https://github.com/nunit/nunit/issues/1670#issuecomment-232896282

ChrisMaddock:
Please feel free, @kdubau 👍 

@CharliePoole @kdubau - I was hoping to just emulate the MSTest adapter, and use the Visual Studio interface for attachments like in that screenshot. I did a little digging, the Visual studio TestResult object has an Attachments property, which holds and Attachment set. Hopefully, we can interface with it in the same way as the MSTest adapter does [here ](https://github.com/Microsoft/testfx/blob/a0cfbcffe9f292064bf57d784eae9e2bc3ea2103/src/Adapter/MSTest.CoreAdapter/ObjectModel/UnitTestResult.cs#L178) - I haven't tried it out yet though!

CharliePoole:
@ChrisMaddock Amazing! I had no idea that the MSTest adapter was open source. And it's even MIT-licensed, so I don't have to be afraid to look at it! 😸

[...]

PR #347, opened 2017-06-28
Add support for file attachments

Daniel-Svensson:
* Update nunit version to 3.7.1 for tests
* Add tests for 1 and 2 attachments
* Improve test resilency to changes so that we don't have to update
  test count att multiple locations
* Replace broken test ExplicitTestDoesNotShowUpInResults (which did not work)
  with an extra testcase for TestResultIsReportedCorrectly

Should hopefully solve #332 

### Summary of review comments 

- [x] Update code to C#7 (nameof and string interpolation)
- [x] Update remaining tests to nuget 3.7.1
- [x] Move attachment xml parsing into a separate method
- [x] Remove redundant code in TestExecutionTests
   * new List
   * typecast
- [x] Change asserts to fluent syntax
- [x] Don't create files when running tests
- [ ] Add some additional tests (as unit tests using raw xml)
  I guess these will live in TestConverterTests
     * xml with empty/no description
     * failure/warning paths (bad & missing paths)
     * Probably move one of the mock tests and make into unit test

### DIFF
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -199,6 +199,35 @@ namespace NUnit.VisualStudio.TestAdapter
             if (outputNode != null)
                 vsResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, outputNode.InnerText));
 
+            var attachments = resultNode.SelectNodes("attachments/attachment");
+            if (attachments.Count > 0)
+            {
+                var attachmentSet = new AttachmentSet(new Uri(NUnitTestAdapter.ExecutorUri), "Attachments");
+                vsResult.Attachments.Add(attachmentSet);
+
+                foreach (XmlNode attachment in attachments)
+                {
+                    var path = attachment.SelectSingleNode("filePath")?.InnerText ?? string.Empty;
+                    var description = attachment.SelectSingleNode("description")?.InnerText;
+
+                    try
+                    {
+                        // We only support absolute paths since we dont lookup working directory here
+                        // any problem with path will throw an exception
+                        var fileUri = new Uri(path, UriKind.Absolute);
+                        attachmentSet.Attachments.Add(new UriDataAttachment(fileUri, description));
+                    }
+                    catch (UriFormatException ex)
+                    {
+                        _logger.Warning(string.Format("Ignoring attachment with path '{0}' due to problem with path: {1}", path, ex.Message));
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.Warning(string.Format("Ignoring attachment with path '{0}': {1}.", path, ex.Message));
+                    }
+                }
+            }
+
             return vsResult;
         }
 
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -199,36 +199,46 @@ namespace NUnit.VisualStudio.TestAdapter
             if (outputNode != null)
                 vsResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, outputNode.InnerText));
 
-            var attachments = resultNode.SelectNodes("attachments/attachment");
-            if (attachments.Count > 0)
-            {
-                var attachmentSet = new AttachmentSet(new Uri(NUnitTestAdapter.ExecutorUri), "Attachments");
+            var attachmentSet = ParseAttachments(resultNode);
+            if (attachmentSet.Attachments.Count > 0)
                 vsResult.Attachments.Add(attachmentSet);
 
-                foreach (XmlNode attachment in attachments)
-                {
-                    var path = attachment.SelectSingleNode("filePath")?.InnerText ?? string.Empty;
-                    var description = attachment.SelectSingleNode("description")?.InnerText;
+            return vsResult;
+        }
 
-                    try
-                    {
-                        // We only support absolute paths since we dont lookup working directory here
-                        // any problem with path will throw an exception
-                        var fileUri = new Uri(path, UriKind.Absolute);
-                        attachmentSet.Attachments.Add(new UriDataAttachment(fileUri, description));
-                    }
-                    catch (UriFormatException ex)
-                    {
-                        _logger.Warning(string.Format("Ignoring attachment with path '{0}' due to problem with path: {1}", path, ex.Message));
-                    }
-                    catch (Exception ex)
-                    {
-                        _logger.Warning(string.Format("Ignoring attachment with path '{0}': {1}.", path, ex.Message));
-                    }
+        /// <summary>
+        /// Looks for attachments in a results node and if any attachments are found they
+        /// are added to <paramref name="vsResult"/>
+        /// </summary>
+        /// <param name="resultNode">xml node for test result</param>
+        /// <returns>attachments to be added to the test, it will be empty if no attachments are found</returns>
+        private AttachmentSet ParseAttachments(XmlNode resultNode)
+        {
+            var attachmentSet = new AttachmentSet(new Uri(NUnitTestAdapter.ExecutorUri), "Attachments");
+
+            foreach (XmlNode attachment in resultNode.SelectNodes("attachments/attachment"))
+            {
+                var path = attachment.SelectSingleNode("filePath")?.InnerText ?? string.Empty;
+                var description = attachment.SelectSingleNode("description")?.InnerText;
+
+                try
+                {
+                    // We only support absolute paths since we dont lookup working directory here
+                    // any problem with path will throw an exception
+                    var fileUri = new Uri(path, UriKind.Absolute);
+                    attachmentSet.Attachments.Add(new UriDataAttachment(fileUri, description));
+                }
+                catch (UriFormatException ex)
+                {
+                    _logger.Warning($"Ignoring attachment with path '{path}' due to problem with path: {ex.Message}");
+                }
+                catch (Exception ex)
+                {
+                    _logger.Warning($"Ignoring attachment with path '{path}': {ex.Message}.");
                 }
             }
 
-            return vsResult;
+            return attachmentSet;
         }
 
         // Public for testing
--- a/src/NUnitTestAdapterTests/TestDiscoveryTests.cs
+++ b/src/NUnitTestAdapterTests/TestDiscoveryTests.cs
@@ -65,7 +65,7 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
         public void LoadMockassembly()
         {
             // Sanity check to be sure we have the correct version of mock-assembly.dll
-            Assert.That(NUnit.Tests.Assemblies.MockAssembly.Tests, Is.EqualTo(34),
+            Assert.That(NUnit.Tests.Assemblies.MockAssembly.TestsAtRuntime, Is.EqualTo(NUnit.Tests.Assemblies.MockAssembly.Tests),
                 "The reference to mock-assembly.dll appears to be the wrong version");
 
             TestCases = new List<TestCase>();
--- a/src/NUnitTestAdapterTests/TestExecutionTests.cs
+++ b/src/NUnitTestAdapterTests/TestExecutionTests.cs
@@ -54,16 +54,20 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
             MockAssemblyPath = Path.Combine(TestContext.CurrentContext.TestDirectory, "mock-assembly.dll");
 
             // Sanity check to be sure we have the correct version of mock-assembly.dll
-            Assert.That(MockAssembly.Tests, Is.EqualTo(34),
+            Assert.That(MockAssembly.TestsAtRuntime , Is.EqualTo(MockAssembly.Tests),
                 "The reference to mock-assembly.dll appears to be the wrong version");
             new List<TestCase>();
-            testResults = new List<TestResult>();
             testLog = new FakeFrameworkHandle();
 
             // Load the NUnit mock-assembly.dll once for this test, saving
             // the list of test cases sent to the discovery sink
             executor = ((ITestExecutor) new NUnit3TestExecutor());
             executor.RunTests(new[] { MockAssemblyPath }, Context, testLog);
+
+            testResults = testLog.Events
+               .Where(e => e.EventType == FakeFrameworkHandle.EventType.RecordResult)
+               .Select(e => e.TestResult)
+               .ToList();
             this.Summary = new ResultSummary(testResults);
         }
 
@@ -134,6 +138,7 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
         [TestCase("FailingTest", TestOutcome.Failed, "Intentional failure", true)]
         [TestCase("TestWithException", TestOutcome.Failed, "System.Exception : Intentional Exception", true)]
         // NOTE: Should Inconclusive be reported as TestOutcome.None?
+        [TestCase("ExplicitlyRunTest", TestOutcome.None, null, false)]
         [TestCase("InconclusiveTest", TestOutcome.None, "No valid data", false)]
         [TestCase("MockTest4", TestOutcome.Skipped, "ignoring this test method for now", false)]
         // NOTE: Should this be failed?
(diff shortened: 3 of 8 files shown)


src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -199,6 +199,35 @@ private VSTestResult GetBasicResult(XmlNode resultNode)
             if (outputNode != null)
                 vsResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, outputNode.InnerText));
 
+            var attachments = resultNode.SelectNodes("attachments/attachment");
+            if (attachments.Count > 0)
+            {
+                var attachmentSet = new AttachmentSet(new Uri(NUnitTestAdapter.ExecutorUri), "Attachments");
```
Daniel-Svensson: "Attachments" name is hardcoded, feel free to change it if you find any other name more suitable such as "nunit attachments" or similar

src/NUnitTestAdapterTests/TestExecutionTests.cs
```diff
@@ -54,16 +54,20 @@ public void LoadMockassembly()
...
 
             // Load the NUnit mock-assembly.dll once for this test, saving
             // the list of test cases sent to the discovery sink
             executor = ((ITestExecutor) new NUnit3TestExecutor());
             executor.RunTests(new[] { MockAssemblyPath }, Context, testLog);
+
+            testResults = testLog.Events
+               .Where(e => e.EventType == FakeFrameworkHandle.EventType.RecordResult)
+               .Select(e => e.TestResult)
+               .ToList();
```
Daniel-Svensson: testResults was always an empty list in previous implementation, that is how to ExplicitTestDoesNotShowUpInResults managed to pass even if the test was recorded with outcome None

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -199,6 +199,35 @@ private VSTestResult GetBasicResult(XmlNode resultNode)
...
+                vsResult.Attachments.Add(attachmentSet);
+
+                foreach (XmlNode attachment in attachments)
+                {
+                    var path = attachment.SelectSingleNode("filePath")?.InnerText ?? string.Empty;
+                    var description = attachment.SelectSingleNode("description")?.InnerText;
```
ChrisMaddock: Did you check that null is an acceptable value for the description passed into `UriDataAttachment`? It would be good to add a test with a null description, just in case of regression. 🙂

src/mock-assembly/MockAssembly.cs
```diff
@@ -315,4 +321,43 @@ public class GrandChildClass
...
+        {
+            var context = TestContext.CurrentContext;
+            var filepath1 = Path.Combine(context.WorkDirectory, Attachment1Name);
+            var filepath2 = Path.Combine(context.WorkDirectory, Attachment2Name);
+
+            File.WriteAllText(filepath1, Attachment1Contents);
+            TestContext.AddTestAttachment(filepath1, Attachment1Description);
+
+            File.WriteAllText(filepath2, Attachment2Contents);
+            TestContext.AddTestAttachment(filepath2, Attachment2Description);
```
ChrisMaddock: The files created here should be cleaned up at some stage. It feels a little mismatched to create them here, and clean them up at the end of TestExecutionTests - perhaps they should be created and removed in TestExecutionTests? I'm afraid I'm not familiar enough with the adapter tests to know what's best!

src/NUnitTestAdapterTests/TestExecutionTests.cs
```diff
@@ -153,9 +158,48 @@ public void TestResultIsReportedCorrectly(string name, TestOutcome outcome, stri
         }
 
         [Test]
-        public void ExplicitTestDoesNotShowUpInResults()
+        public void AttachmentsShowUpInResults()
         {
-            Assert.Null(testResults.Find(r => r.TestCase.DisplayName == "ExplicitlyRunTest"));
+            TestResult test = testResults.Find(r => r.TestCase.DisplayName == "SingleAttachmentTest");
```
ChrisMaddock: Minor: This and the below could perhaps use `nameof()` - in case of future refactoring. 🙂

ChrisMaddock:
Just to transfer @Daniel-Svensson's comment from #332 - someone should check this out in VS at least, before it's merged. 😄

src/NUnitTestAdapter/TestConverter.cs
CharliePoole: Not a problem. All the names here are hard-coded. And attachments/attachment seems the obvious choice to me.

src/NUnitTestAdapterTests/TestExecutionTests.cs
```diff
@@ -153,9 +158,48 @@ public void TestResultIsReportedCorrectly(string name, TestOutcome outcome, stri
         }
 
         [Test]
-        public void ExplicitTestDoesNotShowUpInResults()
```
CharliePoole: Are you actually removing a test here? I can't find it moved anywhere else.

src/NUnitTestAdapterTests/TestDiscoveryTests.cs
```diff
@@ -65,7 +65,7 @@ public TestDiscoveryTests(IDiscoveryContext context)
         public void LoadMockassembly()
         {
             // Sanity check to be sure we have the correct version of mock-assembly.dll
-            Assert.That(NUnit.Tests.Assemblies.MockAssembly.Tests, Is.EqualTo(34),
+            Assert.That(NUnit.Tests.Assemblies.MockAssembly.TestsAtRuntime, Is.EqualTo(NUnit.Tests.Assemblies.MockAssembly.Tests),
```
CharliePoole: I understand why this works, but the hard-coded number gives me a lot more confidence. That said, we may not need the check at all. it was inherited from the framework project, which had several different versions of mock-assembly.

src/mock-assembly/MockAssembly.cs
CharliePoole: I find this a funny location for the tests themselves. It feels like a kind of system test, ensuring that the framework, engine and adapter work together to display attachment info. But these are unit tests. I would lean toward starting with XML that we expect to come from nunit and making sure we deal with it correctly.

CharliePoole:
@ChrisMaddock FWIW I tend to wait for @OsirisTerje to approve, no matter how many of us have reviewed it. He hasn't given any other instructions and it would make sense to me if he wants to stay on top of new features himself.

src/NUnitTestAdapter/TestConverter.cs
Daniel-Svensson: I had a look at https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/AttachmentSet.cs and https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.testtools.common.uridataattachment.uridataattachment.aspx but did not find any mention of it being required. At least it does not seem to throw

src/NUnitTestAdapterTests/TestExecutionTests.cs
Daniel-Svensson: yes it is remove (replaced with the additional Testcase for TestResultIsReportedCorrectly) 

The old test was flawed (it worked only because testResults where never initialized and always empty)

OsirisTerje:
I'm grabbing this down now, and will have a look.  First impression is good!

src/mock-assembly/MockAssembly.cs
Daniel-Svensson: I am a bit hesitant to check in xml at the moment, since I don't know if the format is set in stone yet. nunit/docs#197 and nunit/nunit#1670 are still open but that is no problem?

The current approach was taken because I did not know nunit good enough to get the xml to persist.
I do believe it is valuable to have end-2-end tests for this functionality in case anything changes, but maybe it also makes sense with some more traditional unit tests such as without description etc?

1. For the files, it might make more sense to just use reference assemblies instead that whey no files need to be created or removed. What do you think?
2. For some extra unit tests it should be easy to add a new testfixture with new test but for the existing tests is there some recommended approach how to test the whole chain (running tests -> nunit xml -> TestResult) if they do not belong where they are?

ChrisMaddock: Please consider the xml to be final (or at least, to be something we would make backwards compatible!). That part was released as part of NUnit Framework 3.7 - the remaining issues are only still open as I've been a little tied up. Now you've sorted this one, and it's just the docs left, I'll try and do it soon. :-)

I'll let Charlie comment on the tests.

OsirisTerje:
Something to start with :
1.  Good you updated to NUnit 3.7.1.  Can you update the remaining 2 (Projects Empty-assemblies and NUnit3AdapterExternalTests) too, so they are consolidated?
2. The method GetBasicResult in the TestConverter now starts to become pretty long. I think it would be nice if it was broken up.  You could move the attachment parts into its own private method. 

Have added some more comments "in place" with the file changes.

OsirisTerje:
But all in all, this is good work !

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -199,6 +199,35 @@ private VSTestResult GetBasicResult(XmlNode resultNode)
             if (outputNode != null)
                 vsResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, outputNode.InnerText));
 
+            var attachments = resultNode.SelectNodes("attachments/attachment");
+            if (attachments.Count > 0)
```
Daniel-Svensson: No, attachment is always a "list"

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -199,6 +199,35 @@ private VSTestResult GetBasicResult(XmlNode resultNode)
...
+                    try
+                    {
+                        // We only support absolute paths since we dont lookup working directory here
+                        // any problem with path will throw an exception
+                        var fileUri = new Uri(path, UriKind.Absolute);
+                        attachmentSet.Attachments.Add(new UriDataAttachment(fileUri, description));
+                    }
+                    catch (UriFormatException ex)
+                    {
+                        _logger.Warning(string.Format("Ignoring attachment with path '{0}' due to problem with path: {1}", path, ex.Message));
```
Daniel-Svensson: 👍 I will use string interpolation, was unsure about coding standard for this project

Btw. I would be really nice if the readme at the root would be updated with some minor details such as requireing VS 2017 for development etc.  If was after writing this that I saw that the new sdk msbuild project was used and it was safe to use new C# functionality.

Daniel-Svensson:
I've updated the initial description with a summary of the changes to be made as I understand it. 
If anybody has comments regarding it (such as if I missed anything please let me know)

src/NUnitTestAdapter/TestConverter.cs
CharliePoole: C#6 is the standard. Personally, I hate string interpolation. It makes me feel like I'm using a scripting language. But I wouldn't tell you not to. 😄

jnm2: If C# 6 is the standard, all the .csproj files should have `<LangVersion>6</LangVersion>`. I guess we never did that for this repo?
I personally would like to leave the C# version restriction up to the project lead of each repo, but whatever the restriction is, it absolutely should be in the csproj.
I also didn't know there were MSBuild 15-only .csprojs in this repo, but again I think that should be at the discretion of the project lead. @OsirisTerje can you document the VS2017 requirement though?

CharliePoole: Strictly speaking, I should have said "has been the standard" up to now. The core team has to eventually decide what, if any, standards are imposed on projects. Meanwhile, I'd say it's up to the VS team lead and team.

As a general observation, I don't think we should rush allowing in newer standards or requiring newer versions of VS - in this case requiring VS itself is OK, since it;s an extension. But the changes that have been made to the extensions as well as the console have already blocked me from working on them effectively. That is (dare I say) a loss.

jnm2: VS2017 is free and once the SDK teams figure an architecture out, MSBuild and the SDKs (which are all open source) will be able to be bin deployed from NuGet rather than installed and used directly a la Cake without any version of VS installed at all. C# 7 is freely available to anyone in VS2015 as well if we are willing to reference the roslyn compilers as design-time project NuGet dependencies.

CharliePoole: Those are all reasons why I should not have stopped working on the projects accoding to you - and yet I did.

Some people prefer to use some things. You want to prioritize those people who prefer the "new thing." I want to take the conservative approach. This may be a basic personality disorder on the part of one of us. 😄 

When I was a very young developer in my 50s I thought that those folks who used the latest stuff were somehow better than those who don't. It's a typical attitude and very like the attitude at some large software companies. The message from NUnit has always been that we value users and contributors even where circumstances force them to use older releases of .NET, of VS or some open source IDE. I already (in our governance discussions) predicted this would change and I'm seeing it happen. It may be endemic to the .NET world itself.

If all that stuff were really available as Open Source __now__ I might see it differently. Anyway, it's not something I need to have a future say in.

OsirisTerje:
Great work!

src/NUnitTestAdapter/TestConverter.cs
OsirisTerje: I'm part of in the middle here I think, I like new stuff, but one thing I like with NUnit is the backwards compatibility.  It makes it possible for industrial software, which I have been working on all my life, more or less, to use the stuff, they often cant upgrade as often as web sites or the like.   Now backwards compatibility have a limit though, at some points it becomes too hard and costly.
For this extension, it is a part of VS2017 (and former versions) but it has to run on the latest VS - and thus I see no problems staying with the latest standard there either, so I am happy about C# 7.  
One comment also on @CharliePoole's comment on string interpolation.   Contrary to Charlie, I really like it - I find the syntax much more readable, and I favour speed of reading.  My 5 cents  and my opinion of course. 
I like new things, but I don't like new things that don't really work, or is partly broken - and thus, I have built up a certain distaste for the .net core stuff.  I would love it to work out, but the fact it breaks in so many situations annoys me, and also the lack of tooling support it has.  I really hope it will improve rather soon.

OsirisTerje:
@Daniel-Svensson  I just noted the test you have in mock_assembly named AttachmentTest, which is being used in the other tests you have for attachments, are failing, because it can't locate the empty_assembly.dll (which is not present in that output directory, mock_assembly has no reference to it).  Is this by intention ?

CharliePoole:
Both mock-assembly and empty-assembly exist for particular tests of how we load assemblies and should not have tests added. Can we find another way to write these tests?

OsirisTerje:
I first thought the test was misplaced, and then tried to move it, but then saw it is being used in the   public void AttachmentsShowSupportMultipleFiles() test, which is in the right place (NUnitAdapterTests project)
Not quite sure what Daniel has intended to check here.

CharliePoole:
So far we had managed to avoid any assembly that contains "tests as data" for our real tests, with the exception of the tests of assembly loading. I think Daniel is trying to create an all-the-way-through test, which we don't really have for anything. Generally, I add test-of-tests data to the Demo assembly and verify it manually.

Daniel-Svensson:
@OsirisTerje when i compiled and ran the tests both the mock and empty assembly were copied to the output folder of the actual test project (not mock assembly output directory). doesn't that happen for you? I tried both netcoreapp and normal .net target.

As for the tests, my intention was to add some unit tests that worked on raw xml but never got so far. I won't have time for development the next couple of weeks so if anybody is interested in adding some tests please go ahead. It is possible to set a breakpoint in the converter while debugging the test in order to look at the actual xml for attachments.

OsirisTerje:
@Daniel-Svensson Thanks!   Yes, for the NUnit3TestAdapterTest project they are, but the actual test is also run in VS Test Explorer in the mock assembly, and for that one there is no empty-assembly dll, and it fails there.  It is mostly confusing, not a show stopper.   But as @CharliePoole pointed out, the mock assembly is meant as stubs for the assembly loading, so we should not place other tests there.  I'll see what I can with it post-release. :-)

PR #350, opened 2017-07-03
Added support for CollectSourceInformation flag

Shyam-Gupta:
Refer RFC: https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0010-Source-Information-For-Discovered-Tests.md
Added support for handling CollectSourceInformation flag in runsettings in NUnit VS adapter.

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -36,6 +36,8 @@ namespace NUnit.VisualStudio.TestAdapter
 
         public string TestAdapterPaths { get; private set; }
 
+        public bool CollectSourceInformation { get; private set; }
+
         #endregion
 
         #region Properties - TestRunParameters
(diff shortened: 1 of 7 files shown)


OsirisTerje:
This need some explanation (and add an issue please).  I see the changes, but I am not able to understand what this is doing.

codito:
Created issue https://github.com/nunit/nunit3-vs-adapter/issues/351 for this change.

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -101,6 +103,7 @@ public void Load(string settingsXml)
             TargetPlatform = GetInnerText(runConfiguration, "TargetPlatform");
             TargetFrameworkVersion = GetInnerText(runConfiguration, "TargetFrameworkVersion");
             TestAdapterPaths = GetInnerText(runConfiguration, "TestAdapterPaths");
+            CollectSourceInformation = GetInnerTextAsBool(runConfiguration, "CollectSourceInformation", true);
```
codito: `CollectSourceInformation` is true by default to allow for older VS to work seamlessly, they will never set the setting.

jnm2:
I like this.

Issue #358, opened 2017-07-13
Add test name pattern to the adapter runsettings


OsirisTerje:
See #326 for discussion 

Currently the test name pattern is hardcoded, including a limit of 40 chars, in the adapter.  
By adding the test name pattern, as documented here https://github.com/nunit/docs/wiki/Template-Based-Test-Naming :  , as a separate setting in the nunit section in runsettings, the user can change this as wanted.  
The default will remain as {m}{a:40} meaning test method name, max 40 chars.

CharliePoole:
While I know you understand the meaning, since this is a public post I wanted to correct the meaning of `{m}{a:40}`. It means...

The full name of the method, without truncation, followed by all the arguments in parentheses, separated by commas, with string arguments truncated to 40 chars. The total length of the name is indeterminate.

Issue #357, opened 2017-07-13
Not all generic test fixtures can be found if unqualified type names match


chwarr:
When using a [generic test fixture](https://github.com/nunit/docs/wiki/TestFixture-Attribute#generic-test-fixtures) if two of the types have the same unqualified name (e.g., Product.A.RedWidget and Product.B.RedWidget), only one of the tests will be found an executed.

In the following test, I expect two passed tests and two failed tests (total of four). Instead I get two failed tests and one passed test (total of three).

If I run the same test with nunit3-console (version 3.6.1), I see all four of the tests as expected.

    namespace Product
    {
        public interface IWidget
        {
            bool IsOK { get; }
        }
    
        namespace A
        {
            public class GoodWidget : IWidget
            {
                public bool IsOK => true;
            }
    
            public class BadWidget : IWidget
            {
                public bool IsOK => false;
            }
    
            public class RedWidget : IWidget
            {
                public bool IsOK => true;
            }
        }
    
        namespace B
        {
            public class RedWidget : IWidget
            {
                public bool IsOK => false;
            }
        }
    }
    
    namespace nunit_type_params
    {
        using Product;
        using Product.A;
    
        using NUnit.Framework;
    
        [TestFixture(typeof(GoodWidget))]
        [TestFixture(typeof(BadWidget))]
        [TestFixture(typeof(Product.A.RedWidget))]
        [TestFixture(typeof(Product.B.RedWidget))]
        public class TakesTypeParam<T> where T : IWidget, new()
        {
            [Test]
            public void IsGood()
            {
                var t = new T();
                Assert.That(t.IsOK);
            }
        }
    }

![Visual Studio Test Explorer only showing three tests instead of four](https://user-images.githubusercontent.com/10605667/28146702-9ef174fc-6730-11e7-97aa-82f68d5b1443.png)

* NUnit version 3.7.1
* NUnit3TestAdapater version 3.8.0-alpha1
* Targeting .NET Framework 4.6.1
* Visual Studio version details:
> Microsoft Visual Studio Enterprise 2015
> Version 14.0.25431.01 Update 3
> Microsoft .NET Framework
> Version 4.7.02046
>
> Installed Version: Enterprise
>
> Architecture and Modeling Tools   00322-90200-03147-AA547
> Microsoft Architecture and Modeling Tools
>
> UML® and Unified Modeling Language™ are trademarks or registered trademarks of the Object Management Group, Inc. in the United States and other countries.
>
[...]

PR #359, opened 2017-07-13
Added DefaultTestNamePattern to runsetting section for NUnit.  #358

OsirisTerje:
Adding DefaultTestNamePatterns to the runsettings nunit section, so that format of test method names , including truncation of length of arguments, can be controlled by the user.

### DIFF
--- a/src/NUnitTestAdapter/NUnitTestAdapter.cs
+++ b/src/NUnitTestAdapter/NUnitTestAdapter.cs
@@ -141,7 +141,7 @@ namespace NUnit.VisualStudio.TestAdapter
             {
                 return TestEngine.GetRunner(package);
             }
-            catch(Exception ex)
+            catch (Exception ex)
             {
                 TestLog.Error("Error: Unable to get runner for this assembly. Check installation, including any extensions.");
                 TestLog.Error(ex.GetType().Name + ": " + ex.Message);
@@ -203,9 +203,15 @@ namespace NUnit.VisualStudio.TestAdapter
             {
                 package.Settings[PackageSettings.DomainUsage] = "Single";
             }
-
-            // Force truncation of string arguments to test cases
-            package.Settings[PackageSettings.DefaultTestNamePattern] = "{m}{a:40}";
+            if (Settings.DefaultTestNamePattern != null)
+            {
+                package.Settings[PackageSettings.DefaultTestNamePattern] = Settings.DefaultTestNamePattern;
+            }
+            else
+            {
+                // Force truncation of string arguments to test cases
+                package.Settings[PackageSettings.DefaultTestNamePattern] = "{m}{a:40}";
+            }
 
             // Set the work directory to the assembly location unless a setting is provided
             var workDir = Settings.WorkDirectory;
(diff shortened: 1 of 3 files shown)


src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -203,9 +203,15 @@ private TestPackage CreateTestPackage(string assemblyName)
...
-            // Force truncation of string arguments to test cases
-            package.Settings[PackageSettings.DefaultTestNamePattern] = "{m}{a:40}";
+            if (Settings.DefaultTestNamePattern != null)
+            {
+                package.Settings[PackageSettings.DefaultTestNamePattern] = Settings.DefaultTestNamePattern;
+            }
+            else
+            {
+                // Force truncation of string arguments to test cases
+                package.Settings[PackageSettings.DefaultTestNamePattern] = "{m}{a:40}";
```
CharliePoole: Any desire to increase the default? Or do you think this would confuse people? A value of 80 might be worth considering for some values that are paths.

OsirisTerje: First I thought it could confuse people, so I thought it best to just do one change, and not multiple, but would it really confuse?   If you hit the limit, then the test results duplicates, so you would see different numbers, but we could just state that in the release notes.  A default limit of 80 as you say could be easier on people.   I am also checking with MS if there are any real limits in VS on this, waiting for answer on that.

OsirisTerje:
I wonder if something would break if they change the whole pattern, if that could be the case, we should go for just the truncation number.  If not, it could be interesting for development to experiment with different patterns.   
As you say it would be easier for the user just to set the truncation number, but then it also hides the real working of this.  Since this can be set in NUnit, would it be too much for the user to do the same in the runsettings?

CharliePoole:
In VS 2012 the limit was something like 450 total.

It's a powerful feature when used globally. You can.  Set the name of every test the same!

CharliePoole:
Speaking as the devil's advocate, VS has expectations about names. What if the name no longer starts with the method bane, for example. Whatever VS requires is particular to the adapter.

OsirisTerje:
First, I got a response from MS saying:  "_The test platform does not impose any limit. Underlying runtime/language/metadata constraints will apply as a appropriate.
For e.g.: see here:_ https://msdn.microsoft.com/en-us/library/sx08afx2.aspx_ "

So I think we should just go for the NUnit default here too, {m}{a}.  

Second:  I played a bit around with different formats, and VS accepted all of them with no real issues. The only thing I noted was that it didn't give out the readable type parameters with {c}.{m}{a}.  Like in issue #357, replaced the Test attribute with different variations 
`[TestCase(TestName="{c}.{m}{a}")]
        public void IsGood()`
gives out
image

Notice the mouse over information, that is more like what one would like to see.  If I use the fullname instead, I do get 4 items there,  but also a lot more noise than necessary
image

It would be nice to have a pattern that would give the same output like the mouse over information.

CharliePoole:
Just clarifying... nunit default is "{m}"

src/NUnitTestAdapter/NUnitTestAdapter.cs
jnm2: I still vote for `"{m}{a}"`.

jnm2:
> So I think we should just go for the NUnit default here too, {m}{a}.

🎉

OsirisTerje:
@CharliePoole If I try {c}{a}
image

According to the template doc page, it should expand the type parameters within angle brackets, but I can't see it does. 

VS will just group them under the same name if the name is shortened down (e.g.. dropping {m}).

PR #361, opened 2017-07-14
Release 3.8.0

OsirisTerje:
***  DO NOT MERGE ***

Changes preparing for the release 3.8.0

- Updating dependencies, including nunit engine
- Adding NUnit3Testadapter as a package to the nunittestadapterTest project(makes it more easy to work including debugging.)


rprouse:
This looks good to me. I should test with my various .NET Core projects, but I don't see anything missing.

OsirisTerje:
@rprouse  Running your test project , before updating the adapter - that is, using the alpha version, gives a failure for VSTest target for the 'dotnet test' command:
image

The command I execute is:
image

IIRC this worked end of May.
What am I missing ?

(I want to use your test suite to verify the .net core parts.  (from https://github.com/rprouse/NUnitPlatformTests)  )

rprouse:
I assume you updated the adapter nuget package for all test projects?

OsirisTerje:
I started out with not updating anything, just to make sure I had a stable baseline, which I then don't have...

OsirisTerje:
Error caused by old dotnet version (old laptop).  Updated and now it works.

OsirisTerje:
Appveyor build fails due to failing downloads from nuget.  
This builds locally, so continue merging
@rprouse  Can you check what's going on with the appveyor build?  I don't think I have access into that one.

rprouse:
The build failure is the Cake build `build.ps1` downloading it's copy of .NET Core. It is possible that the download locations have been moved and we need to update `build.ps1` to the latest version. It might build locally because .NET Core was previously downloaded, or you might have that version already in your path on your machine. Cake installs a local copy in the solution root in `.dotnet`.

Also note, the version that Cake installs has nothing to do with our .NET Core code. Cake is now uses .NET Core itself, so bootstraps itself if it isn't installed. We should do the same for our projects that use .NET Core.

```
.\build.ps1 -Target "Appveyor"
dotnet-install: Downloading https://dotnetcli.azureedge.net/dotnet/Sdk/1.0.1/dotnet-dev-win-x64.1.0.1.zip
```

Issue #302, opened 2017-02-16
BadImageFormatException building solution with unmanaged projects


MaxKot:
I have a solution with both managed and native projects. NUnit adapter seems to try to load a native DLL and fails whenever I rebuild the project. You can see the solution here: [Log4JDash](/MaxKot/Log4JDash/).

Error message (Log4JParserC.dll is an unmanaged library and Log4JDash.exe is an unmanaged application):

> ------ Discover test started ------
> NUnit Adapter 3.7.0.0: Test discovery starting
> Assembly contains no NUnit 3.0 tests: D:\develop\Log4JDash\src\Log4JDash.Web\bin\Log4JDash.Web.dll
> Assembly contains no NUnit 3.0 tests: D:\develop\Log4JDash\src\Log4JParserDashNet\bin\x86\Release\Log4JParserDashNet.exe
> Assembly contains no NUnit 3.0 tests: D:\develop\Log4JDash\src\Log4JParserNet\bin\x86\Release\Log4JParserNet.dll
> Error: Unable to get runner for this assembly. Check installation, including any extensions.
> BadImageFormatException: Format of the executable (.exe) or library (.dll) is invalid.
> Assembly not supported: D:\develop\Log4JDash\src\Release\Log4JDash.exe
> Error: Unable to get runner for this assembly. Check installation, including any extensions.
> BadImageFormatException: Format of the executable (.exe) or library (.dll) is invalid.
> Assembly not supported: D:\develop\Log4JDash\src\Release\Log4JParserC.dll
> NUnit Adapter 3.7.0.0: Test discovery complete
> ========== Discover test finished: 27 found (0:00:00,3920224) ==========

Visual Studio info:

> Microsoft Visual Studio Community 2015
> Version 14.0.25431.01 Update 3
> Microsoft .NET Framework
> Version 4.6.01055

Similar issue is [also reported](/nunit/nunit3-vs-adapter/issues/231#issuecomment-263929798) by @anttirt:

> I have the same issue as MaxKot; it's related to #248.
> 
> The issue is otherwise benign except that the `Error` log causes focus to switch to the `Tests` section in the output window after every build of the solution, which can be very annoying. :)
> 
> The NUnit runner _should_ somehow detect that a project is a native code project and skip it without emitting an error message, but I don't know if there's an API for that off the top of my head. The built in test discoverer for Visual Studio has the following to say:
> 
>```
[...]

PR #369, opened 2017-07-25
Removed exception catching and error reporting in GetRunnerFor, since…

OsirisTerje:
… both calling methods have the same. #302 

The GetRunnerFor method is called from the discoverer and the executor and both calls are within try/catch blocks.  We should be able to safely remove it from the GetRunnerFor method, and keep it in the calling methods.
Further, as the #302 issue points out, it should not be an error if it is a BadImageException.  This is correctly handled in the calling method.


src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -136,17 +136,7 @@ protected void Initialize(IDiscoveryContext context, IMessageLogger messageLogge
...
-            {
-                return TestEngine.GetRunner(package);
-            }
-            catch (Exception ex)
-            {
-                TestLog.Error("Error: Unable to get runner for this assembly. Check installation, including any extensions.");
-                TestLog.Error(ex.GetType().Name + ": " + ex.Message);
-                throw;
-            }
+            return TestEngine.GetRunner(package);
```
CharliePoole: It's easy to get this kind of double exception handling if we don't have a test. Maybe we should add a test that shows this method __does__ throw the exception in this situation.

CharliePoole:
Seems like you have  an odd dependency here. Shouldn't it be depending on the version you are building, not on a package?

OsirisTerje:
The former failure was missing adapter for the demo project, where it is added, as a package. 
Current failure seems to be downloading nuget packages, but hard to see what is wrong here.  
@CharliePoole @rprouse :  Any ideas?

rprouse:
Initial pass through the build output, the adapter tests successfully builds all dependent projects, but then fails itself. It looks like none of its references are being pulled in since it isn't finding the VS SDK namespaces. Still looking...

rprouse:
The build failures seem to be totally unrelated to your changes. It is failing to build the test project, not the demo which you changed. The first error I see is,

```
AdapterSettingsTests.cs(6,17): error CS0234: The type or namespace name 'VisualStudio' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\projects\nunit3-vs-adapter\src\NUnitTestAdapterTests\NUnit.TestAdapter.Tests.csproj]
```

I might be missing something in that wall of text, but it looks like the adapter built successfully. I checked the `appveyor.yml` file and it specifies Visual Studio 2017. I checked the list of environments and there isn't anything newer.

I have no idea. I am going to kick off another build. I might also try to rebuild master.

rprouse:
We haven't had a successful build since the release, so it could be any number of changes causing the failure. Does it build from the command line for you if you delete your packages directory?

OsirisTerje:
@rprouse Yes, removed local packages, and it still builds fine.  It does say that the .net  sdk  1.0.1 is already installed though.  Is that the case on the build server?

CharliePoole:
@OsirisTerje This old issue just popped back up in a search related to a problem I'm working on. Mono.Cecil, by design, aims to read managed assemblies. As noted it throws an exception otherwise.

We would probably benefit by examining unmanaged assemblies and returning a flag that says they are unmanaged rather than trying to find a driver for it. Just wanted to get this notion on your radar. If the adapter could easily tell that the assembly is unmanaged, it could simply ignore it. As soon as we let Mono.Cecil get called, the trouble arises.

Issue #392, opened 2017-09-28
Improve performance of discovery


OsirisTerje:
It has been found that discovery of test cases are slowed down by the processing of traits.

navin22:
With PR #406 I am seeing the below performance improvements for 10K empty nUnit tests in a project.

For Discovery: 67.5% perf improvement in console and 55.9% perf improvement in IDE.
Framework   /Adapter | VS CLI (vstest.console.exe) | VS IDE (VS2017 Latest)
-- | -- | --
3.8.1   / 3.8.0 | 5.24s | 8.4s
3.9.0-dev-04564 / 3.9.0-dev-00575 | 1.7s | 3.7s

For Execution: 26.7% perf improvement in console and 22.35% perf improvement in IDE.
Framework /Adapter | VS   CLI (vstest.console.exe) | VS   IDE (VS2017 Latest)
-- | -- | --
3.8.1 / 3.8.0 | 15.65s | 21.25s
3.9.0-dev-04564   / 3.9.0-dev-00575 | 11.47s | 16.5s

The above testing is done in following machine configuration:
OSName       : Microsoft Windows 10 Enterprise
OSVersion    : 10.0.15063
MachineType  : x64-based PC
Processor    : Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz
LogicalCores : 8
RAMSize      : 15.9589996337891GB

PR #393, opened 2017-09-28
Reading tratis/properties from xml test-run tree using cache.

navin22:
- Found that `XmlNode.SelectNodes` is a costly call. And this happens multiple time if the `test-run` tree is big for each leaf (`test-case`) nodes. Hence reduced the number of calls to it.
- Used a cache to store all the ancestor properties of `test-case` node.
- Removed recursion and used a loop to traverse up the tree from test-case until test-run.
- Removed `TraitFeature.IsSupported`
- Added tests


CharliePoole:
This PR is getting a bit bigger than we usually like. It's now doing two major things:

1. Removing use of reflection in detecting traits feature so that we no longer support VS2012 RC.
2. Introducing a cache of properties in order to improve performance.

The first item impacts the feature set directly. Many people would say it calls for a 4.0 release, although I wouldn't go that far. At a minimum, however, I'd like to see it in a separately merged PR, so that we get some feedback from users and make sure it doesn't have any side effects we have not planned on.

OsirisTerje:
On 1) That is the reason I have pushed for more tests, and asked about the different xml blocks.  I have gather some, and need to add the latest on about seperate testcase with category attribute.  

On 2) That cache together with the removal of the access to the xml navigation is the primary fixes for this PR and issue.  The removal of the reflection can be seperated out, but I doubt anyone will react it.  We checked the usage (got some MS metrics info), and it is very low for VS2012 RTM, and further, those not wanting to update VS would probably not want to update the adapter either.

CharliePoole:
@OsirisTerje Well, it's up to you as the project lead, of course. I'll just point out that we made a big deal out of dropping .NET 1.0 support only a few years ago. When the change is as big as invalidating a platform, we only do it if the usage is low, of course. But if we do it, we have generally made a pretty big deal of announcing it, etc. rather than slipping it in as part of another change.

I understand that the adapter may well go by a different set of rules, of course, and one reason I'm raising this is that I would like to see you articulate those rules somewhere.

OsirisTerje:
@CharliePoole Good points.  I agree that announcing the dropping of VS2012 RTM support for future adapters is a very good point.  That way we can get some response, if possible.  Dropping VS2012 RTM however, is much less than dropping a .net framework, but I still agree.   
Have you announced earlier on other channels than the discussion lists and nunit.org news sites?

About rules for this, I am not sure we can predict the future well enough to set up any valid rules for this.  MS changing the interface back then was a surprise, it has been stable since then.   The stuff around .net core concerns me, and also the Objectmodel 11 issue (with 3.5. versus 4.0 support).   I like the backwards compatibility principle, until we can no longer do it without too much pain, like now with the VS2012 RTM.  I am also concerned about some of the bugs we have that indicates we are not fully stable on all platforms and  covers all variations.  (E.g. #399 ) , we should be.  I could try to write up something around these things, but not sure if it really would help.  @jnm2 @ChrisMaddock : What do you guys think?

OsirisTerje:
I am also wondering if we should merge this into an integration or feature branch first, before going all the way to the master.  The reason is it now lives in the fork, and I would like to contribute up some of the tests easily.  As @CharliePoole says, it starts to become big, and we need to ensure it really works.  @rprouse : I am also concerned about the Objectmodel 11 as discussed.

jnm2:
I don't think we should ever support release candidates (VS2012 RC mentioned at the beginning) but I'd expect dropping of VS2012 to be a v4-only item *unless* we had already announced at the beginning of the v3 series some criteria by which we might drop VS2012 during v3.

jnm2:
I think we should at least give it two or three releases, all of which mention the imminent VS2012 obsoletion at version 3.X in the NuGet package manager, as the only real way of gathering feedback.

jnm2:
I mean sure: I'm usually the first to ask, why are you using five-year-old VS? But per our recent discussions, we want to keep people with corporate policies in mind even to the extent of building NUnit itself, let alone consuming it. Also I have a random reason to need VS2012 at work for certain projects (https://stackoverflow.com/questions/20978721/visual-studio-2013-crashes-with-netfx40-legacysecuritypolicy-dynamic-operation), so it's not totally implausible to me.

OsirisTerje:
The issue is RTM, not the RC.  We never really supports RCs. 

In your case above, anything preventing you from using Update 1-5 of VS 2012 ?   It is *only* the RTM specific version, not complete VS2012.

jnm2:
Huh, I should put that through VS feedback now that I'm not faced with using Connect anymore.

jnm2:
> In your case above, anything preventing you from using Update 1-5 of VS 2012 ? It is only the RTM specific version, not complete VS2012.

Ah, this is a more interesting question which I missed! Personally I think it's fine to require the very latest servicing update of each product we support in parallel.

jnm2:
Can anyone think of a counterexample?

CharliePoole:
@OsirisTerje Yes, I meant RTM. RC was a combination of a typo and the fact that it actually did work with the RC as well as with many pre-RC releases. In fact, it originally was developed using VS2010 with an extension created by Microsoft to implement the test explorer window. 😈 

What I'm getting at is that we have had a policy up to now, even if it was not clearly spelled out in writing. Just as people get used to our bugs and sometimes complain when we fix what we thought was a bug and they considered a feature, people have gotten used to a policy of pretty strong backward compatibility. Therefore, if it's going to change, I think you need to find a place to announce it. To me, the mailing list and an issue would be a good approach. Of course, we know people don't always read these things, but IMO you should make an effort to tell them. I don't think you need to __ask__ however, just announce what your policy is.

CharliePoole:
@jnm2 Your points are well taken - on both sides of the argument. 😄 

It's possible for a piece of software that extends some product to say we only support the latest release of the product for our new releases and folks using the older releases do not get new features.

It's also possible to support multiple builds for each version of the base product you are extending, although that requires a bit of work.

And finally, it's possible to say we support a specific range of releases of the base product, which has been our position for a long time with the adapter.

If we drop support for VS2012 RTM, I don't think that's a change in strategy, just a change in the details of which version we support. 

Originally, I decided the strategy as well as the details in collaboration with the team that created the explorer. Later, I did it with input from the team that gradually built up. Now it's in @OsirisTerje 's hands with input from all of us. I think any of the above strategies can work, but it will need to be made clear to users who are now used to the old strategy.

CharliePoole:
@OsirisTerje What about a development branch?

CharliePoole:
@jnm2 To what?

jnm2:
> And finally, it's possible to say we support a specific range of releases of the base product, which has been our position for a long time with the adapter.

> If we drop support for VS2012 RTM, I don't think that's a change in strategy, just a change in the details of which version we support.

Yes. That sounds good to me.

> > Can anyone think of a counterexample?
> 
> To what?

A counterexample to my [suggestion](https://github.com/nunit/nunit3-vs-adapter/pull/393#issuecomment-334531533) that it's fine to starting requiring certain minimum VS servicing release in order to use a minor update of the NUnit adapter.

OsirisTerje:
@CharliePoole Yep.  Since we're working on Issue #392, we could just create a dev-392 branch.  I'll do that, and @navin22 , can you change this PR to ask for a merge into that branch?

navin22:
@OsirisTerje : Sure let me know, once that branch is ready.

OsirisTerje:
@navin22   It's up  :-)

navin22:
Created https://github.com/nunit/nunit3-vs-adapter/pull/401

PR #398, opened 2017-10-03
Get .NET Core compiling and tests running from Cake

### LINKED ISSUE #302: BadImageFormatException building solution with unmanaged projects

### LINKED ISSUE #395: Test do not run with .NET Core 2.0 installed
Since I installed .NET Core 2.0, the tests do not run from the command line,

```
.\build.ps1 -Configuration Release -Target Test
```

```
========================================
TestAdapterNetCore
========================================
Executing task: TestAdapterNetCore
The specified framework 'Microsoft.NETCore.App', version '1.0.5' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\src\nunit\nunit3-vs-adapter\.dotnet\shared\Microsoft.NETCore.App
  - The following versions are installed:
      1.0.4
      1.1.1
  - Alternatively, install the framework version '1.0.5'.
```

rprouse:
I am not sure when this stopped working or exactly why. I thought it was because I installed .NET Core 2.0, but it seems to have been happening before that. Cake build was installing .NET Core 1.0.1 and adding it first in the path. This was causing the .NET Core app to fail to build.

To fix this, I first tried updating Cake to the latest version. This helped, but wasn't quite right because it installed .NET Core 1.0.4. I then switched the Cake bootstrap script to instead install .NET Core 1.1.4. This got everything compiling and the tests running.

This led to one .NET Core test failing. This was added after the tests broke in #369 to fix #302. No error or warning messages are reported by the adapter when attempting to load a native assembly using the .NET Core adapter, so I don't think #302 is an issue on .NET Core and so the test is invalid. I switched the test to only run full framework.

I also noticed that an old version of `NUnit3TestAdapterTests.csproj` got re-added to the directory, so I deleted it again. The project was renamed.

Fixes #395


build.ps1
```diff
@@ -31,12 +31,16 @@ Param(
     [string[]]$ScriptArgs
 )
 
-$CakeVersion = "0.19.3"
+$CakeVersion = "0.22.1"
 $DotNetChannel = "preview";
-$DotNetVersion = "1.0.1";
-$DotNetInstallerUri = "https://raw.githubusercontent.com/dotnet/cli/v1.0.1/scripts/obtain/dotnet-install.ps1";
+$DotNetVersion = "1.1.4";
```
rprouse: This file, `build.ps1` is a direct copy of the Cake 0.22.1 release version except for this one line that I updated from `1.0.4` to `1.1.4`. Everything else is as shipped.

build.cake
```diff
@@ -5,7 +5,7 @@
 //////////////////////////////////////////////////////////////////////
 
 var target = Argument("target", "Default");
-var configuration = Argument("configuration", "Debug");
+var configuration = Argument("configuration", "Release");
```
rprouse: Cake itself no longer passes in Release by default, so better to set it here to ensure that CI builds don't mistakenly switch to Debug. Charlie ran into this and we have been updating all repos this way.

rprouse:
Woot, builds are green again! 🎉 😸

OsirisTerje:
Looks good.

Issue #397, opened 2017-10-02
Improve verbosity and diagnostic logging


OsirisTerje:
This issue covers some improvements to  logging and verbosity, what we need to do to make information in the adapter more visible, without introducing too much noise.  It mostly will expose some of the internal information in the standard output window.

PR #402, opened 2017-10-05
Verbosity improvement for adapter settings  #397

OsirisTerje:
We're making changes to the usage of runsettings, and there is hard to see if they come in correct, so this PR contains updates to the visibility of the runsettings when verbosity is set higher (>=4).  
Since we're schema-checking the runsettings, this diagnostics will make it easier to see if you got the information correctly in. 

Example:


CharliePoole:
Code LGTM - I didn't check why it's failing CI.

Issue #404, opened 2017-10-06
Change format of Fully qualified Name


OsirisTerje:
The FQN is now in a format given by NUnit, which includes NUnit specific information, and is passed on to VS.  VS require it to be in standard format.

CharliePoole:
This has, of course, been true for years. VS has unstated, undocumented assumptions about the test name, which the MS dev have consistently refused to clarify for us. That's the main thing that led me, as the original author of the extension, to give up on it.

Do we now know with any precision what those assumptions are? Is it documented? Will future changes be documented?

I think this issue should say what the changes will be and it should remain only a discussion item until the format is agreed upon with Microsoft. I realize there have been private discussions but I'd like to see it spelled out as well as agreed to here in public.

OsirisTerje:
We're currently output'ing a non-standard - fqn.   The change here just puts it back to a normal standard fqn.

CharliePoole:
I'm asking for a definition of the format of a "standard fqn" for a TestCase. I'm asking that it be reviewed here, by our team and by somebody from the VS team. Otherwise, we just continue to do what we have done for years: make a guess of what Microsoft wants and try it, then change it when it no longer works.

peterwald:
@CharliePoole You make some good points. 

I'm going to have some conversations internally to make sure we communicate clearly about what the FQN needs to be. I'll respond back soon with more information.

peterwald:
The FullyQualifiedName should follow the pattern of CLR fully qualified names and reference the method that implements the test. Optionally, the FullyQualifiedName can include angle brackets for generic parameters, and parentheses for method parameters.

Here are some examples:

```
Namespace.Class.Method
Namespace.Class.Method()
Namespace.Class.Method(x: 0, y: 10)
Namespace.Class+Inner.Method(str: "Hello")
Namespace.Class<T>.Method<V>(val: "x")
```

CharliePoole:
Comments on what NUnit provides...

* In each case, Method may be replaced by the user. However, the adapter actually has the class and method name and could construct from those.

* We never use empty parentheses

* Parameter names are not available, so we would use `Method(0, 10)` rather than `Method(x: 0, y: 10)`

* When angle brackets are used, the __actual__ type is provided rather than the type parameter name.

* Classes may also take arguments in parentheses.

@peterwald are any of these points issues for you? We might be able to overcome some of them in the adapter using other info that's in the NUnit XML.

peterwald:
[...]

PR #405, opened 2017-10-06
Fully qualified name change

OsirisTerje:
See issue #404


CharliePoole:
@OsirisTerje Is this ready for review or do you need to make changes based on the discussion on #404?

OsirisTerje:
Docs from Peter here:  https://github.com/Microsoft/vstest-docs/pull/95

rprouse:
@OsirisTerje are you waiting for anything in this PR to merge and close?

OsirisTerje:
Just need to resolve the conflicts and its good to go
...in progress.....

CharliePoole:
@OsirisTerje This looks good to me based on what we have been told by Microsoft - that the presence of arguments is not required and that FQNs don't have to be unique. 

Have you verified that in the IDE? Does it work now, or does it depend on changes they are making?

OsirisTerje:
This is unfortunately not working as it should, or as hoped.  We loose information and what is displayed in the test explorer are less than before.  
image

(1) This is the former way it was displayed, one test per specific set of test cases.  
(2) With this change it now displays like this, it looks like only one of the tests is displayed, but this just because the mismatch between the display name and the FQN.  
(3) Running the tests displays the 3 different test cases.

Having the results like 2-3 is one the biggest complaints we get.  Using the special FQN names they display better.  

What is needed here is a change in the Test Explorer and an explicit support for test cases (aka data driven tests).  
The discovery must list all variations in the test explorer.  One way it could be done is to add one more hierarchical level. That would have made more sense. 

@peterwald  I suggest we delay this change so that we don't annoy people with reduced functionality, and redo this when the test explorer can handle test cases properly.

peterwald:
@OsirisTerje I agree with holding this change. The issue here is that if FQN is not unique, then the ID must be made unique. 

Either we need to include the parameters in the FQN in order to make it unique, or you need to assign the `TestCase.Id` before returning it. If the Id is not assigned, then it is constructed from a hash of the Source,Executor, & FQN. The Source and Executor are not changing, so the FQN would determine the uniqueness of the test case.

We are already normalizing the FQN currently, so the existing strategy should continue to work until we have a chance to address the data-driven tests.

Issue #327, opened 2017-05-12
Latest Mono produces Portable .pdb instead of .mdb(Mono.Cecil update required)


DavidKarlas:
Latest Mono is producing Portable .pdb instead of .mdb by default, resulting in this exception:
```
 System.IO.FileNotFoundException: Could not find file "/Users/davidkarlas/Projects/nunit24214asf/nunit24214asf/bin/Debug/nunit24214asf.dll.mdb".
File name: '/Users/davidkarlas/Projects/nunit24214asf/nunit24214asf/bin/Debug/nunit24214asf.dll.mdb'
  at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share, System.Int32 bufferSize, System.Boolean anonymous, System.IO.FileOptions options) [0x00195] in /private/tmp/source-mono-2017-04/bockbuild-xamarin/profiles/mono-mac-xamarin/build-root/mono-x64/mcs/class/corlib/System.IO/FileStream.cs:232 
  at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share, System.Int32 bufferSize, System.Boolean isAsync, System.Boolean anonymous) [0x00000] in /private/tmp/source-mono-2017-04/bockbuild-xamarin/profiles/mono-mac-xamarin/build-root/mono-x64/mcs/class/corlib/System.IO/FileStream.cs:149 
  at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access) [0x00000] in /private/tmp/source-mono-2017-04/bockbuild-xamarin/profiles/mono-mac-xamarin/build-root/mono-x64/mcs/class/corlib/System.IO/FileStream.cs:86 
  at (wrapper remoting-invoke-with-check) System.IO.FileStream:.ctor (string,System.IO.FileMode,System.IO.FileAccess)
  at Mono.CompilerServices.SymbolWriter.MonoSymbolFile.ReadSymbolFile (System.String mdbFilename) [0x00000] in <682b0e3580fe403dad53adbaea0f93fd>:0 
  at Mono.CompilerServices.SymbolWriter.MonoSymbolFile.ReadSymbolFile (System.String mdbFilename, System.Guid assemblyGuid) [0x00000] in <682b0e3580fe403dad53adbaea0f93fd>:0 
[...]

PR #422, opened 2017-12-04
Updated adapter to send 'AdjustedFQN' TestProperty as ClassName.MethodName for TestName attribute scenario

Shyam-Gupta:
Updated adapter to send FQN as ClassName.MethodName

### DIFF
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -149,7 +149,7 @@ namespace NUnit.VisualStudio.TestAdapter
             }
             else
             {
-                bool isTestNameScenario = IsTestNameScenario(testNode, methodName, displayName);
+                bool isTestNameScenario = IsTestNameScenario(methodName, displayName);
 
                 if (isTestNameScenario)
                 {
@@ -163,7 +163,7 @@ namespace NUnit.VisualStudio.TestAdapter
             return testCase;
         }
 
-        private bool IsTestNameScenario(XmlNode testNode, string testCaseMethodName, string testCaseDisplayName)
+        private bool IsTestNameScenario(string testCaseMethodName, string testCaseDisplayName)
         {
             if (string.Compare(testCaseMethodName, testCaseDisplayName) == 0)
             {
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -24,6 +24,8 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Security.Cryptography;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Xml;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
@@ -74,11 +76,11 @@ namespace NUnit.VisualStudio.TestAdapter
             string id = testNode.GetAttribute("id");
             if (_vsTestCaseMap.ContainsKey(id))
                 return _vsTestCaseMap[id];
-           
+
             // Convert to VS TestCase and cache the result
             var testCase = MakeTestCaseFromXmlNode(testNode);
             _vsTestCaseMap.Add(id, testCase);
-            return testCase;             
+            return testCase;
         }
 
         public TestCase GetCachedTestCase(string id)
@@ -127,20 +129,24 @@ namespace NUnit.VisualStudio.TestAdapter
         /// </summary>
         private TestCase MakeTestCaseFromXmlNode(XmlNode testNode)
         {
+            var className = testNode.GetAttribute("classname");
+            var methodName = testNode.GetAttribute("methodname");
+            var fullyQualifiedName = $"{className}.{methodName}";
+            var fullName = testNode.GetAttribute("fullname");
+
             var testCase = new TestCase(
-                                     testNode.GetAttribute("fullname"),
-                                     new Uri(NUnitTestAdapter.ExecutorUri),
-                                     _sourceAssembly)
+                          fullyQualifiedName,
+                          new Uri(NUnitTestAdapter.ExecutorUri),
+                          _sourceAssembly)
             {
                 DisplayName = testNode.GetAttribute("name"),
                 CodeFilePath = null,
-                LineNumber = 0
+                LineNumber = 0,
+                Id = GuidFromString($"{NUnitTestAdapter.ExecutorUri}{_sourceAssembly}{fullName}")
             };
 
             if (_collectSourceInformation && _navigationDataProvider != null)
             {
-                var className = testNode.GetAttribute("classname");
-                var methodName = testNode.GetAttribute("methodname");
                 var navData = _navigationDataProvider.GetNavigationData(className, methodName);
                 if (navData.IsValid)
                 {
@@ -148,7 +154,7 @@ namespace NUnit.VisualStudio.TestAdapter
                     testCase.LineNumber = navData.LineNumber;
                 }
             }
-            
+
             testCase.AddTraitsFromTestNode(testNode, TraitsCache);
 
             return testCase;
@@ -259,7 +265,7 @@ namespace NUnit.VisualStudio.TestAdapter
                 case "Failed":
                     return TestOutcome.Failed;
                 case "Skipped":
-                    return resultNode.GetAttribute("label")=="Ignored"
+                    return resultNode.GetAttribute("label") == "Ignored"
                         ? TestOutcome.Skipped
                         : TestOutcome.None;
                 case "Warning":
@@ -283,6 +289,17 @@ namespace NUnit.VisualStudio.TestAdapter
                 case "Inconclusive":
                 default:
                     return TestOutcome.None;
+            }
+        }
+
+        private Guid GuidFromString(string data)
+        {
+            using (var provider = SHA1.Create())
+            {
+                byte[] hash = provider.ComputeHash(Encoding.Unicode.GetBytes(data));
+                byte[] toGuid = new byte[16];
+                Array.Copy(hash, toGuid, 16);
+                return new Guid(toGuid);
             }
         }
 
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -141,12 +141,19 @@ namespace NUnit.VisualStudio.TestAdapter
 
             if ((_collectSourceInformation || bIsTestNameScenario) && _navigationDataProvider != null)
             {
-                var className = testNode.GetAttribute("classname");
-                var navData = _navigationDataProvider.GetNavigationData(className, methodName);
-                if (navData.IsValid)
+                try
                 {
-                    testCase.CodeFilePath = navData.FilePath;
-                    testCase.LineNumber = navData.LineNumber;
+                    var className = testNode.GetAttribute("classname");
+                    var navData = _navigationDataProvider.GetNavigationData(className, methodName);
+                    if (navData.IsValid)
+                    {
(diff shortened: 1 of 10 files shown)


OsirisTerje:
1. Missing the issue
2. Test data should not be altered, they are based on what NUnit returns
3. Explain changes

Shyam-Gupta:
@OsirisTerje 

Hi Terje,

Sorry for delay. Created the issue: https://github.com/nunit/nunit3-vs-adapter/issues/423

I am working on this PR. Its not ready yet. Will ping you once the changes are done.

Thanks.

CharliePoole:
I added a question to the issue. I'm basically uncertain whether changing the name (FQN) that we give to the Test Explorer is the most robust way to go in the future. We currently provide the source information and it's not clear in the issue whether what we are providing is wrong or if the explorer has been changed so as to ignore what we provide.

A big concern for me  is that if the explorer has changed, we might need to detect what version we are running under and do different things.

OsirisTerje:
I assume it is the CollectSourceInformation setting that will be sent down from the newer Test Explorer, and this is what we then can use to switch between legacy mode (to call it so for older versions) and a new way of working when the Roslyn test detector is operating, if I have got this right,  @Shyam-Gupta ?

OsirisTerje:
See issue https://github.com/nunit/nunit3-vs-adapter/issues/423

CharliePoole:
@OsirisTerje So, somewhat ironically, if CollectSourceInformation is set to true we __won't__ collect it any longer but will instead fake the FQN?

It seems like there should be a more general way to do this. What if the test assembly isn't written in C# in the first place?

OsirisTerje:
@CharliePoole Yup, but in the execution phase we always return that data, so we have to decide what a FQN really is, and it does make sense to make it namespace.classname.methodname.   The issue I see with this is that the current test explorer will then give us only one test case with multiple results for our testcases. Today we have this in certain situations and that has been a common complaint.  So for legacy test explorers I would at least prefer the current method where we set the FQN equal to the ns.classname.displayname, that gives us at least a good experience in the TE

CharliePoole:
So what will TE do if an NUnit test assembly is written in a language not supported by the Roslyn compiler?

OsirisTerje:
Your guess as good as mine, but perhaps not send down the CollectSourceInformation setting, it is sent down once per assembly.

CharliePoole:
But since we support such languages, don't we need to know?

CharliePoole:
Here's an idea: what about duplicating the demo program in this repo in some other languages. It could stay here or go to another repo, whatever we prefer.

OsirisTerje:
That makes sense.  What language should we choose?  Which is not backed by roslyn ofc.

CharliePoole:
I'm thinking managed C++... although F# would be a nice choice as well.

I'll write an issue.

OsirisTerje:
F# would be nice, but isn't that backed by Roslyn by now?

CharliePoole:
I quickly scanned the source and didn't see it, but I could have missed it.

CharliePoole:
Both Rob and I asked questions on the issue, which are not answered. We are in a cart-before-the-horse situation here: we haven't selected among several alternative approaches, yet we have a PR that implements a particular approach. IMO, we should neither approve nor disapprove this PR until we decide on the approach that is to be taken. That's what issues are for!

OsirisTerje:
@Shyam-Gupta Can you please answer the different questions raised above?  Also see the questions on the PR.

Shyam-Gupta:
@CharliePoole @OsirisTerje 
Since VS 2017 Update 5, Visual Studio Test Explorer has started using Roslyn to fetch source information for discovered tests. This change was done to improve test discovery performance. With this change we are seeing ~22% perf improvement for nUnit and ~58% for xUnit for 10k unit tests discovery. 

As part of this change, for each project Test Explorer first detects whether Roslyn can be used to fetch source information. If it can, then it adds following tag in RunSettings and passes it to the test adapter:

<CollectSourceInformation>false<CollectSourceInformation>

This setting indicates that adapter need not collect source information. The newest version of nUnit adapter (3.9.0) has already been updated to honor this setting. Hence newer adapter will not send source information and it will be fetched by Test Explorer.

Older adapters, which do not honor this setting, will continue to fetch source information. To handle that Test Explorer will first detect whether the discovered tests already have this information. If they have it, then Test Explorer will not attempt to fetch this information again from Roslyn. Thus it takes care of compat. More details about this change are available [here](https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0010-Source-Information-For-Discovered-Tests.md).

This change is also required to support test discovery from Visual Studio on Mac. Please check following issue for more details: [https://github.com/nunit/nunit3-vs-adapter/issues/327](https://github.com/nunit/nunit3-vs-adapter/issues/327)

Shyam-Gupta:
@CharliePoole @OsirisTerje
This PR implements a workaround to fetch source information for test methods having 'TestName' attribute. Due to time constraints, we have implemented it for VS 2017 Update 6 release only. For next release we will implement the proper fix as discussed in bug 423.

Approach used for workaround: 

For each discovered test, check whether actual test method name and display name are same. If they are same then its not 'TestName' scenario. Hence no extra step required.
If they are not same then verify whether actual test method name has generic parameters or function parameters. Get rid of those and again check if the two names are same.
If they are still not same then compute AdjustedFQN = {className}.{methodName}, and stash it as 'AdjustedFQN' property in TestCase properties. 
TestWindow will check for this property. If it is present then it will use value of AdjustedFQN to fetch source information.

Shyam-Gupta:
@OsirisTerje 

_"Test data should not be altered, they are based on what NUnit returns"_

We have generated test data with help of following runsettings. Hence it is the actual value what NUnit returns.
```
<NUnit>
    <verbosity>5</verbosity>
    <DumpXmlTestDiscovery>true</DumpXmlTestDiscovery>
</NUnit>
```

I have answered the questions on this PR and [bug 423](https://github.com/nunit/nunit3-vs-adapter/issues/423). Please let me know if more information is required.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -127,33 +124,81 @@ public IList<VSTestResult> GetVSTestResults(XmlNode resultNode)
...
+                }               
+            }
+            else
+            {
+                bool isTestNameScenario = IsTestNameScenario(methodName, displayName);
+
+                if (isTestNameScenario)
+                {
+                    // Stash property in test case object
+                    testCase.SetPropertyValue(Constants.TestCaseAdjustedFQNProperty, string.Format("{0}.{1}", className, methodName));
```
CharliePoole: Why do we give you two things in one property rather than setting two different properties? It seems to me that giving you __individual__ properties for whatever information you need is a much better long term strategy. Then you can make use of them in any way you like.

This is important because __we will have to maintain whatever fix we put in here in the future__. So if we decide in the future to give you class name, method name and individual args as properties, we will nevertheless be stuck with this "AlternateFQN" property as legacy.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -127,33 +124,81 @@ public IList<VSTestResult> GetVSTestResults(XmlNode resultNode)
...
                 var navData = _navigationDataProvider.GetNavigationData(className, methodName);
                 if (navData.IsValid)
                 {
                     testCase.CodeFilePath = navData.FilePath;
                     testCase.LineNumber = navData.LineNumber;
+                }               
+            }
+            else
+            {
+                bool isTestNameScenario = IsTestNameScenario(methodName, displayName);
```
CharliePoole: This line and the method it calls attempts to reverse engineer NUnit in order to decide if the user changed the test name. That's a very fragile thing to do, and it's not clear why you need to do it. Why not __always__ provide the methodname and class name and make use of them by preference to find source information rather than maintaining a hybrid approach in the test explorer?

Of course, for the present, you would have to maintain a hybrid approach anyway, in case the method name and class name are not provided to you, but it makes sense to me that we should drive toward a cleaner solution in the future both for NUnit and other adapters.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -127,33 +124,81 @@ public IList<VSTestResult> GetVSTestResults(XmlNode resultNode)
...
+        private bool IsTestNameScenario(string testCaseMethodName, string testCaseDisplayName)
+        {
+            if (string.Compare(testCaseMethodName, testCaseDisplayName) == 0)
+            {
+                return false;
+            }
+            else
+            {
+                // Can be TestName, Generic, or Parameterized test case scenario
+                int index = testCaseDisplayName.IndexOf('<');
```
CharliePoole: Suppose the display name is `MyMethod("A<B")`.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -127,33 +124,81 @@ public IList<VSTestResult> GetVSTestResults(XmlNode resultNode)
...
+                    return string.Compare(testCaseMethodName, testCaseDisplayName) != 0;
+                }
+                else
+                {
+                    // Parameterized test method
+                    index = testCaseDisplayName.IndexOf('(');
+
+                    if (index > -1)
+                    {
+                        testCaseDisplayName = testCaseDisplayName.Remove(index).TrimEnd();
```
CharliePoole: I think this doesn't properly allow for the fact that the user can put __anything__ into the display name, including parentheses. What if the name is `(`? Crazy but possible and we should not throw. In addition, if the method has arguments, then one of them may be a string containing a left parenthesis.

The code can be fixed to allow for that, of course, returning false if there is any doubt. However, this illustrates for me how fragile it is to try to reverse engineer the NUnit test names. In an early version of the adapter, we did that ourselves and it didn't work for us. That's why we introduced attributes for method and class name. Putting code in the adapter that tries to reverse engineer nunit is a step backward and I think we should try to avoid it by simply using the method and class names provided in all cases.

src/NUnitTestAdapter/TestConverter.cs
Shyam-Gupta: Generic parameters are not retained in methodname. If we always use classname and methodname then we will not be able to find source information for generic overloaded methods.

src/NUnitTestAdapter/TestConverter.cs
Shyam-Gupta: TestWindow will use "AlternateFQN" only if it is available. If it is not available, Test Window will continue to use the FQN returned by NUnit at present.
Hence in future once we implement the right solution, we can remove this property.

src/NUnitTestAdapter/TestConverter.cs
Shyam-Gupta: Thanks for the feedback Charlie. Let me see if I can find a better solution. I will get back soon.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -127,33 +124,45 @@ public IList<VSTestResult> GetVSTestResults(XmlNode resultNode)
...
                 {
                     testCase.CodeFilePath = navData.FilePath;
                     testCase.LineNumber = navData.LineNumber;
+                }               
+            }
+            else
+            {
+                var testName = testNode.GetAttribute("TestName");
+
+                if (!string.IsNullOrEmpty(testName))
```
CharliePoole: I don't understand this. Where do you expect the `TestName` attribute to come from?

Shyam-Gupta: Added 'testname' attribute in test case xml. Please check PR: https://github.com/nunit/nunit/pull/2657/files

Shyam-Gupta:
Closing this PR for time being. We will come up with a more cleaner approach soon.

PR #429, opened 2017-12-13
Multiple versions of NUnit Test Demo for the adapter

### LINKED ISSUE #424: Expand and reorganize Demo solution
We use the NUnit 3 Test Demo for exploratory testing,, particularly for seeing how things appear in the VS IDE. This would be facilitated if the demo were available for some other languages besides C# and with the solution moved to the top level of the project. Currently, it is nested under the adapter solution and doesn't get as much visibility as it should. As a further step, we might move it to its own repository but this issue does not envision doing that right away.

CharliePoole:
** Do not merge yet **

FIxes #424 

The initial commits here are a start on accomplishing #424. I'd like to get comments from @OsirisTerje on the overall direction before continuing.

There are two VS2017 builds for the demo program, both in C#. One of them targets .NET 4.5 and the other .NET Core. We are not yet running the .NET Core demo in the script.

There are three builds using VS2015 and targeting .NET 4.5. They are written, respectively, in C#, VB and C++. The C++ version doesn't include all the tests from the original version due to limitations in use of managed C++ to run NUnit tests.

After working on this for a while, my conclusion is that it should be in a separate repo for several reasons:

1. Up to now we haven't run the demos in CI and running them increases the run time.
2. There's no reason to run the demos unless we change them as their primary use is for hands-on testing in the IDE.
3. Keeping them separate facilitates installing different versions of the adapter when testing.

Before going any further, I think we should come to a decision about using a separate repo or not. There's no point adding all this stuff to the current repo and then dropping it later. If we want it to be separate I may as well do it from the start.


OsirisTerje:
@CharliePoole I think a separate repo makes sense.  We should perhaps also consider if the platform examples that @rprouse   has in https://github.com/rprouse/NUnitPlatformTests could be merged with the demo code.

CharliePoole:
I'll go ahead with that and then close this issue. As you point out, the structure will have to be more flexible if different apps are combined. Any comments on the structure I have set up?

BTW I notice that tests in Appveyor take a long time in the restore phase. I wonder if there is more caching that we could do.

CharliePoole:
Closing this without merging, since the test demo project has moved to its own repository.

PR #431, opened 2017-12-21
Remove Demo project from this repo - now in its own repo

### LINKED ISSUE #424: Expand and reorganize Demo solution

CharliePoole:
Fixes #424 

The NUnit3 Test Demo project is now in its own repository: nunit/nunit3-vs-adapter-demo. This PR removes the demo from the adapter repo and updates the build script.


OsirisTerje:
Looks good.

Issue #419, opened 2017-11-21
Test result is Skipped when an exception has been thrown and the only warning is in TearDown


nfink:
The VS test adapter reports a test result as Skipped instead of Failed under the following circumstances:
 - An exception was thrown during the test (not an assertion failure, but an actual exception).
 - A warning has been asserted in the TearDown of the test.
 - No warnings have been asserted prior to the exception being thrown.

That means that for example, code with the following is reported as Failed:
(Setup): Assert.Warn("...")
(Test): throw new Exception()
(TearDown): Assert.Warn("...")

But the following is reported as Skipped:
(Test): throw new Exception()
(TearDown): Assert.Warn("...")

The attached screenshots show the discrepancy in results between the test adapter and the console runner. The gist linked below has the code that was run.

* NUnit and NUnit3TestAdapter versions: NUnit 3.9.0, NUnit3TestAdapter 3.9.0
* Visual Studio edition and full version number: VS Professional 2015 v14.0.25424.00 Update 3
* A short repro, preferably attached or pointing to a git repo or gist: https://gist.github.com/nfink/ed40e164ce8966bb40df59e6397b852e
* What .net platform and version is being targeted: .NET Framework 4.5.2

![nunittestadapter](https://user-images.githubusercontent.com/10527639/33089820-de932848-cebf-11e7-9d7f-771cc616d1cb.png)
![nunitconsole](https://user-images.githubusercontent.com/10527639/33089839-e90f3b4a-cebf-11e7-877d-2a3d60a47315.png)

OsirisTerje:
Thanks @nfink , I'll have a look

OsirisTerje:
It seems this bug comes from NUnit itself.  I haven't checked the console yet, but I would assume the adapter and the console interpret the results differently.

The adapter looks at the assertions reported.  
For the code given as:
(Setup): Assert.Warn("...")
(Test): throw new Exception()
(TearDown): Assert.Warn("...")
we get the following back from NUnit:
![image](https://user-images.githubusercontent.com/203432/33225881-8cdc14b0-d180-11e7-9aae-4e9b23b1cc47.png)
This seems to be as it should, but when we look at the other, that goes:
(Test): throw new Exception()
(TearDown): Assert.Warn("...")
we get only the teardown warning, the exception is now *not* reported as an assertion error.
![image](https://user-images.githubusercontent.com/203432/33225892-c28420d0-d180-11e7-949c-1d915c5319b3.png)

[...]

PR #437, opened 2018-01-21
Fix for issue #419

OsirisTerje:
*** DO NOT MERGE **

Issue #419 is partially fixed.

Updated to NUnit 3.9.0
Updated/consolidated other packages.
Fixed attachment test case.


build.cake
```diff
@@ -12,7 +12,7 @@ var configuration = Argument("configuration", "Release");
 //////////////////////////////////////////////////////////////////////
 
 var version = "3.10.0";
-var modifier = "";
+var modifier = ".2";
```
OsirisTerje: I know........

CharliePoole: You seem to be using modifier in a different way than intended. Is this what you have done all along? Normally, I would have said version = "3.10.0.2" and modifier ="" (or "beta" or "rc-2" etc.).

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -116,8 +116,15 @@ public IList<VSTestResult> GetVSTestResults(XmlNode resultNode)
...
             }
 
-            return results;
+            return new TestResultSet {TestCaseResult = testcaseResult, TestResults = results};
+        }
+
+        public struct TestResultSet
+        {
+            public IList<VSTestResult> TestResults { get; set; }
+            public TestResult TestCaseResult { get; set; }
```
CharliePoole: This all gets a bit confusing without a prefix to distinguish NUnit from VS results.

build.cake
OsirisTerje: I use the modifier to get an increasing number, then I use the configuration Debug, which causes the script to add the suffix '-dbg'   I only use this during development, and clear this before commit, which I just forgot here.    If you set modifier = beta,  you get 3.10.0beta-dbg, note the missing dot between 0 and beta.

src/NUnitTestAdapter/TestConverter.cs
OsirisTerje: Good point, I'll fix.

OsirisTerje:
If you look in the method TestFinished, you'll see I get the result from the testcase, and use that for the  overall test result outcome, then I just add the assertion results as result outputs, but they don't change the outcome.  
I work by the assumption you stated earlier, that NUnit will always return the outcome for the testcase, instead of checking all the assertions.
(If this assumption don't hold, we miss a testcase for that)

CharliePoole:
I think relying on NUnit's outcome for each test is all you can do. If NUnit itself is wrong, then we would have to fix it. WRT the individual assertion results, I could not see where you were getting them in your changes - is this in the part you didn't change?

build.cake
CharliePoole: I see, of course there's no problem with what you do in development.

With the modifiers, we actually used "-beta" or "-rc-2" when we were using them.

OsirisTerje:
Yes, that code is unchanged, each assertion is added as a result to the result collection, plus a recordresult is sent to VS.

OsirisTerje:
No further comments, so I'll merge it.

Issue #310, opened 2017-03-27
Consistent Category Display in Test Explorer Window


sailro:
Hi,

When mixing tests using different test frameworks, it would be great to have consistency when displaying "by traits" in the test explorer window.

See the following image:
MSTest is using {Category}
NUnit adapter is using Category [{Category}]

![image](https://cloud.githubusercontent.com/assets/638167/24370465/5eee0904-12dc-11e7-8c20-26f77f8b2bf7.png)

Regards
Seb

OsirisTerje:
Now that MSTest have been open-sourced we can have a look what they have done.  Following the standard Traits mechanism give the results you see for NUnit, but MSTest seems to have another way for categories.

sailro:
Thanks!

PR #446, opened 2018-02-03
Changed categories from traits to category properties, aligning them …

OsirisTerje:
…with what MSTest does, and VSTest supports.  #310 

Handing of categories is moved from the traits class, which where extension methods, into its own class CategoryList.   
NUnit delivers all traits as property/value pairs, and those who are marked as Category are now processed as properties.  The rest are unchanged. 
The tests have been updated to check for both categories and traits.


OsirisTerje:
@CharliePoole Good point, checking.....
image

Seems to work fine :-)

Issue #296, opened 2017-01-14
Mono.Cecil causes OutOfMemoryException with new .csproj PDBs


jnm2:
I converted the old-style test .csproj targeting .NET Framework 4.6.2 to a new-style .csproj targeting the exact same .NET Framework 4.6.2 (VS2017 build 15.0.26020.0):

```xml
<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
  <PropertyGroup>
    <TargetFramework>net462</TargetFramework>
    <RuntimeIdentifier>win</RuntimeIdentifier>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="**\*.cs" />
    <EmbeddedResource Include="**\*.resx" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="NUnit" Version="3.6.0" />
  </ItemGroup>
</Project>
```

All tests run exactly the same as they always have using the NUnit console runner, but no tests are discovered using the NUnit VS adapter. I found out why- Mono.Cecil is causing OutOfMemoryExceptions while reading the PDB to find source code locations:

```
   at Microsoft.Cci.Pdb.MsfDirectory..ctor(PdbReader reader, PdbFileHeader head, BitAccess bits)
   at Microsoft.Cci.Pdb.PdbFile.LoadFunctions(Stream read, Dictionary`2& tokenToSourceMapping, String& sourceServerData, Int32& age, Guid& guid)
   at Mono.Cecil.Pdb.PdbReader.PopulateFunctions()
   at Mono.Cecil.Pdb.PdbReader.ProcessDebugHeader(ImageDebugDirectory directory, Byte[] header)
   at Mono.Cecil.ModuleDefinition.ProcessDebugHeader()
   at Mono.Cecil.ModuleDefinition.ReadSymbols(ISymbolReader reader)
   at Mono.Cecil.ModuleReader.ReadSymbols(ModuleDefinition module, ReaderParameters parameters)
   at Mono.Cecil.ModuleReader.CreateModuleFrom(Image image, ReaderParameters parameters)
   at Mono.Cecil.ModuleDefinition.ReadModule(Stream stream, ReaderParameters parameters)
   at Mono.Cecil.ModuleDefinition.ReadModule(String fileName, ReaderParameters parameters)
   at NUnit.VisualStudio.TestAdapter.NavigationDataProvider.CacheTypes(String assemblyPath)
[...]

PR #450, opened 2018-02-10
Test in all target environments using the correct version of the NUnit VSTest adapter

jnm2:
Blocking my work on https://github.com/nunit/nunit3-vs-adapter/issues/296. This was complex enough to peel away into a targeted PR.

From my findings in https://github.com/nunit/nunit3-vs-adapter/compare/jnm2/replace_cecil:

- Any code that uses DiaSession must run in an environment that has already loaded a version of Microsoft.VisualStudio.TestPlatform.ObjectModel.dll which is newer than the one we must compile against for net35. The one we compile against tries to load msdia110typelib_clr0200.dll from the VS2012 installation. While we must compile against it, we must not distribute it or run our tests against it.

- We got away with running tests against it in the past, but so many of our tests require navigation data (even unrelated tests) that we will not be able to continue with this misconfiguration once we start using TP.ObjectModel.dll's DiaSession constructor. This includes both direct referencing or pure reflection.

- Self-executing adapter tests are out of the question unless we write a host that locates VSTest from `.dotnet` or `vswhere` and loads the correct `Microsoft.VisualStudio.TestPlatform.ObjectModel.dll`, or unless we barely run any tests at all. The point of our VSTest adapter has nothing to do with self-executing tests. Running those few tests that can pass is merely duplicating the testing already done by the tests which we *do* run using VSTest. Self-executing tests are complexity with no benefit.

- Since we publish a VSTest adapter, what we do care about is testing it in every environment that might use it. Ones I know about currently include:
  * VSTest, from VS installation or dotnet SDK (dotnet vstest), against any .NET Framework or .NET Core assembly
  * dotnet test, against any .NET Framework or .NET Core project
  
  If VSTest supports frameworks other than .NET Framework and .NET Core such as UWP (or when VSTest finally supports .NET Standard test assemblies), we should look at expanding that someday.

Therefore, I removed our `TestAdapter` and `TestAdapterNetCore` self-executing test targets and expanded our `TestAdapterUsingVSTest` into four targets:

- `VSTest-net45`
- `VSTest-netcoreapp1.0`
- `DotnetTest-net45`
- `DotnetTest-netcoreapp1.0`

As soon as the self-executing .NET Core tests were gone, we no longer needed the publish step. I was thus able to undo that part of 3bf25c82b6e4d9b8cfdb7cebbb9302fb52238de5. Test failures forced me to undo other parts. With the new test targets, the folder nesting level for the assemblies built for .NET Core now matches .NET Framework.

Does this change seem sound?


jnm2:
@OsirisTerje I have permission to merge without review. I'm so used to clicking it as soon as it turns green that I'm afraid I may cause an accident.

CharliePoole:
@jnm2 @OsirisTerje Just a thought... Being allowed to do something by a software system is not equivalent to "what you are supposed to do." That's one reason I drag my feet as far as automating various "supposed to do" things like this, like standards, etc. I believe the best teams are grown when people actually talk to one another rather than relying on software permissions.

On a more practical note... only owners (me and Rob) and admins (set by each project's lead) are allowed to __override__ the review requirement. It's possible that @OsirisTerje has not set up the review requirement for that repository. Recollect that we did agree that such things were up to each project lead.

FWIW if __I__ were the project lead, I would __not__ set up the GitHub review requirement because it leads to the bad situation that @jnm2 describes - people get used to the system telling them what it's OK to do. See paragraph 1. Of course, that's just me. 😸

jnm2:
Hey, I just realized that this isn't ready to merge yet. The second commit starts skipping all net45 tests because on net45 the engine loads the tests in a different appdomain and it can't find `Microsoft.VisualStudio.TestPlatform.ObjectModel.dll`.

I can't think of any way to make this work other than a .runsettings file containing `<DisableAppDomain>`. (The engine doesn't appear to provide any chance of obtaining a reference to the appdomain after creating it and before loading assemblies into it, and `ImageRequiresDefaultAppDomainAssemblyResolver` [doesn't have any effect](https://github.com/nunit/nunit-console/blob/91ef2ae1b1077ad450d27667c6c4b3ec84b8cdf5/src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs#L43-L45) because the engine is itself already loaded into a secondary appdomain.)

Running the tests in the same appdomain as the engine seems to cause some remoting URL conflicts. I'll try to see what that's about.

jnm2:
Hey, WIPbot showed up! How cool is that! 😄

jnm2:
Starting to look like we'll need a new release of the engine which allows you to override the default URL of the test agency before starting it without having to know about and set up every other service on your own.

jnm2:
Got it working:
<details>
```cs
internal static class TestAdapterUtils
{
    public static ITestDiscoverer CreateDiscoverer()
    {
        var discoverer = new NUnit3TestDiscoverer();
        InitializeForTesting(discoverer);
        return discoverer;
    }

    public static ITestExecutor CreateExecutor()
    {
        var executor = new NUnit3TestExecutor();
        InitializeForTesting(executor);
        return executor;
    }

    private static void InitializeForTesting(NUnitTestAdapter adapter)
    {
#if NET45
        adapter.InternalEngineCreated += engine =>
        {
            engine.Services.Add(new SettingsService(true));
            engine.Services.Add(new DomainManager());
            engine.Services.Add(new ExtensionService());
            engine.Services.Add(new DriverService());
            engine.Services.Add(new RecentFilesService());
            engine.Services.Add(new ProjectService());
            engine.Services.Add(new RuntimeFrameworkService());
            engine.Services.Add(new DefaultTestRunnerFactory());
            engine.Services.Add(new TestAgency("TestAgency for " + TestContext.CurrentContext.Test.Name, 0));
            engine.Services.Add(new ResultService());
            engine.Services.Add(new TestFilterService());
        };
#endif
    }
}
```
</details>

src/NUnitTestAdapterTests/TestAdapterUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+                engine.Services.Add(new DomainManager());
+                engine.Services.Add(new ExtensionService());
+                engine.Services.Add(new DriverService());
+                engine.Services.Add(new RecentFilesService());
+                engine.Services.Add(new ProjectService());
+                engine.Services.Add(new RuntimeFrameworkService());
+                engine.Services.Add(new DefaultTestRunnerFactory());
+                engine.Services.Add(new TestAgency("TestAgency for " + TestContext.CurrentContext.Test.Name, 0));
+                engine.Services.Add(new ResultService());
+                engine.Services.Add(new TestFilterService());
```
jnm2: The engine needs to provide some way to override a service before starting all the services without forcing you to maintain an almost identical list like this.

jnm2:
Got it working! 🎉 Ready for review!

jnm2:
I locally rebased https://github.com/nunit/nunit3-vs-adapter/compare/jnm2/replace_cecil on a test merge of this PR and it's looking very positive! This PR was the hard part! 😊

Only thing left is some of the navigation tests like the async ones. I can look at we we had before https://github.com/nunit/nunit3-vs-adapter/pull/186.

src/NUnitTestAdapterTests/TestAdapterUtils.cs
ChrisMaddock: I added something like this in a recent test. I think it needed `InternalsVisibleTo`...

ChrisMaddock: Oh. Looks like it never got committed. 😞 https://github.com/nunit/nunit-console/pull/320#issuecomment-359180326

jnm2:
I discovered that [there is a difference](https://github.com/nunit/nunit3-vs-adapter/issues/296#issuecomment-365484701) between the IDE's vstest and dotnet vstest when it comes to .NET Framework tests, so we need to be testing:

- `vswhere`…`vstest.console.exe`, both target frameworks
- `dotnet vstest`, both target frameworks
- `dotnet test`, both target frameworks

This is done in the last commit.

jnm2:
NuGet timed out on VSTS. I queued a new build which succeeded but didn't update the status of the commit.

jnm2:
I'm not *absolutely* sure that we need to run all tests against `dotnet vstest` because `dotnet test` ends up calling the same stuff. But the tests are so fast, why not?

jnm2:
Hooray, I have nothing left to do! The Mono.Cecil removal PR is waiting on this one at https://github.com/nunit/nunit3-vs-adapter/pull/454.

Issue #47, opened 2015-07-21
Test Adapter runs explicit tests when TFS TestCaseFilter is used


CharliePoole:
See the original issue filed against  the 2.0 adapter: nunit/nunit-vs-adapter#71

CharliePoole:
The 2.0 issue, nunit/nunit-vs-adapter#71, suggests two ways to deal with this problem:

1. Translate the TFS filter to an NUnit Filter.

2. Review the list of test names received from the filter after they are discovered by NUnit and remove any that are marked explicit.

Upon further review, (2) is not a solution. Tests need to be removed only if the filter does not explicitly select them. So, we have to parse the filter expression in either case... for (1) so we can generate the nunit filter and for (2) so we can retroactively correct the selection made by the filter in the first place. If we have to do so much work for (2) then I would lean toward (1) as the best approach.

That decided, however, another big choice looms!

In the V2 adapter, we implemented the use of MsTest-style filtering because it was the easiest approach. NUnit V2 didn't have it's own filtering language.

But in NUnit 3.0, we already have a filtering DSL, which is used by the console runner and is implemented as a service of the TestEngine. In fact, we could trivially transform an expression like "cat==Data || Priority == High" into a set of NUnit test filters.

We could also, with a bit more work, make a copy of our existing code to parse our DSL and change it to parse MsTest-style filters. As with our existing code, we would emit an nunit filter that honored the Explicit attribute. Of course, this would give us two DSLs to support and for users to remember.

Options:
1. Continue to use the MsTest filter language but parse it ourselves rather than leaving it to VS.
2. Replace it with our own DLS as used in the --where option of the console runner. This could be justified by the fact that the breaking change is being made prior to the 3.0 release.
3. Support both, which is obviously more work.
4. Support a hybrid of the two by allowing both MsTest-style expressions and NUnit expressions in the filter. This is probably the most work and might confuse users.

@OsirisTerje @rprouse What do you think?

jnm2:
I can't get TestCategoryFilter to function at all in TFS 2015.1 build vNext with NUnit adapter 3.0.9.0. Is this related?

CharliePoole:
@jnm2 No, this issue is a pretty well-known problem. I haven't tried vNext at all. If the problem persists, you may want to file another issue, showing details of what you are doing and what is happening (or not).

CharliePoole:
Any rational way forward on this issue involves parsing the filter expression.

[...]

PR #457, opened 2018-02-21
Refactor TraitsFeature

jnm2:
When I came to implement the Explicit category for https://github.com/nunit/nunit3-vs-adapter/issues/47, I found TraitsFeature a bit daunting to wrap my head around.

The addition of tracking the Explicit category would introduce even more complexity, especially because I wanted to easily switch it between a trait, a category, and a brand new property. The cached representation `List<Trait>` was not very easy to tinker with, the way it was mapped back and forth to TestCases and from XmlNodes. So I did a spike.

This PR introduces `TestTraitInfo` as a way to isolate the logic of parsing, caching, combining, and applying of test properties from each other. I expect a performance increase due to several choices in the new implementation, but simplicity was the real goal.

So, for example, this would be the entire implementation of the new Explicit category: https://github.com/nunit/nunit3-vs-adapter/commit/0261bf65f633d777123f90b62d07749c9014f6a3

@OsirisTerje Hopefully the commit messages are good explanations. If this is too many changes, that's no problem. It's just an idea.


jnm2:
@OsirisTerje What do you say—skip this for 3.10 and do the equivalent of https://github.com/nunit/nunit3-vs-adapter/commit/0261bf65f633d777123f90b62d07749c9014f6a3?

OsirisTerje:
Yes, that would be awesome.  

About the refactoring, some of it looks good, some of it I don't understand, and some are in between :-)  Let's talk a bit about it before we go further, but please get  0261bf6  in !

jnm2:
Alright, I'm on it!

jnm2:
Now it'll take a bit of cleverness for me to get 0261bf6 in without this PR and it won't be nearly as clear, but I'm on it!

jnm2:
Wait a minute... that was a lot easier than I though it was going to be 😄

jnm2:
Haha, I don't know why I thought it was complicated before... 😄 PR is up already! https://github.com/nunit/nunit3-vs-adapter/pull/467

jnm2:
This got stale and may be superseded by @OsirisTerje's plans.

PR #469, opened 2018-03-03
Upgrade Cake from 0.22 so that error messages are readable

jnm2:
Cake 0.22 has a bug where all exceptions got wrapped in AggregateException , effectively preventing you from seeing the error message. It kept getting in the way.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -31,7 +31,7 @@ Param(
     [string[]]$ScriptArgs
 )
 
-$CakeVersion = "0.22.1"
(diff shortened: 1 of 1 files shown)


jnm2:
@OsirisTerje I always delete my branches after my PRs are merged. You absolutely don't have to, but if you feel like doing deleting them after merging, please feel free. 😃

OsirisTerje:
@jnm2 Cool,  was just a  bit too careful there.  Thanks for deleting it.

PR #471, opened 2018-03-03
Block explicit tests for 3.10

jnm2:
The rest of the 3.10 work for #47.

I used a new, **non-category**, **non-trait**, separate, hidden bool TestProperty. This is cleaner than re-using the display values we put in the Category or Traits collections; otherwise, we'll detect NUnit categories or properties named Explicit as the run state when they are not, and block them too.

The last commit is the minimum possible part of the refactor https://github.com/nunit/nunit3-vs-adapter/pull/457 to remove the otherwise necessary ugly hack. I wanted to go further and do the same thing to the Category property, but that just would be starting to redo the same work in https://github.com/nunit/nunit3-vs-adapter/pull/457. I still want to do https://github.com/nunit/nunit3-vs-adapter/pull/457 ASAP post-3.10.

### DIFF
--- a/src/NUnitTestAdapter/TraitsFeature.cs
+++ b/src/NUnitTestAdapter/TraitsFeature.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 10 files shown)


OsirisTerje:
This commit is very big.   Is this a mix of code from multiple things?   The implementation of the hidden property looks good, but there is a lot of other stuff here I don't really get, my bad.   Can you explain for me what it is doing?    
I might need to pull down this branch and see what is up here before I can properly review it.  
I see the PR also reports it has conflicts with master now.

jnm2:
Rebased on master.

No problem! This is not a mix of code from other things. Besides the XML doc typo, everything here was essential to the approach I took. I wanted to use the list of commits to document why each change was necessary as far as I could see.

I am happy to talk about what I did or try alternatives. 😊 Explaining the entire diff at once might be hard for me. Is there a specific change in a specific commit that I can start with?

OsirisTerje:
LTGM

jnm2:
@OsirisTerje I added the documentation. Does that look clear enough for future readers?

src/NUnitTestAdapter/TraitsFeature.cs
```diff
@@ -35,10 +35,32 @@ public static void AddTrait(this TestCase testCase, string name, string value)
...
         private const string NunitTestCategoryLabel = "Category";
-       
-       
+
+
+        /// <summary>
+        /// Stores the information needed to initialize a <see cref="TestCase"/>
+        /// which can be inherited from an ancestor node during the conversion of test cases.
+        /// </summary>
+        public sealed class CachedTestCaseInfo
```
jnm2: Maybe a good name for this is InheritableTestInfo?

OsirisTerje:
Good to go , thanks !

PR #477, opened 2018-03-07
Publishing SourceLinked PDBs

### LINKED ISSUE #461: Publishing symbols with releases
To repeat the blurb from https://github.com/nunit/nunit/issues/2421:

> This will affect the team's day jobs and side projects since it will be easy to drop into NUnit code any time you want to understand what's going on. It will also affect NUnit users who may return the courtesy by taking some initiative in diagnosing bugs they observe. Creating a repro is much easier when you aren't dealing with a black box!

While wanting to step into adapter code might be less common than wanting to step into the framework, 
it's still something that would benefit those of us who do.

What do you think of packaging either a GitLinked or source-linked PDB for the adapter assembly into the VSIX and NuGet packages?

I'm happy to do (a few) size comparisons like I did at https://github.com/nunit/nunit/issues/2421#issuecomment-358849839. We went with GitLink for the framework which increased the nupkg size to 2.51× the original.

jnm2:
Closes #461

### DIFF
--- a/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
+++ b/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
@@ -7,6 +7,8 @@
     <TargetFrameworks>net35;netcoreapp1.0</TargetFrameworks>
     <CodeAnalysisRuleSet>..\..\Osiris.Extended.ruleset</CodeAnalysisRuleSet>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <SourceLinkOriginUrl>https://github.com/nunit/nunit3-vs-adapter</SourceLinkOriginUrl>
+    <SourceLinkCreate>true</SourceLinkCreate>
   </PropertyGroup>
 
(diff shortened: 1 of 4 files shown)


src/NUnitTestAdapter/NUnit.TestAdapter.csproj
```diff
@@ -7,6 +7,8 @@
     <TargetFrameworks>net35;netcoreapp1.0</TargetFrameworks>
     <CodeAnalysisRuleSet>..\..\Osiris.Extended.ruleset</CodeAnalysisRuleSet>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <SourceLinkOriginUrl>https://github.com/nunit/nunit3-vs-adapter</SourceLinkOriginUrl>
```
jnm2: Necessary because the `origin` remote URL is not necessarily `https://github.com/nunit/nunit3-vs-adapter`.

OsirisTerje:
LGTM

PR #480, opened 2018-03-07
Copy adapter PDB to project build output

jnm2:
I didn't realize until too late that we are the ones responsible for copying the adapter to build output since it's not in `/lib`. This is totally my bad. I tried it on various configurations of csproj but a .NET Framework target was always the one listed first, and I should have tried it on a .NET Core-only project.

Source stepping works with no fuss for any .NET Framework project because the adapter is executed from `%temp%\VisualStudioTestExplorerExtensions\{version}` or from the VSIX installation folder, but with .NET Core, the adapter is executed from the build output folder.

Copying the PDB out beside the adapter assembly is the only way I know to allow source-stepping to work, besides telling the user to locate `%userprofile%/.nuget/packages/nunit3testadapter/{version}/build/{framework}/NUnit3.TestAdapter.pdb` themselves.

I'll update the docs with the correct workaround until the next release.

### DIFF
--- a/nuget/net35/NUnit3TestAdapter.props
+++ b/nuget/net35/NUnit3TestAdapter.props
@@ -6,6 +6,11 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>False</Visible>
     </Content>
+    <Content Include="$(MSBuildThisFileDirectory)NUnit3.TestAdapter.pdb">
+      <Link>NUnit3.TestAdapter.pdb</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Visible>False</Visible>
+    </Content>
     <Content Include="$(MSBuildThisFileDirectory)nunit.engine.dll">
       <Link>nunit.engine.dll</Link>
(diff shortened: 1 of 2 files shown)


jnm2:
Just like with the NUnit framework instructions where you have to add something to your csproj to get the PDB to copy, this is what you'd add here:

```xml
  <Target Name="IncludeAdapterSymbolsInOutput" BeforeTargets="BeforeBuild">
    <ItemGroup>
      <None Include="@(FileDefinitions->'%(ResolvedPath)')"
            Condition="'%(FileDefinitions.PackageName)' == 'NUnit3TestAdapter' and '%(FileName)%(Extension)' == 'NUnit3.TestAdapter.pdb'"
            CopyToOutputDirectory="PreserveNewest" />
    </ItemGroup>
  </Target>
```

I'll update the docs since this is the way it'll be for 3.10 if you need to step into adapter source while debugging .NET Core projects. Admittedly, may not be a common scenario. 🙂

jnm2:
That's cool! Updated docs at https://github.com/nunit/docs/wiki/Adapter-Source-Stepping/260f4f0b3b01e59165af970d7e54e945b00df195 since the problem will not exist in the initial release.

PR #493, opened 2018-03-26
Adding Category Filter for Category Trait

abhishkk:
## Description
We can pass filters using `/TestCaseFilters` argument while running tests from command line (vstest.console.exe).

NUnit understands `TestCategory `as a filter but not `Category`.
Example: `/TestCaseFilter:TestCategory=SampleValue` works but `/TestCaseFilter:Category=SampleValue` doesn't work.

Internally NUnit maps `TestCategory` filter to `Category` trait.
So while writing test, user needs to write Category trait but while running tests user needs to pass TestCategory as filter.
Example:
```
        While writing: 

        [TestCase]
        [Category("SampleValue")]
        public void TestMethod1()
        {
            Assert.AreEqual(30, 30);
        }
       
        While testing:
        vstest.console.exe NUnitTestProject.dll /TestCaseFilter:TestCategory=SampleValue
```

## Issue
While running NUnit tests from [vstest task](https://github.com/Microsoft/vsts-tasks/tree/master/Tasks/VsTest), we can run tests in:
1. Single agent i.e. non-distributed flow
2. Multiple agents i.e. distributed flow
3. Test run with TIA (test impact analysis) as true

For 2nd and 3rd scenario, there comes a requirement in which tests needs to be filtered out at discovery time. As discovery time filtering was not supported for adapters till now (its [supported](https://github.com/Microsoft/vstest/pull/1074) now), TestPlatform filters tests at discovery on its own for these scenarios. 

While doing filtering at discovery time, TestPlatform has exact mapping of Filters to attributes. i.e. For Category attribute, Category filters is understood.

Thus behavior change in TestPlatform side filtering(understand Category filter for Category attribute) and adapter side filtering(understand TestCategory filter for Category attribute) forces users to use
1. TestCategory as filter for running test in single agent (as discovery time filtering is not done by TestPlatform here. Filtering is done by NUnit adapter at execution time)
2. Category as filter for running tests in multiple agents or when TIA is enabled. (as discovery time filtering is done by TestPlatform here)

## Solution
Adding Category filter in NUnit adapter as well. Now both Category and TestCategory filters will be understood by NUnit and will be internally mapped to Category attribute.

So now users can pass either Category or TestCategory and both will work.

## Note
I am planning to add discovery time filtering as well in near future. I will get back soon with that PR as well.

### DIFF
--- a/src/NUnitTestAdapterTests/Filtering/TestFilterTests.cs
+++ b/src/NUnitTestAdapterTests/Filtering/TestFilterTests.cs
@@ -40,6 +40,8 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Filtering
             Assert.NotNull(prop);
             prop = TfsTestFilter.PropertyProvider("TestCategory");
             Assert.NotNull(prop);
+            prop = TfsTestFilter.PropertyProvider("Category");
+            Assert.NotNull(prop);
         }
 
         [Test]
@@ -47,6 +49,8 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Filtering
         {
             var testFilter = new TfsTestFilter(null);
             var trait = TfsTestFilter.TraitProvider("TestCategory");
+            Assert.NotNull(trait);
+            trait = TfsTestFilter.TraitProvider("Category");
             Assert.NotNull(trait);
         }
 
@@ -68,7 +72,7 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Filtering
         }
 
         [Test]
-        public void PropertyValueProviderCategoryWithOneCategory()
+        public void PropertyValueProviderWithOneCategoryAndTestCategoryAsFilter()
         {
             var tc = new TestCase("Test1", new Uri("executor://NUnitTestExecutor"), "NUnit.VSIX");
             tc.AddTrait("Category", "CI");
@@ -78,7 +82,17 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Filtering
         }
 
         [Test]
-        public void PropertyValueProviderCategoryWithNoTraits()
+        public void PropertyValueProviderWithOneCategoryAndCategoryAsFilter()
+        {
+            var tc = new TestCase("Test1", new Uri("executor://NUnitTestExecutor"), "NUnit.VSIX");
+            tc.AddTrait("Category", "CI");
+            var testFilter = new TfsTestFilter(null);
+            var obj = TfsTestFilter.PropertyValueProvider(tc, "Category");
+            Assert.AreSame("CI", obj);
+        }
+
+        [Test]
+        public void PropertyValueProviderWithNoTraitsAndTestCategoryAsFilter()
         {
             var tc = new TestCase("Test1", new Uri("executor://NUnitTestExecutor"), "NUnit.VSIX");
             var testFilter = new TfsTestFilter(null);
@@ -87,7 +101,16 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Filtering
         }
 
         [Test]
-        public void PropertyValueProviderCategoryWithMultipleCategories()
+        public void PropertyValueProviderWithNoTraitsAndCategoryAsFilter()
+        {
+            var tc = new TestCase("Test1", new Uri("executor://NUnitTestExecutor"), "NUnit.VSIX");
+            var testFilter = new TfsTestFilter(null);
+            var obj = TfsTestFilter.PropertyValueProvider(tc, "Category");
+            Assert.IsNull(obj);
+        }
+
+        [Test]
+        public void PropertyValueProviderWithMultipleCategoriesAndTestCategoryAsFilter()
         {
             var tc = new TestCase("Test1", new Uri("executor://NUnitTestExecutor"), "NUnit.VSIX");
             tc.AddTrait("Category", "CI");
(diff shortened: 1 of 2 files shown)


abhishkk:
@pvlakshm  @cltshivash

rprouse:
LGTM, @OsirisTerje?

rprouse:
@abhishkk in the future, we would prefer issues entered for these PRs. It helps us track what goes into each release and also let's us know what you are working on.

OsirisTerje:
Hi @abhishkk .  It would be nice if you started out by raising an issue.  This PR has no issue.  

The adapter matches TestCategory with Category.  Do you say that this is no longer working? 
TIA is not VS, but vstest specific, right?  So how does this relate to TIA?  TIA is run on vstest, and discovery is not being used, we go directly to execution.  
The filter syntax used in VS and the filter used in vstest (CI/TFS/VSTS/Test task) is different. Do you say that this is now harminozed (as requested) or still not?   And if not, will this PR make it work in both cases?

OsirisTerje:
@rprouse It is fine that we add "Category" as a filter term that can be used, but it needs to work both on VS and TFS/VSTS, otherwise we just add confusion.  I would also like to know if we're fixing an issue here, or adding new capabilities.  If we're fixing an issue, we have had no reports to that effect yet.

abhishkk:
@rprouse @OsirisTerje 
> It would be nice if you started out by raising an issue. This PR has no issue.

Sure. I will create an issue for future PRs. Should I create one for this PR as well?


> The adapter matches TestCategory with Category. Do you say that this is no longer working?

NUnit adapter matches TestCategory with Category properly. But it doesn't match Category with Category.

> TIA is not VS, but vstest specific, right? So how does this relate to TIA? TIA is run on vstest, and discovery is not being used, we go directly to execution.

TIA is vstest specific. But in case of TIA, we do discovery time filtering at vstest/TestPlatform level and never comes to adapter for filtering (as discovery time filtering is not supported in NUnit adapter side). 

During filtering at vstest/TestPlatform level, we map Traits and filters with exact names. So we understand Category filter rather than TestCategory filter (as trait name is Category in NUnit).

Following are the fixes we are making to solve this mismatch:
1. For flows(TIA and multi-agent) where we do filtering at vstest/TestPlatform instead of going to adapter, we made a fix in vstest/TestPlatform after which we understand both Category and TestCategory in those flows.
2. For flows(single agent) where filtering is done by adapter, we are making changes in NUnit adapter via this PR after which we will understand both Category and TestCategory in these flows as well.

> The filter syntax used in VS and the filter used in vstest (CI/TFS/VSTS/Test task) is different. Do you say that this is now harmonized (as requested) or still not? And if not, will this PR make it work in both cases?

Filter syntax used in VS and filter used in vstest should ideally match. But because of mismatch discussed above, users were forced to pass TestCategory in single agent flows of vstest and Category in TIA and multi-agent flows. Fixes which we discussed above (including this PR) will make passing filters from vstest and VS in sync. This PR will work with both cases.

> If we're fixing an issue, we have had no reports to that effect yet.

We have multiple issues reported for this. One of them is https://github.com/Microsoft/vsts-tasks/issues/3206

OsirisTerje:
@abhishkk Thanks for the explanations.   

And yes please, raise an issue for this.  When done, I'll add it to the 3.11 milestone.

src/NUnitTestAdapterTests/Filtering/TestFilterTests.cs
```diff
@@ -121,12 +158,30 @@ public void PropertyValueProviderCategoryFail()
         [TestCase("NS11", new[] { "nUnitClassLibrary.NestedClasses.NC11" })]
         [TestCase("NS2", new[] { "nUnitClassLibrary.NestedClasses+NestedClass2.NC21" })]
         [TestCase("NS21", new[] { "nUnitClassLibrary.NestedClasses+NestedClass2.NC21" })]
-        public static void CanFilterConvertedTestsByCategory(string category, IReadOnlyCollection<string> expectedMatchingTestNames)
+        public void CanFilterConvertedTestsByCategoryWithTestCategoryAsFilter(string category, IReadOnlyCollection<string> expectedMatchingTestNames)
```
jnm2: Why you remove my static modifier? 😭 :D
Don't really care. `static` makes it more likely that people will choose to write more functional, parallel-safe tests.

src/NUnitTestAdapterTests/Filtering/TestFilterTests.cs
```diff
@@ -78,7 +82,17 @@ public void PropertyValueProviderCategoryWithOneCategory()
         }
 
         [Test]
-        public void PropertyValueProviderCategoryWithNoTraits()
+        public void PropertyValueProviderWithOneCategoryAndCategoryAsFilter()
+        {
+            var tc = new TestCase("Test1", new Uri("executor://NUnitTestExecutor"), "NUnit.VSIX");
+            tc.AddTrait("Category", "CI");
```
jnm2: Unrelated: I wish we weren't testing by modifying the TestCase object directly because the *way* we apply categories to TestCases may change in the future. @OsirisTerje I'll try to rememeber this when I dust off the refactoring PR.

abhishkk:
Issue: https://github.com/nunit/nunit3-vs-adapter/issues/495

jnm2:
@abhishkk 

> During filtering at vstest/TestPlatform level, we map Traits and filters with exact names. So we understand Category filter rather than TestCategory filter (as trait name is Category in NUnit).

I would prefer that VSTest mapped the category attribute to TestCategory traits at the vstest/TestPlatform level. NUnit doesn't have traits; it has properties, and they are not translated 1:1 into traits for different reasons.

jnm2:
@abhishkk Just saw your comment. Will move my question to the issue. Thanks for doing that! 😊

PR #499, opened 2018-04-03
Use PackageReference instead of packages.config in VSIX project

### LINKED ISSUE #451: C# 7
While working on https://github.com/nunit/nunit3-vs-adapter/issues/296, I'm running into cases where I want to use default literals or `?? throw`  or `is var` and `out var`. For example:

```cs
while (fullName.TryGetNextNamespaceSegment() is var segment && segment != null)
{
    // ...
}
```

Instead of:

```cs
for (;;)
{
    var segment = fullName.TryGetNextNamespaceSegment();
    if (segment == null) break;
    // ...
}
```

Since none of the projects have an explicit version set, and the solution already requires VS2017, and @oznetmaster doesn't transpile this codebase to C# 3, it seems like C# 7 is guaranteed to be available to anyone in the code. Does this mean we accept C# 7.0 code?

What about using `<LangVersion>latest</LangVersion>` or `7.1` or `7.2` so that I can use default literals or readonly structs? C# 7.1 requires VS 15.3 (or a package reference) and C# 7.2 requires VS 15.5 (or a package reference).

jnm2:
Closes #451

The first commit does the minimum necessary to use PackageReference instead of packages.config.
The other does extra cleanup to make the project file smaller.


src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj
```diff
@@ -1,24 +1,16 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Microsoft.VSSDK.BuildTools.15.1.192\build\Microsoft.VSSDK.BuildTools.props" Condition="Exists('..\..\packages\Microsoft.VSSDK.BuildTools.15.1.192\build\Microsoft.VSSDK.BuildTools.props')" />
+﻿<Project ToolsVersion="15.0">
```
jnm2: Since we've already decided to support only VS2017

src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj
```diff
@@ -17,7 +14,7 @@
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>NUnit3TestAdapter</RootNamespace>
-    <AssemblyName>NUnit3TestAdapter</AssemblyName>
+    <TargetName>NUnit3TestAdapter</TargetName>
```
jnm2: NuGet restore gets confused when two projects have the same AssemblyName. Setting the TargetName to control the output .vsix name is more clear anyway since we're not even compiling an assembly in the VSIX project.

src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj
```diff
@@ -1,24 +1,16 @@
...
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <SchemaVersion>2.0</SchemaVersion>
     <ProjectTypeGuids>{82b43b9b-a64c-4715-b499-d71e9ca2bd60};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
     <ProjectGuid>{D6990436-E3BA-4AC1-8A7D-9FFF9C3B19AD}</ProjectGuid>
     <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>NUnit3TestAdapter</RootNamespace>
-    <AssemblyName>NUnit3TestAdapter</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetName>NUnit3TestAdapter</TargetName>
```
jnm2: NuGet restore gets confused when two projects have the same AssemblyName. Setting the TargetName to control the output .vsix name is more clear anyway since we're not even compiling an assembly in the VSIX project.

src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj
```diff
@@ -27,27 +19,11 @@
     <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
     <DeployExtension>False</DeployExtension>
     <IsProductComponent>false</IsProductComponent>
```
jnm2: It seems like the eight lines above could be reduced, but I don't know enough about VSIX projects to want to try yet.

src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj
```diff
@@ -27,27 +19,11 @@
...
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <DebugType>pdbonly</DebugType>
-    <Optimize>true</Optimize>
-    <OutputPath>..\..\bin\Release\</OutputPath>
-    <DefineConstants>TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <DeployExtension Condition="'$(AppVeyor)' != ''">False</DeployExtension>
-  </PropertyGroup>
```
jnm2: Since we aren't compiling an assembly...

src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj
```diff
@@ -27,27 +19,11 @@
...
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
```
jnm2: This is still necessary so that VS can detect which configurations and platforms to show in Configuration Manager. If these conditions are removed, VS changes the solution file to list this project as `x86`.

src/NUnit3TestAdapterInstall/Properties/AssemblyInfo.cs
```diff
@@ -1,33 +0,0 @@
...
-//      Major Version
-//      Minor Version
-//      Build Number
-//      Revision
-//
-// You can specify all the values or you can default the Build and Revision Numbers
-// by using the '*' as shown below:
-// [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
```
jnm2: Since we're not compiling an assembly in this project

src/NUnit3TestAdapterInstall/app.config
```diff
@@ -1,11 +0,0 @@
...
-<configuration>
-	<runtime>
-		<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
-			<dependentAssembly>
-				<assemblyIdentity name="nunit.engine" publicKeyToken="2638cd05610744eb" culture="neutral"/>
-				<bindingRedirect oldVersion="0.0.0.0-3.0.5797.27553" newVersion="3.0.5797.27553"/>
-			</dependentAssembly>
-		</assemblyBinding>
-	</runtime>
-	<startup/></configuration>
```
jnm2: Since we're not compiling an assembly in this project, nor are we dealing with versions of the engine < 3.0.5797.

build.cake
```diff
@@ -103,13 +103,12 @@ Task("NuGetRestore")
...
-    NuGetRestore(PROJECT_DIR + "src/NUnit3TestAdapterInstall/NUnit3TestAdapterInstall.csproj",
-                 new NuGetRestoreSettings {
-                     PackagesDirectory = PROJECT_DIR + "packages"
-                 });
+    MSBuild(ADAPTER_SOLUTION, new MSBuildSettings
+    {
+        Configuration = configuration,
+        Verbosity = Verbosity.Minimal,
+        ToolVersion = MSBuildToolVersion.VS2017
+    }.WithTarget("Restore"));
```
jnm2: Neither DotNetCoreRestore nor NuGetRestore are happy about the VSIX project, but MSBuild has become my default ever since NuGet was integrated into the MSBuild SDK.

jnm2:
Merging, per @OsirisTerje.

PR #500, opened 2018-04-10
[WIP]NUnit3VsAdapter to support managed sources only

mayankbansal018:
Following RFC: https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0020-Improving-Logic-To-Pass-Sources-To-Adapters.md

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -25,7 +25,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
+using System.ComponentModel;
 using System.IO;
 using System.Linq;
 using System.Xml;
@@ -43,6 +43,7 @@ namespace NUnit.VisualStudio.TestAdapter
     [FileExtension(".dll")]
     [FileExtension(".exe")]
     [DefaultExecutorUri(NUnit3TestExecutor.ExecutorUri)]
+    [Category("managed")]
     public sealed class NUnit3TestDiscoverer : NUnitTestAdapter, ITestDiscoverer
     {
         private Dump.DumpXml dumpXml;
--- a/src/NUnitTestAdapterTests/NUnit3TestDiscovererTests.cs
+++ b/src/NUnitTestAdapterTests/NUnit3TestDiscovererTests.cs
@@ -1,0 +1,41 @@
+﻿// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Terje Sandstrom
+//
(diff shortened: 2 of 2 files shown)


OsirisTerje:
@mayankbansal018 Can you please create an issue for this too, and link to the PR ?

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -43,6 +43,7 @@ namespace NUnit.VisualStudio.TestAdapter
     [FileExtension(".dll")]
     [FileExtension(".exe")]
     [DefaultExecutorUri(NUnit3TestExecutor.ExecutorUri)]
+    [Category("managed")]
```
jnm2: This line is causing `dotnet test --framework=net46` to crash VSTest.Console.exe with:

> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified.

The net46 tests use the net35 adapter, and the net35 adapter is here referencing `System, Version=2.0.0.0`.
I'm not sure why `dotnet test --framework=net46` crashes, while a direct call to VSTest.Console.exe in my Visual Studio installation does not crash. Perhaps the difference is simply in which version of VSTest is running?

/cc @OsirisTerje

jnm2: If we start building with .NET Core SDK 2.0.0, this problem disappears.

2.0.0 is the oldest SDK which solves the problem. With SDK 2.0.0, `dotnet test` runs VSTest version `15.7.0-preview-20180221-13`
1.1.9 is the newest SDK which does not solve the problem. With SDK 1.1.9, `dotnet test` runs VSTest version `15.0.0.0`.

@OsirisTerje This might be a reason to go ahead with https://github.com/nunit/nunit3-vs-adapter/pull/508?

OsirisTerje: @jnm2 Since you have been in here, can you check this now ,  #508 has been merged.

jnm2: Yep, that fixed it. I double-checked by rebasing this branch on master on my machine.

mayankbansal018:
Thanks for merging this

jnm2:
@OsirisTerje Well I can say it rebased without conflicts and `.\build.ps1 -t test` succeeded on my machine...

Issue #489, opened 2018-03-22
VS15.6 Real-Time Test Discovery yields unrunnable NUnit tests


yaakov-h:
In Visual Studio 15.6 (with real-time test discovery), the Visual Studio Test Explorer shows tests that do not exist, or tests with sources. In both cases, the tests cannot be run via the Test Explorer.

For example, in Visual Studio Enterprise v15.6.3 with NUnit Adapter 3.10.0.21 and NUnit Framework 3.10.1, and the following code that targets .NET Framework:

```csharp
using NUnit.Framework;
using System.Collections.Generic;

namespace NUnitSetName
{
    public class Class1
    {
        [Test]
        public void SimpleTest() { }

        [TestCase(1)]
        public void TestCase(int i) { }

        [TestCase(1, TestName = "{m} with a 1")]
        [TestCase(1, TestName = "an overriden test name with a 1")]
        public void NamedTestCase(int i) { }

        [TestCaseSource(nameof(Sauce))]
        public void SourcedTestCase(int i) { }

        public static IEnumerable<TestCaseData> Sauce
        {
            get
            {
                yield return new TestCaseData(1).SetName("totally overriden name");
                yield return new TestCaseData(1).SetName("{m} - enhanced name");
            }
        }
    }
}
```

After opening the solution, the Test Explorer "discovers" tests including `{m}` as a literal:

![image](https://user-images.githubusercontent.com/426009/37749887-3c51bf6c-2dde-11e8-98ea-867fad98656f.png)

After building the solution it runs through the test case sources and shows the proper test names, but does not discard the incorrectly discovered ones that are not actually valid tests:

![image](https://user-images.githubusercontent.com/426009/37749899-49c3737a-2dde-11e8-9b88-554080e68eb4.png)

After running all tests, not all tests were run:

![image](https://user-images.githubusercontent.com/426009/37749910-533e19dc-2dde-11e8-8c08-05a8883a1354.png)

Attempting to select one of the last two tests and running it just yields no result, and a fairly useless messages in the Tests Output window:

```
[22/03/2018 2:39:32 PM Informational] ------ Run test started ------
[22/03/2018 2:39:32 PM Informational] NUnit Adapter 3.10.0.21: Test execution started
[22/03/2018 2:39:32 PM Informational] Running all tests in C:\Temp\NUnitSetName\NUnitSetName\bin\Debug\NUnitSetName.dll
[...]

Issue #423, opened 2017-12-05
Source information can't be obtained in some scenarios in VS 2017 Update 5


Shyam-Gupta:
VS 2017 Update 5 extracts source information for discovered tests using Roslyn APIs. It first determines whether it can obtain this information for the test projects. If yes then it passes following tag in runsettings to the adapter:
<CollectSourceInformation>false</CollectSourceInformation>
NUnit adapter 3.9.0 already honors this setting. 

However, for some scenarios, VS is not able to obtain this information:

Examples :
Case #1

    public class MyDataClass
    {
        public static IEnumerable<TestCaseData> TestCases
        {
            get
            {
                yield return new TestCaseData(12, 3).SetName("NewTestName").Returns(4);
                yield return new TestCaseData(12, 4).Returns(3);
            }
        }
    }
    [TestFixture]
    public class MyTests2
    {
        [Test, TestCaseSource(typeof(MyDataClass), "TestCases")]
        public int DivideTest(int n, int d)
        {
            return n / d;
        }
    }
 
Case #2

        [TestCase(TestName="foobar")]
        public void TestMethod()
        {
        }

**Reason:** In above mentioned cases, the fully qualified name for the test methods as reported by nunit doesn't correspond to a test case present in the source code. Hence using Roslyn APIs SourceInformationProvider is not able to find the test method and hence fails to fetch its source information.

Nunit adapter\engine which sends test method Fully Qualified Name, should be fixed to send correct test method name in above scenarios

CharliePoole:
@Shyam-Gupta I updated your description to use backticks so that the XML is properly displayed in GitHub.

CharliePoole:
One way to fix this if for NUnit to "fake" the FQN as suggested. Is there also the option for us to __provide__ the source information as part of the `TestCase` as we currently do? Would that be a better solution?

Our current code is providing source code information. Unless something has changed, it should be provided in the example cases cited. Can you indicate what is currently provided by the adapter in those cases?

peterwald:
@CharliePoole providing source information from the adapter would certainly work, although that negates the performance improvement that this change was created to provide. 

Another option would be for the adapter to provide the Type & Method name in a well-known property on the test case (in the property bag). Then on the VS side, we can use the Type & Method to do the source information lookup without requiring the adapter to load symbols.

CharliePoole:
[...]

PR #502, opened 2018-04-20
Now using fullname (full test namespace)

qubicaamf:
In order to avoid duplicates on "dotnet test -t" output, the parameter DisplayName should use the test's full namespace.
This way should be good to enable Visual Studio Code (and [Test Explorer Extension](https://github.com/formulahendry/vscode-dotnet-test-explorer)) to identify and match test cases like xUnit.

### DIFF
--- a/src/NUnitTestAdapterTests/IssueNo24Tests.cs
+++ b/src/NUnitTestAdapterTests/IssueNo24Tests.cs
@@ -10,7 +10,6 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
 {
     [Explicit]
     [TestFixture]
-    [Ignore("Temporary disabled...")]
     [Category("LongRunning")]
     class IssueNo24Tests
     {
--- a/src/NUnitTestAdapterTests/IssueNo24Tests.cs
+++ b/src/NUnitTestAdapterTests/IssueNo24Tests.cs
@@ -10,6 +10,7 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
 {
     [Explicit]
     [TestFixture]
+    [Ignore("Temporary disabled...")]
     [Category("LongRunning")]
     class IssueNo24Tests
     {
--- a/src/NUnitTestAdapter/TestConverter.cs
+++ b/src/NUnitTestAdapter/TestConverter.cs
@@ -144,7 +144,7 @@ namespace NUnit.VisualStudio.TestAdapter
                                      new Uri(NUnitTestAdapter.ExecutorUri),
                                      _sourceAssembly)
             {
-                DisplayName = testNode.GetAttribute("name"),
+                DisplayName = testNode.GetAttribute("fullname"),
                 CodeFilePath = null,
                 LineNumber = 0
             };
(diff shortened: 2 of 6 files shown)


rprouse:
This has caused 33 test failures. See the AppVeyor build report.

qubicaamf:
Tests fixed now, ready to be merged.

src/NUnitTestAdapterTests/IssueNo24Tests.cs
```diff
@@ -10,6 +10,7 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
 {
     [Explicit]
     [TestFixture]
+    [Ignore("Temporary disabled...")]
```
OsirisTerje: Why is Ignore added, it is already marked Explicit.  We don't intent to enable this test.

qubicaamf: Sorry, it was a mistake.
I'm going to push again right now.

OsirisTerje:
Please also see issues  #489 and #423.

qubicaamf:
1. No issue was created, actually the main issue was described on the first comment.
2. On a simple namespace like this:
```csharp
using System;
using NUnit.Framework;

namespace NUnitTestProject1
{
	[TestFixture]
	public class UnitTest1
	{
		[Test]
		public void Test1() {}

		[TestCase(TestName = "{C}")]
		public void Test2() {}
	}
}
```
...the output on TE is:
image

Just note the duplicate entry on NUnitTestProject1 and the {C} test under UnitTest1.
Except for {C}, the duplicate entry will appear even if you disable the templating on that Test.

3. As @OsirisTerje described: _This change just makes DisplayName equal to the already existing FullName property of VSTest TestCase._.
As i can see, using the template engine, the TRX (XML) produced will have the name attribute, modified accordingly to the template string selected, on every xml node.
Download and check yourself the following reports (generated using '*dotnet test -l trx*' command, over an example project having NUnitTestProject1 namespace):
[without_template_string_trx.txt](https://github.com/nunit/nunit3-vs-adapter/files/1942541/without_template_string_trx.txt)
[with_template_string_trx.txt](https://github.com/nunit/nunit3-vs-adapter/files/1942542/with_template_string_trx.txt)

### Conclusion
With this PR we can get a better test explorer via console/bash, but still producing errors on Visual Studio 2017 Text Eplorer.
As mentioned on issue #489, there are problems using template strings for test names.

Note that after these changes, the behavior is the same as xUnit in Test Explorer:
image
(*disabled template for testname in this screenshot*)

With this status and at this moment, where a new version of dotnet core is going to be released, maybe is too soon merge this PR.
Speaking of general purposes about the '*dotnet test -t*' command, maybe the right choice should be to request a new parameter/modifier, just to show the already existing FullName instead of name.

OsirisTerje:
@qubicaamf   Thanks for the explanations!  

1.  We have the practice of raising an issue, then apply a PR.   (This also makes release notes easier, as we always refer to issues, not PRs.)
2.  The issue here comes from the new source based discovery in VS 15.6.   If you turn off SBD, then it displays as expected in TE:

image

This issue is related to the two I mentioned above, and must be handled with the upcoming FQN changes and alignment needed there between VSTest, NUnit and XUnit concepts.   

Not quite sure what you mean by duplicates in this case, you have two tests, and two are shown.

Also tried dotnet test -t and it worked as it should. 
image

OsirisTerje:
@rprouse  Agree, and done.

Issue #506, opened 2018-05-18
Test categories aren't propagated to vstest trx logs


andreyns:
With the latest version of the nuni3-adapter unit test descriptions in trx files created with vstest.console (2017) don't have TestCategory elements.

## Expected behaviour
All categories assigned to test cases are reflected in TestCategory elements in trx logs, for example:
```xml
    <UnitTest name="ThatEmptyContainsHeaders" storage="d:\work\nunit3-vs-adapter\src\nunittestadaptertests\bin\debug\netcoreapp1.0\nunit.visualstudio.testadapter.tests.dll" id="5d22f9c8-13f7-4be4-b020-438dbf715a39">
      <TestCategory>
        <TestCategoryItem TestCategory="DumpXml" />
      </TestCategory>
      <Execution id="cf9ffa57-0013-4c9a-877e-5104b828b8ec" />
      <TestMethod codeBase="D:\work\nunit3-vs-adapter\src\NUnitTestAdapterTests\bin\Debug\netcoreapp1.0\NUnit.VisualStudio.TestAdapter.Tests.dll" executorUriOfAdapter="executor://nunit3testexecutor/" className="NUnit.VisualStudio.TestAdapter.Tests.DumpXmlTests" name="ThatEmptyContainsHeaders" />
    </UnitTest>
```

## Repro steps
1. build the nunit3-vs-adapter repo;
2. invoke vstest.console with the following arguments. note the logger argument
```
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" "D:\work\nunit3-vs-adapter\src\NUnitTestAdapterTests\bin\Debug\netcoreapp1.0\NUnit.VisualStudio.TestAdapter.Tests.dll" /logger:"trx" /TestAdapterPath:"D:\work\nunit3-vs-adapter\bin\Debug"
```
3. the test will spit out a new result trx-file. Open it in a text editor and note that none of the test cases have TestCategory items, for example:
```xml
    <UnitTest name="ThatEmptyContainsHeaders" storage="d:\work\nunit3-vs-adapter\src\nunittestadaptertests\bin\debug\netcoreapp1.0\nunit.visualstudio.testadapter.tests.dll" id="c3d0232c-20e9-44be-a6e9-c48b0b4ef694">
      <Execution id="7a1b8f5c-23fd-4562-832f-3febbda631f9" />
[...]

PR #507, opened 2018-05-18
Return test cases categories in a Trx compatible way

andreyns:
#506 a small mstest compatibility patch to fix test cases categories in trx files


rprouse:
This looks reasonable, but if the VSTest converter is fixed in the future this will cause duplicate categories. @OsirisTerje any thoughts?

OsirisTerje:
I would like to hear MS PG comments on this before we act on it.  It is a workaround, but could easily backfire as you mention.  
We could hide it behind a runsettings option, but then again, it would normally not be used by anyone except those familiar with the options.

PR #508, opened 2018-05-23
Proposed .NET Core SDK upgrade to allow running of all tests

jnm2:
The VSTest team made me aware that the registration-free COM error is fixed in the most recent releases of the .NET Core SDK. (https://github.com/Microsoft/vstest/issues/1432) This means we are no longer required to skip navigation data tests when using the `dotnet` CLI.

The first commit updates to SDK 2.1.201 and removes our workaround.

**Question 1:** is it okay to stop testing against .NET Core 1.0 (via SDK 1.1) and test against .NET Core 2.0 (via SDK 2.1)? If not, should we also be testing against .NET Core 1.1 and .NET Core 2.1-rc in addition to the other two officially-supported versions? (https://www.microsoft.com/net/Support/Policy)

<br>

The second commit is because I happened to notice that the version of VSTest run by `dotnet test` (bundled in .NET Core SDK 2.1.201) is 15.7.0-preview-20180221-13, and I remembered VSTest.Console.exe is now designed to be consumed via NuGet so that we no longer need to be dependent on the version of VS which is installed on the user machine (🎉).

**Question 2:** Is there any reason to keep testing against the VSTest installation deployed by the Visual Studio installer instead of the NuGet package?

<br>

The final commit is because the `DotnetVSTest-` tasks are running test runs which are identical to our `DotnetTest-` tasks. All we had been truly testing was that the dotnet CLI is working the way it is documented to work.
Also, the `DotnetVSTest-` test runs are now identical to the `VSTest-` test runs except that the  `VSTest-` tasks now use a newer release version of VSTest.

**Question 3:** What if we dropped our `VSTest-` tasks and just used the `DotnetTest-` tasks? We can do that now that we don't have the COM error and now that the VSTest team seems to be shipping the product the exact same way to NuGet, Visual Studio, and bundled within the .NET Core SDK.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -177,8 +177,6 @@ foreach (var (framework, vstestFramework, adapterDir) in new[] {
             VSTest(GetTestAssemblyPath(framework), settings);
         });
 
-    const string NoNavigationTests = "TestCategory != Navigation";
-
     Task($"DotnetTest-{framework}")
         .IsDependentOn("Build")
         .Does(() =>
@@ -189,8 +187,7 @@ foreach (var (framework, vstestFramework, adapterDir) in new[] {
                 Framework = framework,
                 NoBuild = true,
                 TestAdapterPath = adapterDir,
-                Settings = File("DisableAppDomain.runsettings"),
-                Filter = framework == "net46" ? NoNavigationTests : null
+                Settings = File("DisableAppDomain.runsettings")
             });
         });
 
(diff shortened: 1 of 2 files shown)


OsirisTerje:
Question 1:  Imho, I think this is ok.  The support is ending in a year, and I see no reason to stay on this.
Question 2: Can't see any
Question 3: As you indicate, we are testing the adapter, and not any external stuff.  Agree.

Issue #487, opened 2018-03-16
[Explicit] tests are executed in Rider/R#


roji:
When running the tests for my entire project in Rider, tests marked with `[Explicit]` are executed. [This issue was investigated by Jetbrains](https://youtrack.jetbrains.com/issue/RIDER-11989), who concluded that it's an issue on the NUnit side, where there adapter is hard-coded to run explicit ones. The issue also occurs on Visual Studio 15.6.2 with Resharper 2017.3.3.

NUnit: 3.10.1
NUnit3TestAdapter: 3.10.0
Rider: 2017.3.1 (on Linux)
Targets: net45, netcoreapp2.0

Test project csproj:

```xml
<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
  <PropertyGroup>
    <TargetFrameworks>net451;netcoreapp2.0</TargetFrameworks>
    <DebugType>portable</DebugType>
    <AssemblyOriginatorKeyFile>../../Npgsql.snk</AssemblyOriginatorKeyFile>
    <SignAssembly>true</SignAssembly>
    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="../../src/Npgsql/Npgsql.csproj" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="NUnit" Version="3.10.1" />
    <PackageReference Include="NLog" Version="5.0.0-beta06" />
    <PackageReference Include="Microsoft.CSharp" Version="4.4.1" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
    <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
  </ItemGroup>
  <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
    <Reference Include="System.Transactions" />
  </ItemGroup>
  <ItemGroup>
    <Reference Include="Microsoft.CSharp" />
  </ItemGroup>
</Project>
```

roji:
Note: in the old runner there was https://github.com/nunit/dotnet-test-nunit/issues/25, which led to https://github.com/nunit/nunit3-vs-adapter/issues/47, but that issue is marked as resolved in 3.10, and that doesn't seem to be the case.

OsirisTerje:
[...]

PR #509, opened 2018-05-23
Productize NUnit.Explicit

jnm2:
If I understand correctly, @estrizhok indicated that it would be useful to be able to rely on the continued presence of the NUnit.Explicit property so that the Rider UI could natively respect the concept of explicit test selection.
@OsirisTerje, it looks like you were in favor of this approach. If we encourage this route, can we add these tests to ensure that we don't accidentally change the property name, type or behavior, and forget that it might affect external runners?

Should we hold off on merging this until https://github.com/nunit/nunit3-vs-adapter/issues/487 is settled?


OsirisTerje:
@jnm2 You merge?

jnm2:
@OsirisTerje Sure, should we first wait to hear whether JetBrains decides to depend on this property, or are we sure this is something we want to encourage regardless?

OsirisTerje:
I think I saw it was fixed already, so let us just add it in.

jnm2:
Awesome! 😃

PR #510, opened 2018-05-24
Add a CollectDataForEachTestSeparately setting

### LINKED ISSUE #426: Exception thrown while loading tests if In-Proc VSTest DataCollector is used
Hello, 

When a custom VSTest In-Proc DataCollector is used, NUnit Test Adapter cannot find nunit.framework.

Part of output:

> NUnit Adapter 3.9.0.0: Test execution started
> Running all tests in C:\Users\pkruk\Documents\Visual Studio 2017\Projects\VSTestIntegration\SampleLibrary.NUnitTests\bin\Debug\SampleLibrary.NUnitTests.dll
> Exception NUnit.Engine.NUnitEngineException, Exception thrown executing tests in C:\Users\pkruk\Documents\Visual Studio 2017\Projects\VSTestIntegration\SampleLibrary.NUnitTests\bin\Debug\SampleLibrary.NUnitTests.dll
> An exception occurred in the driver while loading tests.
>    at NUnit.Engine.Runners.DirectTestRunner.LoadDriver(IFrameworkDriver driver, String testFile, TestPackage subPackage)
> [...]
> Innerexception: System.IO.FileNotFoundException: Could not load file or assembly 'nunit.framework' or one of its dependencies. The system cannot find the file specified.
> File name: 'nunit.framework'

Attaching:
[nunit.output.txt](https://github.com/nunit/nunit3-vs-adapter/files/1547986/nunit.output.txt)
[InProcTestDataCollector.zip](https://github.com/nunit/nunit3-vs-adapter/files/1547987/InProcTestDataCollector.zip)

This happens with NUnit3TestAdapter.3.9.0
I also used vstest built locally from recent master but also vstest from VS 2017 (15.3.0-preview-20170628-02).

At first I thought this might be a problem with VSTest itself, but I have tested xUnit and MSTest and they both work correctly with In-Proc data collectors. Also, I have quickly checked there is some special code in NUnit Test Adapter which checks for presence of InProcDataCollector in run settings, and in that case behaves differently than normally (while I could not find any similar code in xUnit) which hints at issue in NUnit Test Adapter code.

drognanar:
Add a setting to make NUnit run tests sequentially, making the InProcDataCollectors receive the events before a test starts and after a test ends. This allows adapters to opt into the sequential mode.

It adds a fallback for the old LiveUnitTesting adapter, so that when present the tests are running in a sequential mode, so that the new NUnit adapter keeps working with older releases of VS.

Fixes #426

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -248,9 +248,14 @@ namespace NUnit.VisualStudio.TestAdapter
             var inProcDataCollectorNode = doc.SelectSingleNode("RunSettings/InProcDataCollectionRunSettings/InProcDataCollectors");
             InProcDataCollectorsAvailable = inProcDataCollectorNode != null && inProcDataCollectorNode.SelectNodes("InProcDataCollector").Count > 0;
 
+            // Older versions of VS do not pass the CollectDataForEachTestSeparately configuration together with the LiveUnitTesting collector.
+            // However, the adapter is expected to run in CollectDataForEachTestSeparately mode.
+            // As a result for backwards compatibility reasons enable CollectDataForEachTestSeparately mode whenever LiveUnitTesting collector is being used.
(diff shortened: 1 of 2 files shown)


OsirisTerje:
Lgtm

PR #522, opened 2018-06-22
WIP   Issue497 filter

OsirisTerje:
WIP   *** DO NOT MERGE ***
Changed test case filter to use NUnit Filter.
Added a featureflag for testing purposes, have now left it default true, but keep open the possibility of defaulting to false if further tests show it is not stable.  Local tests so far looks good.


CharliePoole:
@OsirisTerje Re our email issues... I did get an email that you asked for my review of this.

OsirisTerje:
Ok, then that part works at least, but if you didnt get any emails from the mentions today, something is wrong with github.

src/NUnitTestAdapter/NUnitTestFilterBuilder.cs
```diff
@@ -1,44 +1,52 @@
...
-            ITestFilterBuilder filterBuilder = _filterService.GetTestFilterBuilder();
-
-            foreach (TestCase testCase in testCases)
-                filterBuilder.AddTest(testCase.FullyQualifiedName);
-
+            var filterBuilder = _filterService.GetTestFilterBuilder();
+            if (tfsFilter!=null && settings.UseTestCaseFilterConverter)
+            {
+                var nunitfilter = new VsTest2NUnitFilterConverter(tfsFilter.TfsTestCaseFilterExpression.TestCaseFilterValue);
+                filterBuilder.SelectWhere(nunitfilter.ToString());
```
CharliePoole: Since there is no error detection, It seems like there is an underlying assumption here that the converter will always produce a valid where clause. Do we know that's true for all possible values of the original filter?

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -67,6 +67,18 @@ public interface IAdapterSettings
         void RestoreRandomSeed(string dirname);
     }
 
+
+    public interface IFeatureFlags
+    {
+        bool UseTestCaseFilterConverter { get; }
+    }
+
+    public class FeatureFlags : IFeatureFlags
```
jnm2: Is this alternate implementation of IFeatureFlags public by design, or can it be private and maybe even a struct?
I have to think about why both AdapterSettings and FeatureFlags implement IFeatureFlags and where in the codebase FeatureFlags might be used versus AdapterSettings.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-// #define LAUNCHDEBUGGER
+ #define LAUNCHDEBUGGER
```
jnm2: Heh sorry 😁

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -136,7 +130,7 @@ public void RunTests(IEnumerable<TestCase> tests, IRunContext runContext, IFrame
                     var assemblyPath = Path.IsPathRooted(assemblyName) ? assemblyName : Path.Combine(Directory.GetCurrentDirectory(), assemblyName);
 
                     var filterBuilder = CreateTestFilterBuilder();
-                    var filter = filterBuilder.MakeTestFilter(assemblyGroup);
+                    var filter = filterBuilder.MakeTestFilter(assemblyGroup,new TfsTestFilter(runContext));
```
jnm2: Crazy nitpick: space after comma?

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -327,9 +335,9 @@ private void RunAssembly(string assemblyPath, TestFilter filter)
         private NUnitTestFilterBuilder CreateTestFilterBuilder()
         {
 #if NETCOREAPP1_0
-            return new NUnitTestFilterBuilder(new TestFilterService());
+            return new NUnitTestFilterBuilder(new TestFilterService(),Settings,TestLog);
 #else
-            return new NUnitTestFilterBuilder(TestEngine.Services.GetService<ITestFilterService>());
+            return new NUnitTestFilterBuilder(TestEngine.Services.GetService<ITestFilterService>(),Settings,TestLog);
```
jnm2: Spaces after commas?

src/NUnitTestAdapter/NUnitTestFilterBuilder.cs
```diff
@@ -1,44 +1,52 @@
...
+            _filterService = filterService ?? throw new NUnitEngineException("TestFilterService is not available. Engine in use is incorrect version.");
         }
 
         public TestFilter ConvertTfsFilterToNUnitFilter(TfsTestFilter tfsFilter, List<TestCase> loadedTestCases)
         {
             var filteredTestCases = tfsFilter.CheckFilter(loadedTestCases);
             var testCases = filteredTestCases as TestCase[] ?? filteredTestCases.ToArray();
             //TestLog.Info(string.Format("TFS Filter detected: LoadedTestCases {0}, Filterered Test Cases {1}", loadedTestCases.Count, testCases.Count()));
-            return MakeTestFilter(testCases);
+            return MakeTestFilter(testCases,tfsFilter);
```
jnm2: Space after comma?

src/NUnitTestAdapter/NUnitTestFilterBuilder.cs
```diff
@@ -1,44 +1,52 @@
...
         }
 
-        public TestFilter MakeTestFilter(IEnumerable<TestCase> testCases)
+        public TestFilter MakeTestFilter(IEnumerable<TestCase> testCases, TfsTestFilter tfsFilter)
```
jnm2: Order of parameters is the reverse of `ConvertTfsFilterToNUnitFilter`

src/NUnitTestAdapter/NUnitTestFilterBuilder.cs
```diff
@@ -1,44 +1,52 @@
...
-
-            foreach (TestCase testCase in testCases)
-                filterBuilder.AddTest(testCase.FullyQualifiedName);
-
+            var filterBuilder = _filterService.GetTestFilterBuilder();
+            if (tfsFilter!=null && settings.UseTestCaseFilterConverter)
+            {
+                var nunitfilter = new VsTest2NUnitFilterConverter(tfsFilter.TfsTestCaseFilterExpression.TestCaseFilterValue);
+                filterBuilder.SelectWhere(nunitfilter.ToString());
+                if (settings?.Verbosity>4)
```
jnm2: What do you think of putting spaces around the `<` operator and the `!=` operator above?
Apologies for the nitpickiness today!

Is `Verbosity > 4` different from skipping the `if` and using `logger.Debug`?

src/NUnitTestAdapter/TfsTestFilter.cs
```diff
@@ -90,19 +96,10 @@ public TfsTestFilter(IRunContext runContext)
...
-            }
-        }
+        public ITestCaseFilterExpression TfsTestCaseFilterExpression => testCaseFilterExpression ??
+                                                                        (testCaseFilterExpression = runContext.GetTestCaseFilter(SupportedProperties, PropertyProvider));
 
-        public bool IsEmpty
-        {
-            get { return TfsTestCaseFilterExpression == null || TfsTestCaseFilterExpression.TestCaseFilterValue == string.Empty; }
-        }
+        public bool IsEmpty => TfsTestCaseFilterExpression == null || TfsTestCaseFilterExpression.TestCaseFilterValue == string.Empty;
```
jnm2: Some of these lines get pretty long...

One option for long getters would also be
```cs
{
    get => TfsTestCaseFilterExpression == null || TfsTestCaseFilterExpression.TestCaseFilterValue == string.Empty;
}
```

src/NUnitTestAdapter/VSTest2NUnitFilterConverter.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NUnit.VisualStudio.TestAdapter
+{
+    public class VsTest2NUnitFilterConverter
+    {
+        private string NUnitFilter { get; }
+        public VsTest2NUnitFilterConverter(string vstestfilter)
+        {
+            NUnitFilter = vstestfilter.Replace("TestCategory", "Category");
+        }
```
jnm2: This might just be personal preference but doing the heavy lifting of the class doesn't feel very "constructory" to me.
The name is `Converter` but once you have an instance, you can't convert things with it. You can only access a conversion result.
What would you think of introducing a static `string (string)` method?

src/NUnitTestAdapterTests/VSTest2NUnitFilterConverterTests.cs
```diff
@@ -0,0 +1,48 @@
...
+// ***********************************************************************
+
+using NUnit.Framework;
+
+namespace NUnit.VisualStudio.TestAdapter.Tests
+{
+    public class VsTest2NUnitFilterConverterTests
+    {
+        [TestCase("TestCategory!=Whatever", "Category!=Whatever")]
+        [TestCase("TestCategory=Whatever","Category=Whatever")]
```
jnm2: Can you add `FullyQualifiedName = SomeNamespace.TestCategory.SomeMethod`?
What else should we add from https://github.com/Microsoft/vstest-docs/blob/master/docs/filter.md?

build.cake
```diff
@@ -12,7 +12,7 @@ var configuration = Argument("configuration", "Release");
 //////////////////////////////////////////////////////////////////////
 
 var version = "3.11.0";
-var modifier = "";
+var modifier = "-d03";
```
jnm2: Is there going to be a prerelease publish?

CharliePoole:
It's a clever idea to use text conversion rather than parsing, changing the text representation of the VS filter to the text representation of an NUnit filter.

It should work if the languages are sufficiently close. That's my area of doubt. The alternative, of course, is to create a grammar of VS filters - maybe Microsoft has that already - and parse it, generating NUnit filters directly without the intermediate step of TSL.

My general feeling is that if you want to go this way you need to throw a lot more tests at it, including some invalid input and non-translatable input if you can come up with examples. Currently, the code assumes no error will ever occur.

build.cake
OsirisTerje: Just slipped through, I update these locally when debugging.  Will remove again.

CharliePoole:
Of course, 3 was what I originally proposed to you.

The impedance mismatch will always exist but some approaches make it easier than others to give a precise error message.

What happens if you use a non-supported rhs element? What happens if you use invalid syntax, say a bad operator or unbalanced parentheses?

I'll try to make up some test cases later at my computer.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
OsirisTerje: Oops, fixed

src/NUnitTestAdapter/AdapterSettings.cs
OsirisTerje: This FeatureFlags class is used for the runsettings, so that we can separate out this block of settings.  Since the other ones are options, so to speak, I find the featureflags different, and therefore easier to have them in a separate block.  In the code though, it is enough that we can see them in the adaptersettings, therefore the interface is implemented there too, and implementation just passed down.

src/NUnitTestAdapter/VSTest2NUnitFilterConverter.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class VsTest2NUnitFilterConverter
+    {
+        private string NUnitFilter { get; }
+        public VsTest2NUnitFilterConverter(string vstestfilter)
+        {
+            NUnitFilter = vstestfilter
+                .Replace("TestCategory", "Category")
+                .Replace("FullyQualifiedName","test")
+                .Replace("~","=~");
```
Dreamescaper: VS Test filter also support `!~` operator (was added here https://github.com/microsoft/vstest/pull/1803).
Should it be accounted here?

Issue #516, opened 2018-06-13
ArgumentException when whitespace sent to logger


slide:
This is related to https://github.com/nunit/nunit3-vs-adapter/issues/220. The previous fix was to send a single space when there was no text in the message, but TestSessionMessageLogger checks for IsNullOrWhitespace and throws an ArgumentException if the message is null or only whitespace, (https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.Common/Logging/TestSessionMessageLogger.cs#L65) which means sending the single space causes an argument exception.

These are the versions we are referencing:

```xml
<PackageReference Include="NUnit.ConsoleRunner" Version="3.7.0" />
<PackageReference Include="NUnitLite" Version="3.9.0" />
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
<PackageReference Include="NUnit3TestAdapter" Version="3.9.0" />
<PackageReference Include="Mono.Cecil" Version="0.10.0-beta6" />
```

We use nunit to run unit tests for IronPython. Cloning https://github.com/IronLanguages/ironpython2 and running `make.ps1` then `make.ps1 test-test_macurl2path` will show the issue. I worked around this in a local build of the adapter by implementing my own IsNullOrWhiteSpace (since the adapter is being built with a framework prior to IsNullOrWhiteSpace) and didn't send the message if it is null or only whitespace. 

This is the output we see from tests:

```
.\make.ps1 test-test_macurl2path
WARNING: No tests available for 'test_macurl2path' trying to run single test 'net45.test_macurl2path'
dotnet test C:\Users\USER\Code\_ipy\ironpython2/Src/IronPythonTest/IronPythonTest.csproj -f net45 -o C:\Users\USER\Code\_ipy\ironpython2/bin/Release/net45 -c Release --no-build -l trx;LogFileName=single-net45-Release-result.trx -s C:\Users\USER\Code\_ipy\ironpython2\Src\IronPythonTest\runsettings.xml --filter="Name=net45.test_macurl2path"
Microsoft (R) Build Engine version 15.7.177.53362 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 160.02 ms for C:\Users\USER\Code\_ipy\ironpython2\Src\DLR\Src\Microsoft.Scripting\Microsoft.Scripting.csproj.
[...]

PR #523, opened 2018-06-23
Eventlistener will no longer send single spaces when the test output …

OsirisTerje:
…is empty or just whitespace  #516 

The eventlistener will not send anything if the test output is null, empty or just consists of whitespaces


src/NUnitTestAdapter/Internal/Extensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace NUnit.VisualStudio.TestAdapter.Internal
+{
+#if NET35
+
```
jnm2: Nitpick: empty line between the `#if` and the class, but no empty line between the `#endif` and either class.

src/NUnitTestAdapter/NUnitEventListener.cs
```diff
@@ -156,16 +155,15 @@ public void TestFinished(XmlNode resultNode)
         public void SuiteFinished(XmlNode resultNode)
         {
             var result = resultNode.GetAttribute("result");
-            var label = resultNode.GetAttribute("label");
             var site = resultNode.GetAttribute("site");
-
+
```
jnm2: Super silly nitpick: new whitespace added to this line.

src/NUnitTestAdapter/NUnitEventListener.cs
```diff
@@ -180,22 +178,21 @@ public void SuiteFinished(XmlNode resultNode)
 
         private static readonly string NL = Environment.NewLine;
         private static readonly int NL_LENGTH = NL.Length;
-        private IDumpXml dumpXml;
+        private readonly IDumpXml dumpXml;
```
jnm2: Nice to see the cleanup in this file.

src/NUnitTestAdapter/Internal/Extensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class TypeExtensions
+    {
+        public static Type GetTypeInfo(this Type type) => type;
+    }
+#endif
+    public static class StringExtensions
+    {
+        public static bool IsNullOrWhiteSpace(this string value)
+        {
+            return value == null || value.Trim().Length == 0;
```
jnm2: It's good to stay simple since it's not in a hot path. 👍

For comparison: https://referencesource.microsoft.com/#mscorlib/system/string.cs,55e241b6143365ef

src/NUnitTestAdapterTests/ExtensionsTests.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace NUnit.VisualStudio.TestAdapter.Tests
+{
+#if !NETCOREAPP1_0
+    public class ExtensionsTests
+    {
+        [TestCase("\t\t\t")]
+        [TestCase("    \t")]
+        [TestCase("  ")]
+        [TestCase("")]
+        public void ThatIsNullOrWhiteSpaceHandlesTabs(string value)
```
jnm2: Not all of these cases are related to handling tabs. Also can we have `[TestCase(null)]` and `[TestCase("\r\n")]`?

src/NUnitTestAdapter/Internal/Extensions.cs
OsirisTerje: Yes, just used the https://msdn.microsoft.com/en-us/library/system.string.isnullorwhitespace(v=vs.110).aspx as a reference, and under remarks this code is shown as equivalent, and it is easier than the code from the referencesource.

src/NUnitTestAdapterTests/ExtensionsTests.cs
OsirisTerje: Done

src/NUnitTestAdapter/NUnitEventListener.cs
OsirisTerje: :+1:

src/NUnitTestAdapter/NUnitEventListener.cs
OsirisTerje: Happens when a new line is inserted and spaces are default :-(

src/NUnitTestAdapter/Internal/Extensions.cs
OsirisTerje: removed empty line

src/NUnitTestAdapterTests/ExtensionsTests.cs
```diff
@@ -0,0 +1,54 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using NUnit.Framework;
+using NUnit.VisualStudio.TestAdapter.Internal;
+
+namespace NUnit.VisualStudio.TestAdapter.Tests
+{
+#if !NETCOREAPP1_0
```
jnm2: Shouldn't we be running these tests on netcoreapp1.0?

OsirisTerje: Done

src/NUnitTestAdapterTests/NUnitEventListenerTests.cs
```diff
@@ -191,4 +195,46 @@ private void VerifyTestResult(VSTestResult ourResult)
...
+    {
+        private ITestExecutionRecorder recorder;
+        private ITestConverter converter;
+        private IDumpXml dumpxml;
+
+        private const string Testoutput =
+            @"<test-output stream='Progress' testid='0-1001' testname='Something.TestClass.Whatever'><![CDATA[Whatever
+]]></test-output>";
+
+        private const string BlankTestoutput =
```
jnm2: For both these fields, I think you'd capitalize the O since it's two words.

OsirisTerje:
Seems there is some build issues with the test project running NSubstitute......  got to look at that before merging.

src/NUnitTestAdapterTests/NUnit.TestAdapter.Tests.csproj
```diff
@@ -4,7 +4,8 @@
   <PropertyGroup>
     <RootNamespace>NUnit.VisualStudio.TestAdapter.Tests</RootNamespace>
     <AssemblyName>NUnit.VisualStudio.TestAdapter.Tests</AssemblyName>
-    <TargetFrameworks>net46;netcoreapp1.0</TargetFrameworks>
+    <TargetFrameworks>net35;netcoreapp1.0</TargetFrameworks>
```
jnm2: @OsirisTerje Changing net46 to net35 is what is causing the NuGet restore failures. Any reason not to use `net46`?

OsirisTerje: copy paste error when I reverted back from debugging this.  It is very annoying that VS don't really work with multiple targets, so I always remove the .netcore one.......,  just my way.....

Issue #494, opened 2018-03-26
TestContext.AddTestAttachment does not work on Linux environment with specified dotnet logger


Kira-Lappo:
# Summary

Test Attachments are not added to trx file after test run is done on linux environment.

# Repro steps
1. Create Nunit tests project with the only test, which adds some existing file to attachments.
1. Build project and start test execution with command:
    ```dotnet vstest "/path/to/project.dll" /logger:"trx" /ResultsDirectory:./test-results /Diag:./diag/diag.main.log ```
1. Wait until test execution is finished. Take a look at terminal messages, it says that 1 of 1 tests passed.
1. Check created trx file

**Expected result** : the trx file contains section with a test attachment with correct file path;
**Actual result** : the trx file contains stderr section for the executed test:
```
 <StdErr>Failed to attach files from: Attachments&#xD;
Error Details: System.InvalidOperationException:System.InvalidOperationException: This operation is not supported for a relative URI.
   at System.Uri.get_LocalPath()
   at Microsoft.TestPlatform.Extensions.TrxLogger.Utility.Converter.ToResultFiles(AttachmentSet attachmentSet, Guid testResultExecutionId, TestRun testRun, String trxFileDirectory)
   at Microsoft.TestPlatform.Extensions.TrxLogger.Utility.Converter.UpdateTestResultAttachments(TestResult rockSteadyTestResult, TestResult testResult, TestRun testRun, String trxFileDirectory, Boolean addAttachments)</StdErr>
```

# Analysis
I digged into adapter code, checked mstest logic and test execution logs(./diag/diag.main.log).
The reason of the problem is that adapter sends straigh file path to mstest's executor, without **file://** URI schema. 
So, for example, path ```/absolute/path/to/attachment.txt``` is a correct absolute path on linux environment, but mstest consider it as a relative path when it deserializes the data which was recieved from adapter. The path is deserialized without errors, but [calling property **.LocalPath**](https://github.com/Microsoft/vstest/blob/4128f5dc8243221413af2f4dcf68dbcc75cc75dd/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs#L510)(line 510) on relative uri causes exception from the repro steps.

[...]

PR #527, opened 2018-07-06
 TestConverter: added check for attachment path - if it starts with file:// scheme

Kira-Lappo:
The PR fixes the problem from [issue 494](https://github.com/nunit/nunit3-vs-adapter/issues/494).

I added constant **fileUriScheme** into helper method ```AttachmentSet ParseAttachments(XmlNode resultNode)```.
Then I added check for each path in attachments. 
When path is not empty or not already started with **fileUriScheme**, **fileUriScheme** value is added at the beginning of the original path.


OsirisTerje:
LGTM.   
Would be nice to see a unit test covering this though.

Kira-Lappo:
@OsirisTerje I've added two unit tests which check that correct amount of test attachments was converted and that all pathes start with the only "file://" value. I guess it's enaugh for that part.

OsirisTerje:
Superb!  Thanks!

Issue #488, opened 2018-03-21
The use of Cecil 0.10.0-beta6 clashes with systems under test using the Cecil 0.10.0 release


SteveGilham:
If I have a system under test using the new Mono.Cecil 0.10.0 final release, then the presence of the beta version can lead to MethodMissingException being thrown by the system under test.  This was originally reported as a [Cecil issue #500](https://github.com/jbevain/cecil/issues/500), but proved eventually to be due to the presence of the test adapter in the test dependencies.

* NUnit and NUnit3TestAdapter versions 
```
PackageReference Include="NUnit" Version="3.10.1" />
PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
```
* Visual Studio edition and full version number (see Help About)
   Community 15.6.3

* A short repro, preferably attached or pointing to a git repo or gist
   https://gist.github.com/SteveGilham/c86315da2f003cced56d23c128f7c363

* What .net platform and version is being targeted
  .net core 2.0

* If TFS/VSTS issue, what version, hosted, on-premises, and what build task you see this in
  n/a

jnm2:
Thank you for filing! This looks like the same issue as https://github.com/nunit/nunit3-vs-adapter/issues/366. Is that right?

SteveGilham:
Yes -- another manifestation of assembly mismatch when the assembly versions are the same, but the API isn't.  I would be hopeful that this one could be resolved (at least until the next Cecil release cycle starts) by simply packaging the newly released 0.10.0 which is then as new as everything if not newer.

jnm2:
@rprouse and @OsirisTerje, assuming short-term it's okay to force everyone to 0.10.0 instead of forcing everyone to 0.10.0-beta, would it make sense to release the adapter with 0.10.0 bundled inside even though the engine was not compiled against it? Or should the 0.10.0 dependency be done in the engine first?

rprouse:
I would prefer to update it in the engine first to be safe. It could be a point release for just the .NET Standard engine. It is still going to be a problem for anyone not on .NET Core though. The full framework version of the engine targets .NET 2.0, but Mono.Cecil dropped support for 2.0 so we can't update. The only thing that would work there is to update the engine and everything else in that project to .NET 3.5 which would be contentious.

jnm2:
We could use a conditional PackageReference element to pull a different version for the net20 target.

[...]

PR #541, opened 2018-08-26
 Embed Mono.Cecil so user dependencies are no longer overwritten 

jnm2:
Closes nunit/nunit-console#389, #366, and #488 for all platforms! Pretty excited about this after all our frustrations! 🎉

The first commit adds a simple passing acceptance test that proves the adapter is packaged basically correctly. The acceptance tests hardly add any time on my machine after a `git clean -xfd`, despite having to cache all NuGet packages from scratch.

The third commit adds a failing test asserting that the test project's desired version of Mono.Cecil has been loaded, which is the problem discussed in nunit/nunit-console#389, #366, and #488.

The fourth commit fixes the problem by embedding Mono.Cecil as an adapter assembly resource and resolving it on demand.

### DIFF
--- a/lib.cake
+++ b/lib.cake
@@ -125,7 +125,8 @@ private sealed class ProjectFixture
             Framework = targetFramework,
             NoBuild = true,
             Logger = "trx;LogFileName=" + resultsFile.GetFilename(),
-            ResultsDirectory = resultsFile.GetDirectory()
+            ResultsDirectory = resultsFile.GetDirectory(),
+            Settings = _context.File("DisableAppDomain.runsettings")
         });
 
         return resultsFile;
--- a/lib.cake
+++ b/lib.cake
@@ -1,0 +1,133 @@
+public static T WithRawArgument<T>(this T settings, string rawArgument) where T : Cake.Core.Tooling.ToolSettings
+{
+    if (settings == null) throw new ArgumentNullException(nameof(settings));
+
+    if (!string.IsNullOrEmpty(rawArgument))
+    {
+        var previousCustomizer = settings.ArgumentCustomization;
+        if (previousCustomizer != null)
+            settings.ArgumentCustomization = builder => previousCustomizer.Invoke(builder).Append(rawArgument);
+        else
+            settings.ArgumentCustomization = builder => builder.Append(rawArgument);
+    }
+
+    return settings;
+}
+
+void DeleteDirectoryRobust(params string[] directories)
+{
+    DeleteDirectoryRobust(Context, directories);
+}
+
+static void DeleteDirectoryRobust(this ICakeContext context, params DirectoryPath[] directories)
+{
+    DeleteDirectoryRobust(context, Array.ConvertAll(directories, d => d.FullPath));
+}
+
+static void DeleteDirectoryRobust(this ICakeContext context, params string[] directories)
+{
+    if (!directories.Any()) return;
+
+    context.Information("Deleting directories:");
+
+    foreach (var directory in directories)
+    {
+        for (var attempt = 1;; attempt++)
+        {
+            context.Information(directory);
+            try
+            {
+                System.IO.Directory.Delete(directory, recursive: true);
+                break;
+            }
+            catch (DirectoryNotFoundException)
+            {
+                break;
+            }
+            catch (IOException ex) when (attempt < 3 && (WinErrorCode)ex.HResult == WinErrorCode.DirNotEmpty)
+            {
+                context.Information("Another process added files to the directory while its contents were being deleted. Retrying...");
+            }
+        }
+    }
+}
+
+private enum WinErrorCode : ushort
+{
+    DirNotEmpty = 145
+}
+
+public sealed class TempDirectory : IDisposable
+{
+    public DirectoryPath Path { get; }
+
+    public TempDirectory()
+    {
+        Path = new DirectoryPath(System.IO.Path.GetTempPath()).Combine(System.IO.Path.GetRandomFileName());
+        System.IO.Directory.CreateDirectory(Path.FullPath);
+    }
+
+    public void Dispose()
+    {
+        System.IO.Directory.Delete(Path.FullPath, recursive: true);
+    }
+
+    public static implicit operator DirectoryPath(TempDirectory tempDirectory)
+    {
+        return tempDirectory.Path;
+    }
+}
+
+ProjectFixture NewProjectFixture(DirectoryPath projectDirectory, string configuration, DirectoryPath testResultsDirectory)
+{
+    return new ProjectFixture(Context, projectDirectory, configuration, testResultsDirectory);
+}
+
+private sealed class ProjectFixture
+{
+    private readonly ICakeContext _context;
+    private readonly DirectoryPath _projectDirectory;
+    private readonly string _configuration;
+    private readonly DirectoryPath _testResultsDirectory;
+
+    private string ProjectName => _projectDirectory.GetDirectoryName();
+
+    public ProjectFixture(ICakeContext context, DirectoryPath projectDirectory, string configuration, DirectoryPath testResultsDirectory)
+    {
+        _context = context;
+        _projectDirectory = projectDirectory;
+        _configuration = configuration;
+        _testResultsDirectory = testResultsDirectory;
+    }
+
+    public void Build(string adapterPackageVersion)
+    {
+        _context.DeleteDirectoryRobust(_projectDirectory.Combine("bin"));
+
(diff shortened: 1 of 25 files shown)


jnm2:
Build failure because on my local machine, `.\build -t PackageNuGet` produces a package with version `3.11.0`, but on AppVeyor it is `3.11.0-ci-00898-pr-541` and changes each time. I'll have the build patch the acceptance test csproj PackageReference elements with the calculated version.

lib.cake
```diff
@@ -0,0 +1,133 @@
...
+        _configuration = configuration;
+        _testResultsDirectory = testResultsDirectory;
+    }
+
+    public void Build(string adapterPackageVersion)
+    {
+        _context.DeleteDirectoryRobust(_projectDirectory.Combine("bin"));
+
+        var projectFile = _projectDirectory.CombineWithFilePath(ProjectName + ".csproj");
+        _context.XmlPoke(projectFile, "/Project/ItemGroup/PackageReference[@Include = 'NUnit3TestAdapter']/@Version", adapterPackageVersion);
```
jnm2: Had to add this line since the CI version changes with every build.

Ready for review!

/cc @OsirisTerje

bording:
@jmn2 I haven't looked over the full details of this PR yet, but I see how you're preventing the test project from getting the wrong version by not overwriting the version in the bin folder, but how exactly does this ensure that the adapter doesn't get the version referenced by the project?

jnm2:
@bording @OsirisTerje My thought was that nunit.engine uses so little of Mono.Cecil that it should work with whatever version the user wants. The reason the user's version always takes precedence is that it's the version that's in the same folder as nunit.engine, so `AssemblyResolve`/`Resolving` doesn't even happen.
I'm honestly not worried at all about breaking changes in Mono.Cecil, but this PR is at least an increment in the right direction.

The .NET Framework version of the acceptance test loads `Mono.Cecil 0.10.0.0-beta5` when the engine tries to load its referenced `0.9.6.1` (has .NET assembly version 0.9.6.0), so I know that switching versions  isn't an issue on .NET Framework with this technique even though nunit.engine is strongly named. And on .NET Core, there is no assembly binding policy that requires strong-named assemblies to load specific versions of their dependencies.

.NET Core doesn't have AppDomains, but it does have AssemblyLoadContexts. In separate experiments I found that it's possible to load two versions of the same DLL simultaneously by setting up a derived AssemblyLoadContext but that would mean architecturally separating the ITestDiscoverer/ITestExecutor entry point (sets up the AssemblytLoadContext) from any code which depends on nunit.engine (must run inside the derived AssemblyLoadContext).

jnm2:
@bording @OsirisTerje My thought was that nunit.engine uses so little of Mono.Cecil that it should work with whatever version the user wants. The reason the user's version always takes precedence is that it's the version that's in the same folder as nunit.engine, so `AssemblyResolve`/`Resolving` doesn't even happen.
I'm honestly not worried at all about breaking changes in Mono.Cecil, but this PR is at least an increment in the right direction.

The .NET Framework version of the acceptance test loads `Mono.Cecil 0.10.0.0-beta5` (assembly version 10.0.0.0) when the engine tries to load its referenced `0.9.6.1` (has .NET assembly version 0.9.6.0), so I know that switching versions  isn't an issue on .NET Framework with this technique even though nunit.engine is strongly named. And on .NET Core, there is no assembly binding policy that requires strong-named assemblies to load specific versions of their dependencies.

If we do have a need for the engine to load a version that's not in the same folder, it's more complex but possible. .NET Core doesn't have AppDomains, but it does have AssemblyLoadContexts. In separate experiments I found that it's possible to load two versions of the same DLL simultaneously (one in the same folder, one not) by setting up a derived AssemblyLoadContext. That would mean architecturally separating the ITestDiscoverer/ITestExecutor entry point (sets up the AssemblyLoadContext) from any code which depends on nunit.engine (must run inside the derived AssemblyLoadContext).

bording:
@jnm2 There are definitely breaking changes between 0.9.6.0 and 0.10.0, so if I've got a .NET Framework test project referencing 0.10, I'd be worried that nunit.engine would break if it was forced to use 0.10.0 instead.

What version of Mono.Cecil is the .NET Standard version of nunit.engine using? Has it been updated to 0.10.0? There were also some breaks between beta versions and RTM.

bording:
> Would you be able to help me get a sense of the type of binary breaking change or behavioral breaking change that could happen from nunit.engine's use of Cecil?

@jnm2 Take a look at http://cecil.pe/ and search for "breaking change" to see the sort of things that changed over the course of the 0.10.0 development process.

> The acceptance tests run property when redirecting the version on both platforms to 0.10.0.0, so we don't exercise any code paths that cause breaks in that test.

Where does that happen? Seems like the test is referencing 0.10.0-beta5.

jnm2:
> Where does that happen? Seems like the test is referencing 0.10.0-beta5.

Oops, I forgot I redirected them to beta 5 in case the engine moves forward to RTW.

I'll take a closer look at the breaking changes when I get a chance. If I understand correctly, this PR is strictly an improvement over current scenarios.

bording:
> I'll take a closer look at the breaking changes when I get a chance. If I understand correctly, this PR is strictly an improvement over current scenarios.

As long the engine doesn't have a problem with using 0.10.0, then this should definitely allow me to move to the latest version of the adapter!

jnm2:
@rprouse Would it be possible for the engine to be running on net20 if the test project is targeting net35 (which is necessary in order for it to reference Mono.Cecil 0.10.0)?

jnm2:
@rprouse Would it be possible for the engine to be running on net20 if the test project is targeting net35 (which is necessary in order for it to reference Mono.Cecil 0.10.0)? I was thinking this would all be in the same process and therefore that this scenario couldn't happen.

I just realized that VSTest won't ever invoke the adapter on anything lower than net35, so maybe we're definitely safe from the BCL aspect? Still a great idea to run engine repo tests against Mono.Cecil 0.10.0 on net35.

jnm2:
So Mono.Cecil 0.10.0 has a breaking change that does effect the engine!

> CS0030 Cannot convert type 'Mono.Cecil.InterfaceImplementation' to 'Mono.Cecil.TypeReference'	

On this line:

https://github.com/jnm2/nunit-console/blob/e3558779c790a731776e2ea07ab27f45dc97143dO/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs#L271

So I have a little more work to do.

jnm2:
I created a new acceptance test that proves that the .NET Framework build is already loading both 0.9.6.0 and 0.10.0.0 and using each one appropriately (former for engine, latter for tests tests) in the same AppDomain due to strong naming.

It's also passing for .NET Core though in that case it's redirecting `0.10.0-beta6` to `0.10.0`.

@OsirisTerje and @bording, does all of this satisfy you? Or do you really want me to figure out how to solve the problem of loading both 0.10.0-beta6 (for engine) and 0.10.0 (for tests) side by side on .NET Core?

jnm2:
@bording 

> In the long run, it seems like you'd want to solve this, but at the moment this sounds like it should solve the problem I've been having.

For sure, but I have a lot of fires to put out right now and I haven't been able to figure out how to write a failing test to motivate further work. Once that shows up it'll be easier to get the work done. :D

bording:
@jnm2 Do you have an ETA for when an updated package that includes this change will be released? I'd love to see this ASAP to finally be able to upgrade and fix all the pain this has been causing! 😁

jnm2:
@bording I'll cc @OsirisTerje for the next release date. We would also be grateful if you have time to merge this with your nuget.config and update your adapter to `3.11.0-dev-00904` or newer so that you can let us know if we missed something in your scenario before the final release:

```xml
<configuration>
  <packageSources>
    <add key="NUnit prerelease" value="https://www.myget.org/F/nunit/api/v3/index.json" />
  </packageSources>
</configuration>
```

bording:
@jnm2 I just tried it out, and it appears to be working perfectly! Thanks for getting this fixed!

OsirisTerje:
@bording  I'm preparing a new release pretty soon now, we're delayed - but it is very soon.

bording:
@OsirisTerje Any update on this? I'd love to see the updated release come out so I can finally get this problem resolved in my repos.

OsirisTerje:
@bording Still delayed, waiting for some of the other stuff to get finished, but it takes time.  Considering releasing one now with what we have so far, and then another one a bit later when that stuff is finished.  Sorry for keeping you waiting, and thanks for the poke :-)

bording:
> Considering releasing one now with what we have so far, and then another one a bit later when that stuff is finished. 

That has my vote for sure! This is an important fix to get out, and you can always release again when the other stuff is ready.

PR #551, opened 2018-10-12
Update to the new .NET Standard version of the Engine

rprouse:
~~**Do not merge** - this requires the next release of the engine.~~ I made the changes to test the work we are doing in the engine to merge the .NET Standard engine back into the original engine. This uses a local compile/nuget package of that update. I figured that since the work was done I may as well save it for when we are ready.


nvborisenko:
Works to me. Verified it with `.netcoreapp2.0` tests, and `ExtensionService` works as well.
But `TestAdapter` should targets to `netcoreapp2.0` to use 'nunit.engine netstandard2.0' version.

rprouse:
I've updated this to use the upcoming 3.10 engine release (that is why the builds are failing, the engine this is using hasn't been released yet).

I've also merged latest from master and resolved merge conflicts. We should be able to merge this once I get the 3.10 release of the engine out.

nvborisenko:
@rprouse Rob, could you please take #551 PR into work, I guess this is related. That PR adds `netcoreapp2.0` as target framework.

rprouse:
I've done the 3.10 release of the engine so this can now be merged once CI passes. I will make a small change to kick off CI on all platforms.

nvborisenko:
@OsirisTerje we need you here :)

OsirisTerje:
@nvborisenko Thanks for the ping, and sorry to be a bit late here.

PR #555, opened 2018-10-21
Preparing for release of 3.11

OsirisTerje:
Hiding warnings for assemblies without tests which triggers also on non-test assemblies, so this reduces some noise
It now looks like this, and all these are non-test assemblies
image

### DIFF
--- a/nuget/NUnit3TestAdapter.nuspec
+++ b/nuget/NUnit3TestAdapter.nuspec
@@ -13,7 +13,7 @@
         <summary>NUnit 3 adapter for running tests in Visual Studio. Works with NUnit 3.x, use the NUnit 2 adapter for 2.x tests.</summary>
         <description>A package including the NUnit 3 TestAdapter for Visual Studio 2012 onwards.  With this package you don't need to install the VSIX adapter package, and you don't need to upload the adapter to your TFS server.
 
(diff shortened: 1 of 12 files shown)


nuget/NUnit3TestAdapter.nuspec
```diff
@@ -13,7 +13,7 @@
         <summary>NUnit 3 adapter for running tests in Visual Studio. Works with NUnit 3.x, use the NUnit 2 adapter for 2.x tests.</summary>
         <description>A package including the NUnit 3 TestAdapter for Visual Studio 2012 onwards.  With this package you don't need to install the VSIX adapter package, and you don't need to upload the adapter to your TFS server.
 
-Note that this package ONLY contains the adapter, not the NUnit framework.  You must also get the framework. You only need one such package for a solution.
+Note that this package ONLY contains the adapter, not the NUnit framework.  You must also get the framework. For VS 2017 and forward you should add this package to every test project in your solution, earlier version only require a single adapter package.
```
jnm2: Nits: improper use of comma, and `version` should probably be `versions`. Maybe:

```
For VS 2017 and forward, you should add this package to every test project in your solution. (Earlier versions only require a single adapter package per solution.)
```

PR #562, opened 2018-11-15
Update to NUnit.Engine 3.9.0, and netstandard engine 3.8.0

ChrisMaddock:
https://github.com/nunit/nunit-console/pull/479 will introduce some major changes to the engine, which we'll want to test with the adapter before merging.

I spotted the adapter is currently using a version of the engine a few behind the current release. To make the transition easier, I wanted to first update it to the latest released version, before trying to pull in the new changes. 

The update has required a few changes, which I'll line comment below. The acceptance tests are currently failing, but I've ran out of time for today and will need to get back to these. (@jnm2 - I understand these are something you added recently - if there's anything that jumps out to you that I should check out, I'd appreciate the tips!)

### DIFF
--- a/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
+++ b/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
@@ -23,6 +23,10 @@
     <Authors>Charlie Poole, Terje Sandstrom</Authors>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net35'">
+    <DisableHandlePackageFileConflicts>false</DisableHandlePackageFileConflicts>
+  </PropertyGroup>
+
   <ItemGroup>
     <Content Include="..\..\LICENSE.txt" Link="LICENSE.txt" />
   </ItemGroup>
--- a/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
+++ b/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
@@ -33,14 +33,14 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net35'">
     <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="11.0.0" PrivateAssets="All" />
-    <PackageReference Include="nunit.engine" Version="3.7.0" />
+    <PackageReference Include="nunit.engine" Version="3.9.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0'">
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="1.1.2" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="1.1.2" />
     <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="15.0.0" PrivateAssets="All" />
-    <PackageReference Include="nunit.engine.netstandard" Version="3.7.0" />
+    <PackageReference Include="nunit.engine.netstandard" Version="3.8.0" />
   </ItemGroup>
 
   <Target Name="PreventTestPlatformObjectModelCopyLocal" AfterTargets="ResolveReferences">
--- a/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
+++ b/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
@@ -24,6 +24,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net35'">
+    <!-- Suppress errors from https://github.com/dotnet/sdk/issues/1897-->
     <DisableHandlePackageFileConflicts>false</DisableHandlePackageFileConflicts>
   </PropertyGroup>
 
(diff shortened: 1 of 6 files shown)


src/NUnitTestAdapter/NUnit.TestAdapter.csproj
```diff
@@ -23,6 +23,11 @@
     <Authors>Charlie Poole, Terje Sandstrom</Authors>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net35'">
+    <!-- Suppress errors from https://github.com/dotnet/sdk/issues/1897-->
+    <DisableHandlePackageFileConflicts>false</DisableHandlePackageFileConflicts>
+  </PropertyGroup>
+
```
ChrisMaddock: I believe we're seeing https://github.com/dotnet/sdk/issues/1897 on the CI builds. This change suppresses the extraneous errors from the build logs.

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -123,14 +124,26 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
...
+                    if (e.InnerException is BadImageFormatException)
+                    {
+                        // we skip the native c++ binaries that we don't support.
+                        TestLog.Warning("Assembly not supported: " + sourceAssembly);
+                    }
+                    else
+                    {
+                        TestLog.Warning("Exception thrown discovering tests in " + sourceAssembly, e);
+                    }
+                }
```
ChrisMaddock: In the latest engine, an NUnitEngineException is thrown instead of a BadImageFormatException for native assemblies. That's seems like 'more correct' behaviour on the engine's part, so I've updated the adapter to match. Glad there was a unit test to catch this! 😄

OsirisTerje:
@ChrisMaddock Suggest you also update NUnit for the test projects at the same time :-)  With the upcoming changes to VS, the adapter should also be included and updated to the latest.

jnm2:
> There is something with the acceptance.cake script that @jnm2 added 2 months ago.

Oh hey, I missed that the problem was related to acceptance.cake. I'll pull this branch and play with it.

jnm2:
> Skipping running test for project C:\Users\Joseph\Source\Repos\nunit3-vs-adapter\tests\Simple\Simple.csproj. To run tests with dotnet test add "<IsTestProject>true<IsTestProject>" property to project file.

My machine has .NET Core SDK 2.1.500. Something changed since the last SDK I ran this against, probably?

jnm2:
And if you add `<IsTestProject>true<IsTestProject>` and fix the XML error, the next thing you get is this from vstest.console.exe 15.9:

> Framework35 is not supported. For projects targeting .Net Framework 3.5, please use Framework40 to run tests in CLR 4.0 "compatibility mode".

jnm2:
@ChrisMaddock Got everything working! https://github.com/nunit/nunit3-vs-adapter/compare/temp/engine390

@OsirisTerje If these commits get into master, does that solve all issues you were having? Or is there another branch I should look at?

OsirisTerje:
Seems it still fails?  @jnm2

OsirisTerje:
@jnm2 can you merge your changes in here?  I see it works on your branch

ChrisMaddock:
Thanks Joseph, that's awesome! Thanks for looking into it!

jnm2:
@ChrisMaddock (Let me know if you want help getting that fixed.)

Also, I found more info on the fixes I did here and we're putting them through as a PR just to get master working again: https://github.com/nunit/nunit3-vs-adapter/pull/571

They're different from the ones in this branch, so you'll want to drop the two commits named `VSTest 15.9 is requiring <IsTestProject>` and `VSTest 15.9 doesn’t support /Framework:Framework35`.

OsirisTerje:
Take a look at #573 .  Some people add NUnit.Engine as a separate package.  Can we add a check to ensure that we have the intended version and warn otherwise ?

jnm2:
Should we make give the engine the Mono.Cecil treatment by embedding it as a resource (https://github.com/nunit/nunit3-vs-adapter/pull/541)? It was something I had in mind when I did https://github.com/nunit/nunit3-vs-adapter/pull/541.

OsirisTerje:
@jnm2 @ChrisMaddock  Will one of you fix the conflicts ?

OsirisTerje:
@ChrisMaddock   You said:
"Not sure what you mean by this sentence, sorry?
       With the upcoming changes to VS, the adapter should also be included and updated to the latest."

With Update 15.9 and forward, Visual Studio  require the adapter to be present in the test binary output folders, where NUnit itself and the test dll will be.   Earlier the adapter was found by looking in the packages folder, and in some other known locations.  That seems to be deprecated (based on the latest RFCs), and right now there are options in VS that enables those.  These are currently the default, but at some time (and I assume pretty soon), the default will be the opposite.  
I have written about the upcoming changes here:  http://hermit.no/vsix-based-test-adapters-to-be-deprecated-in-visual-studio-for-c-and-visual-basic-to-speed-up-testing/

jnm2:
@ChrisMaddock Nope, no problems. I didn't realize you were aware. :+1:

PR #567, opened 2018-11-21
Updated version numbers

OsirisTerje:


OsirisTerje:
Thanks, wasn't aware of that.   Seems we need some release flow now.  The release 3.11.1 is already out

Issue #474, opened 2018-03-04
Skip surfacing certain properties as UI groupings


jnm2:
`Description` is something that people are not likely to want to group by and it introduces a lot of noise:

<img src="https://user-images.githubusercontent.com/8040367/36951103-6d498ed4-1fcd-11e8-864c-ca5e0f75cc78.png" width="300" />

Likewise, per-value `Timeout` and per-value `Repeat` don't seem that useful as things to group by.

Having them as hidden traits seems fine. That way you can still search for `Trait:"Repeat"` in Test Explorer.

jnm2:
Updated image.

rprouse:
Good idea, that is confusing. We should do a scan of the framework to make sure we get all of the properties users won't want to see in the UI.

CharliePoole:
I haven't looked at the code, but here's what I think happened. We used to suppress internal attributes from appearing as traits. They were marked as beginning with an underscore. For example, once upon a time, the descrition property was _DESCRIPTION. As time went on, we added attributes for our own use that didn't use this convention and some of the old internal attributes got "normal" names. I don't think it was intentional, but just happened.

jnm2:
That makes sense! I wondered when this ever came into play.

https://github.com/nunit/nunit3-vs-adapter/blob/3b9e3953fd4d3996858fc677744799939ed0afe6/src/NUnitTestAdapter/CategoryList.cs#L90-L91

OsirisTerje:
This then needs to be fixed in the framework, right?

CharliePoole:
Not necessarily. You could have a list of properties to bypass in the adapter **or** you could rename the properties in the framework.

jnm2:
I just finished telling @sharwell how to determine whether an `ITest` is being run with a single-threaded apartment. Renaming the framework properties is something I would hope to avoid unless we first provide a strongly-typed API for it.

rprouse:
Last time we changed names we broke other runners, so we probably don't want to do that. A list in the adaptor is a hacky solution, but it is backwards compatibile and won't break anything.

CharliePoole:
I agree and for another reason as well. Properties don't exist for the adapter to use for grouping. That's a usage that has been retrofitted to match what the Test Explorer team wanted when they first came up with Traits and wanted to showcase it for a Build conference. It's logical to make any adaptations in the adapter.

dzmitry-lahoda:
[...]

PR #586, opened 2019-01-18
Added filter for internal attributes  #474

OsirisTerje:
This PR adds a filter for removing the internal NUnit attributes from the output, so that they are not visible in the Test Explorer.  IN case you still want to see them, you can re-enable them using a runsettings property, ShowInternalProperties. 

The remaining non-resolved issue with this is the interaction with the source based discovery.  This *will* show the properties as trait during its discovery, but once you run them, the traits of these will be gone.


src/NUnitTestAdapter/CategoryList.cs
```diff
@@ -82,7 +93,15 @@ public IEnumerable<string> ProcessTestCaseProperties(XmlNode testNode, bool addT
...
 
-        private static bool IsInternalProperty(string propertyName, string propertyValue)
+        /// <summary>
+        /// List based on https://github.com/nunit/docs/wiki/Attributes
+        /// </summary>
+        private readonly List<string> internalProperties = new List<string>
+            { "Timeout", "Repeat", "Apartment", "Author", "Combinatorial", "Culture", "Datapoint",
+              "DatapointSource", "DefaultFloatingPointTolerance", "Description", "LevelOfParallelism",
+              "MaxTime", "NonParallelizable", "Order", "Pairwise", "Platform", "Random", "Range",
+              "RequiresThread", "Retry", "Sequential", "SetCulture", "SetUICulture", "SingleThreaded", "Values", "ValueSource" };
```
rprouse: These should map to the properties in the NUnit class `PropertyNames`. It would be a bit easier if they were in alphabetical order, but I think the list should be 

```c#
"Author", "ApartmentState", "Description", "LevelOfParallelism", "MaxTime", "ParallelScope", "Repeat", "RequiresThread", "SetCulture", "SetUICulture", "TestOf", "Timeout", "IgnoreUntilDate", "Order"
```

The link should also point to https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/PropertyNames.cs

rprouse: The difference is that not all attributes set properties, only those that derive from `PropertyAttribute` and those that do don't necessarily use the same names as the attributes themselves.

OsirisTerje: Thanks, and Fixed :-)

Issue #590, opened 2019-01-24
Support for Visual Studio 2019


OsirisTerje:
VSIX are still supported, Jan 24th, in Preview 2, so we update the NUnit3TestAdapter support to cover that version too.  As long as Visual Studio supports VSIX for adapters , we can do so too, it is not much changes anyway.

PR #591, opened 2019-01-24
Updated for supporting VS 2019 #590

OsirisTerje:


src/NUnit3TestAdapterInstall/source.extension.vsixmanifest
```diff
@@ -23,7 +23,7 @@ For running NUnit 2 tests, use the NUnit 2 adapter.</Description>
         <Asset Type="UnitTestExtension" d:Source="File" Path="NUnit3.TestAdapter.dll" />
     </Assets>
     <Prerequisites>
-        <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[15.0.25904.2,16.0)" DisplayName="Visual Studio core editor" />
-        <Prerequisite Id="Microsoft.VisualStudio.Component.ManagedDesktop.Prerequisites" Version="[15.0.25904.2,16.0)" DisplayName=".NET desktop development tools" />
+        <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[15.8.27729.1,)" DisplayName="Visual Studio core editor" />
```
rprouse: Won't this stop it from installing on early versions of Visual Studio 2017 like 15.0?

src/NUnit3TestAdapterInstall/source.extension.vsixmanifest
```diff
@@ -12,9 +12,9 @@ For running NUnit 2 tests, use the NUnit 2 adapter.</Description>
         <Tags>test, Testing, adapter, NUnit, TDD, NUnit3, UnitTesting</Tags>
     </Metadata>
     <Installation>
-        <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[11.0,16.0)" />
-        <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[11.0,16.0)" />
-        <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[11.0,16.0)"/>
+        <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[11.0,17.0)" />
+        <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[11.0,17.0)" />
+        <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[11.0,17.0)" />
```
rprouse: I thought Visual Studio 2019 was dropping support for VSIX based test adapters. Am I mistaken?

src/NUnit3TestAdapterInstall/source.extension.vsixmanifest
OsirisTerje: Good point.  I The minimum version has been like this for quite some time, and no complaints..... I think we did this when converting over to VS 2017.  I haven't checked with earlier versions, but I'll test it out to make sure.

src/NUnit3TestAdapterInstall/source.extension.vsixmanifest
OsirisTerje: VS 2019 still supports vsix, and even if it is on its way to be deprecated, it seems it will still take some time.  As long as VS supports vsix I think we should just continue supporting it.  It is pretty easy to do, ... so far....

rprouse:
I changed the Azure DevOps build pipeline to use `dotnet restore` instead of `NuGet.exe restore` plus pushed a minor change to the installer project file. Still waiting for AppVeyor, but Azure DevOps build is now succeeding.

src/NUnit3TestAdapterInstall/source.extension.vsixmanifest
rprouse: These versions are a minimum or newer. Wouldn't it be safer to leave them set where they were?

OsirisTerje: I overlooked that, will fix.

PR #592, opened 2019-01-25
Issue302 native execution

### LINKED ISSUE #302: BadImageFormatException building solution with unmanaged projects

OskiKervinen-MF:
Merge request !369 only fixes #302 for test discovery, not for test execution. There you still get an error when a native binary is given to NUnit, because the BadFormatException is wrapped in a NUnitEngineException and that case was not checked in the execution code path.

Added a test to detect this and a fix for the problem.

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestExecutor.cs
+++ b/src/NUnitTestAdapter/NUnit3TestExecutor.cs
@@ -299,6 +299,15 @@ namespace NUnit.VisualStudio.TestAdapter
                 // we skip the native c++ binaries that we don't support.
                 TestLog.Warning("   Assembly not supported: " + assemblyPath);
             }
+            catch( NUnitEngineException e )
+            {
+                if( e.InnerException is BadImageFormatException )
+                {
+                    // we skip the native c++ binaries that we don't support.
+                    TestLog.Warning( "   Assembly not supported: " + assemblyPath );
+                }
+                throw;
+            }
             catch (FileNotFoundException ex)
             {
                 // Probably from the GetExportedTypes in NUnit.core, attempting to find an assembly, not a problem if it is not NUnit here
--- a/src/NUnitTestAdapterTests/TestExecutionTests.cs
+++ b/src/NUnitTestAdapterTests/TestExecutionTests.cs
@@ -166,6 +166,42 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
             VerifyAttachment(attachmentSet.Attachments[1], FixtureWithAttachment.Attachment2Name, FixtureWithAttachment.Attachment2Description);
         }
 
+#if NET46
+        [Test]
+        public void NativeAssemblyProducesWarning()
+        {
+            // Create a fake environment.
+            var context = new FakeRunContext();
+            var fakeFramework = new FakeFrameworkHandle();
+
+            // Find the native exaple dll.
+            var path = Path.Combine( TestContext.CurrentContext.TestDirectory, "NativeTests.dll" );
+            Assert.That( File.Exists( path ) );
+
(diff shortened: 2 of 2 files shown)


OsirisTerje:
Looks good to me too, just want to create a duplicate issue to connect this to, since the #302 is closed already.  I'll do this today, merge it later today too.

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -299,6 +299,15 @@ private void RunAssembly(string assemblyPath, TestFilter filter)
...
                 TestLog.Warning("   Assembly not supported: " + assemblyPath);
             }
+            catch( NUnitEngineException e )
+            {
+                if( e.InnerException is BadImageFormatException )
+                {
+                    // we skip the native c++ binaries that we don't support.
+                    TestLog.Warning( "   Assembly not supported: " + assemblyPath );
+                }
+                throw;
```
jnm2, line 309: This `throw;` means that NUnitEngineExceptions are no longer handled at all. They used to be handled in the `catch (Exception)` section below.

jnm2: Fixed by #804

PR #620, opened 2019-03-17
Find MSBuild in Visual Studio 2019

### LINKED ISSUE #619: Cake build failes with Visual Studio 2019
My current machine only has Visual Studio 2019 installed. Running the Cake build;

```
.\build.ps1 -Target Build
```

Gives the error

```
========================================
NuGetRestore
========================================
Executing task: NuGetRestore
Restoring NuGet Packages for the Adapter Solution
An error occurred when executing task 'NuGetRestore'.
Error: One or more errors occurred.
        MSBuild: Could not locate executable.
```

Removing the line `ToolVersion = MSBuildToolVersion.VS2017` doesn't help. We should be able to build using newer versions of Visual Studio even if our minimum version is lower.

I am seeing the same thing in the NUnit  project, (nunit/nunit#3193) but not in the NUnit Console project. I will look at the `MSBuildSettings` there to see if I can figure it out.

rprouse:
Fixes #619 

Visual Studio 2019 has changed the directory that MSBuild is in.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -104,12 +104,7 @@ Task("NuGetRestore")
     .Does(() =>
 {
     Information("Restoring NuGet Packages for the Adapter Solution");
-    MSBuild(ADAPTER_SOLUTION, new MSBuildSettings
-    {
-        Configuration = configuration,
-        Verbosity = Verbosity.Minimal,
-        ToolVersion = MSBuildToolVersion.VS2017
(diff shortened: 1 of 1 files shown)


rprouse:
The AppVeyor build failed because of an issue in the acceptance tests. It looks transient, so I restarted.

PR #637, opened 2019-07-10
Handling current directory  #609  #303

### LINKED ISSUE #303: Directory.GetCurrentDirectory is C:\WINDOWS\system32 for NUnit3TestAdapter

### LINKED ISSUE #609: Request to change working directory from Windows/System32 to another directory
Hi

This is not yet another "why the working directory points to Windows/System32" , I've read the previous issues about the issue and I understand the reasons behind why the working directory does not point to where the test is located.

My concern is that, pointing the working directory to Windows/System32 _by default_ is dangerous, since a naive test can inadvertently write files to that directory, or any third party application, expecting to output to the working directory, writes files to it.

Which is exactly what just happened to me. I was pointing my tests to use TestDirectory and WorkDirectory as NUnit requires.... but by using a third party assembly which happens to write log files into the current working directory, I've just found that my Windows/System32 directory is full of trash.

Could it be possible to change this behaviour by default, so, the current working directory points to a safer **sink** directory?  for example, the Temp Files directory?

OsirisTerje:
Fixes #609  Fixes #303

### DIFF
--- a/src/NUnitTestAdapter/NUnitTestAdapter.cs
+++ b/src/NUnitTestAdapter/NUnitTestAdapter.cs
@@ -134,6 +134,7 @@ namespace NUnit.VisualStudio.TestAdapter
             {
                 Settings.Load(context);
                 TestLog.Verbosity = Settings.Verbosity;
+                InitializeForbiddenFolders();
                 SetCurrentWorkingDirectory();
             }
             catch (Exception e)
@@ -141,7 +142,7 @@ namespace NUnit.VisualStudio.TestAdapter
                 TestLog.Warning("Error initializing RunSettings. Default settings will be used");
                 TestLog.Warning(e.ToString());
             }
-            InitializeForbiddenFolders();
+            
         }
 
         public void InitializeForbiddenFolders()
--- a/src/NUnitTestAdapter/NUnitTestAdapter.cs
+++ b/src/NUnitTestAdapter/NUnitTestAdapter.cs
@@ -41,6 +41,7 @@ using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
 using NUnit.Common;
 using NUnit.Engine;
+using System.Linq;
 
 namespace NUnit.VisualStudio.TestAdapter
 {
@@ -131,7 +132,7 @@ namespace NUnit.VisualStudio.TestAdapter
             {
                 Settings.Load(context);
                 TestLog.Verbosity = Settings.Verbosity;
-               
+                SetCurrentWorkingDirectory();
             }
             catch (Exception e)
             {
@@ -140,7 +141,21 @@ namespace NUnit.VisualStudio.TestAdapter
             }
         }
 
-        
+        static readonly List<string> ForbiddenFolders = new List<string> { "windows", "program files (x86)", "program files", "programdata" };
+
+        private static void SetCurrentWorkingDirectory()
+        {
+            var dir = Directory.GetCurrentDirectory();
+            bool ok = CheckDirectory(dir);
+            if (!ok)
+                Directory.SetCurrentDirectory(Path.GetTempPath());
+        }
+
+        public static bool CheckDirectory(string dir)
+        {
+            var split = new List<string>(dir.Split('\\', '/'));
+            return !split.Any(segment => ForbiddenFolders.Contains(segment.ToLower()));
+        }
 
         protected ITestRunner GetRunnerFor(string assemblyName)
         {
@@ -222,7 +237,7 @@ namespace NUnit.VisualStudio.TestAdapter
                 Directory.CreateDirectory(workDir);
             package.Settings[PackageSettings.WorkDirectory] = workDir;
             WorkDir = workDir;
-         //   CreateTestOutputFolder(workDir);
+            //   CreateTestOutputFolder(workDir);
             return package;
         }
 
--- a/src/NUnitTestAdapter/NUnitTestAdapter.cs
+++ b/src/NUnitTestAdapter/NUnitTestAdapter.cs
@@ -147,7 +147,7 @@ namespace NUnit.VisualStudio.TestAdapter
 
         public void InitializeForbiddenFolders()
         {
-            ForbiddenFolders = new List<string>
+            ForbiddenFolders = new []
             {
                 Environment.GetEnvironmentVariable("ProgramW6432"),
                 Environment.GetEnvironmentVariable("ProgramFiles(x86)"),
--- a/src/NUnitTestAdapter/NUnitTestAdapter.cs
+++ b/src/NUnitTestAdapter/NUnitTestAdapter.cs
@@ -168,8 +168,8 @@ namespace NUnit.VisualStudio.TestAdapter
         /// </summary>
         public bool CheckDirectory(string dir)
         {
-            var checkdir = (dir.EndsWith("\\") ? dir : dir + "\\").ToLower();
-            return ForbiddenFolders.Any(o => checkdir.StartsWith(o));
+            var checkdir = (dir.EndsWith("\\") ? dir : dir + "\\");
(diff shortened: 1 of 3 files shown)


jnm2:
@OsirisTerje Did you want to autoclose #609 and #303?

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -140,7 +141,21 @@ protected void Initialize(IDiscoveryContext context, IMessageLogger messageLogge
             }
         }
 
-        
+        static readonly List<string> ForbiddenFolders = new List<string> { "windows", "program files (x86)", "program files", "programdata" };
```
jnm2: I think I'd prefer to do a case insensitive StartsWith comparison between `Directory.GetCurrentDirectory()` (always absolute) and `Path.GetFullPath(Environment.GetFolderPath(...))` for each of these things.

@vpenades also asked about `C:\Users` with the exception of `Path.GetTempPath()`; what do you think of that?

vpenades: Instead of using a plain list of paths, like "windows", "program files (x86)", shouldn't be better to use `System.Environment.GetSpecialFolderPath`?

Also, although it's rarely done, users can change windows and program files location, also, it's sometimes localized, for example, "program files (x86)" is "archivos de programa (x86)" in spanish.

So I think it's better to ask the OS for these folders.

OsirisTerje: @vpenades  Good point!  
@jnm2 There are so many places below Users where it would be legal to use, so I could not really see what to exclude there that made sense.  Did I overlook something?

OsirisTerje:
@jnm2 Autoclose ?  Only when we merges this in......

src/NUnitTestAdapter/NUnitTestAdapter.cs
jnm2: > There are so many places below Users where it would be legal to use, so I could not really see what to exclude there that made sense.

This is a good point. Let's not guess.

%programdata% seems just as acceptable to write into as %appdata% or %localappdata%.

Reversing direction, maybe for now only block it if it's under `Environment.GetFolderPath(Environment.SpecialFolder.Windows)`?

jnm2:
> Autoclose ? Only when we merges this in......

Okay. In order to have it autoclose when the PR merges in, you have to put `Closes #609, closes #303` in the body of the top comment. The numbers are ignored if they're in the PR title.

src/NUnitTestAdapter/NUnitTestAdapter.cs
vpenades: @jnm2 I agree it's fine to allow users and programs data.

I would disallow everything under windows and programs files x86 and x64

OsirisTerje: Under .net 3.5 which we are using, we don't have x86 program files, and we don't have windows.   We can look using environment variables, but then again, I would expect those to be locale sensitives.

vpenades: @OsirisTerje 

I am not sure about .net3.5 ... the problem is not only about localization; users can choose to install windows or program files directory in a path different than the default (or even change the drive unit)

Getting the directories on net core... the only solution I could find around is by using [Microsoft.Windows.Compatibility package](https://www.nuget.org/packages/Microsoft.Windows.Compatibility) which I guess it will work under windows, and throw exceptions or return null under other OS.

OsirisTerje:
@jnm2 @vpenades  Ok guys, now it works for ProgramFiles (inkl x86) and Windir, and it should work both for -net std framework and .net core.

src/NUnitTestAdapter/NUnitTestAdapter.cs
jnm2: @OsirisTerje This is a separate thing, but the minimum version of .NET Core that Microsoft supports as of a month ago is .NET Core 2.1. It would make life easier if we no longer supported anything earlier as well.

jnm2: I would be very surprised if environment variable *names* were localized; if they are, what's the point of having them? I would expect environment variable *values* to be localized.

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -131,16 +134,43 @@ protected void Initialize(IDiscoveryContext context, IMessageLogger messageLogge
...
         }
 
-        
+        /// <summary>
+        /// If a directory matches one of the forbidden folders, then we should reroute, so we return true in that case
+        /// </summary>
+        public bool CheckDirectory(string dir)
+        {
+            var checkdir = (dir.EndsWith("\\") ? dir : dir + "\\").ToLower();
+            return ForbiddenFolders.Any(o => checkdir.StartsWith(o));
```
jnm2: I think there's a correctness issue with ToLower. The thing to do as I understand it is to never use ToLower or ToUpper (or their Invariant versions) but instead to use `checkdir.StartsWith(o, StringComparison.OrdinalIgnoreCase)`.

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -131,16 +134,43 @@ protected void Initialize(IDiscoveryContext context, IMessageLogger messageLogge
...
             {
                 TestLog.Warning("Error initializing RunSettings. Default settings will be used");
                 TestLog.Warning(e.ToString());
             }
+            
+        }
+
+        public void InitializeForbiddenFolders()
+        {
+            ForbiddenFolders = new List<string>
```
jnm2: Could be shorter, `new[]` instead of `new List<string>` (not important)

src/NUnitTestAdapter/NUnitTestAdapter.cs
jnm2: http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html has more info

src/NUnitTestAdapter/NUnitTestAdapter.cs
OsirisTerje: Done

src/NUnitTestAdapter/NUnitTestAdapter.cs
OsirisTerje: Cool, that was a new one :-).  But done !

src/NUnitTestAdapter/NUnitTestAdapter.cs
OsirisTerje: @jnm2  We should have some survey out for stuff like this.  Or some metrics.......

src/NUnitTestAdapter/NUnitTestAdapter.cs
```diff
@@ -131,16 +134,43 @@ protected void Initialize(IDiscoveryContext context, IMessageLogger messageLogge
...
+        }
+
+        public void InitializeForbiddenFolders()
+        {
+            ForbiddenFolders = new []
+            {
+                Environment.GetEnvironmentVariable("ProgramW6432"),
+                Environment.GetEnvironmentVariable("ProgramFiles(x86)"),
+                Environment.GetEnvironmentVariable("windir"),
+            }.Where(o => !string.IsNullOrEmpty(o)).Select(o=>o.ToLower()+@"\").ToList();
```
jnm2:
```suggestion
            }.Where(o => !string.IsNullOrEmpty(o)).Select(o=>o + @"\").ToList();
```

jnm2: Other than this, LGTM!

jnm2: @OsirisTerje Was there a reason you needed ToLower()?

OsirisTerje: The ToLower could be removed here too, you're right.  Next time around, as I did the merge :-)

PR #652, opened 2019-08-29
Adding prefilter featureflag to runsettings, with default off. 

### LINKED ISSUE #651: Add featureflag to enable/disable pre-filtering
We have issues in some cases with pre-filtering, ref #648, #649, #650. 

In order to avoid this, we'll add a feature flag to the runsettings file, which will be default off.  You will then have to opt-in to get the performance boost.  

In short what is affected: 
- Any version of NUnit below 3.11.0
- Any test assemblies using SetupFixture
- TestCaseData using SetName instead of SetArgDisplayNames. 

The SetupFixture has to be fixed in the framework, See https://github.com/nunit/nunit/issues/3356  ,  and when this is fixed we can *consider* moving pre-filtering option to default on.

OsirisTerje:
Fixes  #651.  See related issues in same issue.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -12,7 +12,7 @@ var configuration = Argument("configuration", "Release");
 // SET PACKAGE VERSION
 //////////////////////////////////////////////////////////////////////
 
-var version = "3.15.0";
+var version = "3.15.1";
 var modifier = "";
 
 var dbgSuffix = configuration.ToLower() == "debug" ? "-dbg" : "";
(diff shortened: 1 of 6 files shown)


OsirisTerje:
Pre-release here:  There is now a beta version of the fix in https://www.myget.org/feed/nunit/package/nuget/NUnit3TestAdapter/3.15.1-dev-01134  .

PR #663, opened 2019-10-19
Stop building for EOL versions of .NET Core

jnm2:
Same as https://github.com/nunit/nunit/pull/3383.

Best reviewed by commit.


global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "2.1.802",
+    "rollForward": "feature"
```
jnm2: Developers usually already have the .NET Core SDK installed, usually a newer version than required. `"rollForward"` works with the .NET Core 3.0 SDK and means that `2.1.802` becomes the minimum version instead of requiring an exact version. `"feature"` means that any `2.1.*` SDK is accepted.

(See links in https://github.com/dotnet/docs/issues/14480)

If a contributor does not have a 2.1.* SDK installed that is >= 2.1.802, they will get a message like this (I changed the versions for demonstration purposes):

```cs
A compatible installed .NET Core SDK for global.json version [2.1.802] from [C:\Users\Joseph\Source\Repos\nunit3-vs-adapter\global.json] was not found
Install the [2.1.802] .NET Core SDK or update [C:\Users\Joseph\Source\Repos\nunit3-vs-adapter\global.json] with an installed .NET Core SDK:
  2.1.801 [C:\Program Files\dotnet\sdk]
```

global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "2.1.100",
+    "rollForward": "feature"
+  }
+}
```
jnm2: Developers usually already have the .NET Core SDK installed, usually a newer version than required. `"rollForward"` works with the .NET Core 3.0 SDK and means that `2.1.100` becomes the minimum version instead of requiring an exact version. `"feature"` means that any `2.1.*` SDK is accepted.

(See links in https://github.com/dotnet/docs/issues/14480)

If a contributor does not have a 2.1.* SDK installed that is >= 2.1.100, they will get a message like this (I changed the versions for demonstration purposes):

```cs
A compatible installed .NET Core SDK for global.json version [2.1.100] from [C:\Users\Joseph\Source\Repos\nunit3-vs-adapter\global.json] was not found
Install the [2.1.100] .NET Core SDK or update [C:\Users\Joseph\Source\Repos\nunit3-vs-adapter\global.json] with an installed .NET Core SDK:
  3.0.100 [C:\Program Files\dotnet\sdk]
```

I chose `2.1.100` because there's no reason I know of to require a newer SDK feature band than `1xx`.

NUnit3TestAdapter.sln
```diff
@@ -4,14 +4,12 @@ VisualStudioVersion = 16.0.28606.126
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{7CE30108-5D81-4850-BE6B-C8BCA35D3592}"
 	ProjectSection(SolutionItems) = preProject
-		acceptance.cake = acceptance.cake
```
OsirisTerje: Did you remove the acceptance parts ?

OsirisTerje:
Looks like a bit too much to do a detailed review.

NUnit3TestAdapter.sln
jnm2: `acceptance.cake` was removed a while back when I added the acceptance test csproj. This should have been removed in that PR.

jnm2:
CI is failing for the master branch because Azure Pipelines builds run on the master branch but they do not run for PR validation. I'm going to fix them. It would also be easier to coordinate between merging the PR and updating the build definition if we converted the Azure builds to a single YAML-based build.

PR #666, opened 2019-10-21
Move to YAML-based build for Azure Pipelines and fix silently skipped tests

jnm2:
See commit messages.

Here is the test results page. You can also see bin and package artifacts in the top right corner.

https://nunit.visualstudio.com/NUnit/_build/results?buildId=2171:

image


jnm2:
I enabled CI and PR triggers for the YAML-based build definition and disabled them for the others.

Issue #622, opened 2019-03-26
An exception occurred while invoking executor 'executor://nunit3testexecutor/': Incorrect format for TestCaseFilter Error: Missing ')'.


jerryk414:
**ENVIRONMENT**
__________________________________________________________
**NUnit**: 3.10.1
**NUnit3TestAdapter**: 3.10.0
**Visual Studio**: 15.9.2
**.NET Platform**: .NET Core 2.1

**DESCRIPTION**
__________________________________________________________
When executing a category of tests, if a test is included in the list of tests that contains an opening parenthesis, but does not contain the closing parenthesis, the following error may occur, and the test run will fail:

> An exception occurred while invoking executor 'executor://nunit3testexecutor/': Incorrect format for TestCaseFilter Error: Missing ')'. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.

**REPRODUCTION**
__________________________________________________________
This issue can only be reproduced when selecting a group of tests to run - selecting individual tests, or selecting 'Run All' will no reproduce the issue.

It appears that reproducing the issue requires there to be a test that includes an opening parenthesis but does not include a closing parenthesis, and a test that ends with the closing bracket character (']').

You should be able to reproduce the issue by using the following test code:
```
public class TestNameRepro
{
    #region TestCaseSource

    protected static IEnumerable TestName_TestCases
    {
        get
        {
            // Adding a parenthesis to the end of this test name will stop the exception from throwing (e.g. $"TestName(...)")
            yield return new TestCaseData(1).SetName($"TestName(...");

            // Cannot be duplicated without a second test included that ends with a ']'
            yield return new TestCaseData(2).SetName($"TestName(...)]");
        }
    }

    #endregion     
    [Test, TestCaseSource(nameof(TestName_TestCases))]
    public void TestName(int input)
    {
        // Irrelevant  
    }
}
```

Once built, categorize the test explorer using the 'Class' category selection, and select the test suite 'TestNameRepro' and click 'Run Selected Tests':
![image](https://user-images.githubusercontent.com/11689280/55029292-209dec80-4fe0-11e9-8445-b61e44c8c232.png)

The 'Tests' output window should display the error mentioned above:
> An exception occurred while invoking executor 'executor://nunit3testexecutor/': Incorrect format for TestCaseFilter Error: Missing ')'. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.

OsirisTerje:
[...]

Issue #505, opened 2018-05-15
Tracking: VSTest filtering, FullyQualifiedName translation and selected-test execution


peterwald:
This is a tracking (Epic) issue for the recurring family of problems around translating VSTest / MTP filters into NUnit filters, constructing and parsing `FullyQualifiedName`, and keeping test identities stable between discovery and execution. It supersedes the original 2018 scope ("Standardized FullyQualifiedName Support", see https://github.com/Microsoft/vstest-docs/pull/133 and #668).

Symptoms across these issues are usually one of:

- a test is **discovered but never executed** (`No test matches the given testcase filter ...`),
- `Incorrect format for TestCaseFilter` / `Unexpected FQN ... in selection expression` exceptions,
- phantom, never-run entries in Test Explorer,
- the selection is lost on rerun (Azure DevOps / `dotnet test --filter`).

The grouping below is editorial; the native parent/sub-issue links carry progress. Search label: `is:FQNIssue`.

## A. FQN parsing / escaping / special characters

Test is discovered, the filter round-trip mangles the name, the test won't run.

- #1488 - escaped operators `\= \& \| \! \~ \\` discovered but never executed
- #1490 - space before the argument list -> `Unexpected FQN ... in selection expression`
- #1437 - filtering on test name broken in adapter 6.0.0+ (regression)
- #1405 - pipe character `|` in a string argument
- #1377 - tests fail to run after migration to MTP + newest NUnit
- #1349 - MTP filter string includes an unrecognized escape sequence
- #1097 - `).` inside a string argument not detected
- #935 - F# discriminated-union `TestCaseSource` -> `Incorrect format for TestCaseFilter` (External)
- #876 - spaces and special characters unsupported in TestFilter
- #807 - VSTest retry fails with spaces / `&` in a TestCase parameter
- #782 - `IEnumerable` tuple arguments from `TestCaseSource` not run
- #761 - `￿` argument
- #742 - several characters in the name do not work
- #654 - `).` in a TestCase argument breaks Test Explorer grouping

## B. Discovery / execution identity mismatch

Discovery and execution disagree on the test's identity, so the selection cannot be resolved.

- #489 - VS15.6 Real-Time Test Discovery yields unrunnable NUnit tests
- #607 - `No test matches the given testcase filter FQN=...` for `TestCaseData.SetName` names
- #623 - Azure DevOps rerun with `TestFixtureSource` -> `Incorrect format for TestCaseFilter`
[...]

PR #668, opened 2019-10-28
Resolve issue with custom tests names and test Ids

johnmwright:
This PR is to address #622 

Key highlights:

- The FullyQualifiedName is based on Type+Method and independent of DisplayName.
- The TestCaseId is unique across all test cases (including data-driven ones) and deterministic between discovery sessions.

## Background information:
- The VS test system [expects the `TestCase.FullyQualifiedName` to be the actual test method name](https://github.com/microsoft/vstest-docs/blob/master/RFCs/0017-Managed-TestCase-Properties.md#specification) (not the display name), optionally including param _types_ (not values). Basically, the details that would have been returned from `caller.GetType().FullName`
- If you don't provide a value for `TestCase.Id`, the class with [create a GUID](https://github.com/microsoft/vstest/blob/829659f9ca95ccfb54480948d8e472651dd0ff5d/src/Microsoft.TestPlatform.ObjectModel/TestCase.cs#L96) value [based on the `TestCase.FullyQualifiedName`](https://github.com/microsoft/vstest/blob/829659f9ca95ccfb54480948d8e472651dd0ff5d/src/Microsoft.TestPlatform.ObjectModel/TestCase.cs#L240). 
- When running tests, if you instruct the system to run a subset of tests (ie: not "all" tests), a filter is provided by VS that includes a list of tests by `FullyQualifiiedName`, which the NUnitAdapter attempt to apply in `TfsTestFilter.CheckFilter(TestCase)`.
- The test hierarchy in the VS Test Explorer UI uses the `FullyQualifiedName` to determine the namespace groupings
- If multiple tests results are associated with the same TestCase instance (ie: `TestCase.Id` is equal), then [it will show the test once in the UI, with multiple result values](https://github.com/microsoft/vstest-docs/blob/master/RFCs/0017-Managed-TestCase-Properties.md#tests-returning-multiple-results), such as this (which is how MSTest-based tests display results when a method is used for test cases):
image

## How this issue manifested:
- The [current logic](https://github.com/nunit/nunit3-vs-adapter/blob/e12dcb031828181148397058ab43d07e1db2fcad/src/NUnitTestAdapter/TestConverter.cs#L167) sets the `DisplayName` and `FullyQualifiedName` to the display name of the test, which break the contract with VS if the user has utilized the `SetName()` method to override the display name value.  While VS doesn't strictly enforce the `FullyQualifiedName` is an actual method, it does have [some minor validation](https://github.com/microsoft/vstest/blob/829659f9ca95ccfb54480948d8e472651dd0ff5d/src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs#L114) to make sure some special characters are used as expected, such as paired parenthesis, resulting in an error such as:
> An exception occurred while invoking executor 'executor://nunit3testexecutor/': Incorrect format for TestCaseFilter Error: Missing '('. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.

## How I addressed the issue:
- First, but setting the `TestCase.FullyQualfiedName` to the actual test method name, this conforms with the expectations from VS and fixes the parsing issues.  Since the NUnit-parsed data includes a `classname` and `methodname` attribute, the FQN can be created from that (or, at least, a close-enough version).
- By using the method type+name for the `FullyQualifiedName`, if the adapter doesn't explicitly set the `TestCast.Id`, it will result in each TestCase having the default Id based on the FQN, as described in the Background Information above. Meaning: When the adapter recreates the TestCase object for association with the TestResult object returned with the test results, it would return several results with the same TestCase.Id value, since each of those would have the same FQN. This resulted in the UI displaying the results under the same Test (similar to the screenshot above).  This does not match the existing behavior, where each test case is listed as it's own node in the test case explorer.  Additionally, only the first DisplayName was shown in the UI, since it ignored tests with identical Ids, due to [Ids needing to be unique](https://github.com/microsoft/vstest-docs/blob/master/RFCs/0017-Managed-TestCase-Properties.md#uniqueness). Since NUnit already applies a unique id to each test, which is consistent across test runs, I decided to resolve this issue by assigning the `TestCase.Id` property using the NUnit id attribute value.  This will give each TestCase a unique Id even if they have the same `FullyQualifiedName`, resulting in the individual nodes that is currently used.

## Impacts of changes
I primarily used two sets of tests while working on this PR.  The first is a test class I put together from the use cases listed in #622, which I put into [this gist](https://gist.github.com/johnmwright/bf374e3ac588fd9ceb8476148578772c) and ultimately added as a unit test in this PR.  The other is the [sample solution](https://github.com/johnjaylward/nUnitExpressionTest) provided in comments to #622 . I'll be using these to describe the impacts of this change.

Before the changes, when the problematic parameterized tests were run, you would see:
- the actual name of the test (example: `Repo_jwright.DemoTest2` in screenshot 1), in addition to the parameterized custom names (ex: `Repo_jwright.Name with mismatched quote '"c'`) as peers.
- Tests which had mismatched special characters were represented as inconclusive (aka "Not Run") at the end of the test run, due to the "Incorrect format for TestCaseFilter" error (ex: `Repo_jwright.DemoTest2` in screenshot 1)
- Parameterized tests were grouped across many nodes when custom names were given (see various nodes for `ExpressionTransatorTest` in screenshot 2)

Screenshot 1:
image

Screenshot 2:
image

After the changes:
- The Test name itself is no longer a node with results, but rather it's a grouping node (see: `Repo_jwright.DemoTest2` in screenshot 1)
- Tests which have mismatched special characters no longer throw an error and now show their proper results
- Parameterized tests are grouped under the same node based on the FullyQualifiedTestName value (see screenshot 2)

Screenshot 1:
image

image

## Possible Breaking Change
Because this PR changes how the `TestCase.Id` property is generated, this will result in test cases having different IDs than they did before the change was implemented.  Any systems that expect the test ID value to remain the same across history will lose their tracking of test cases.  

This is specifically called out in the inline comments of [`EqtHash.GuidFromString`](https://github.com/microsoft/vstest/blob/829659f9ca95ccfb54480948d8e472651dd0ff5d/src/Microsoft.TestPlatform.ObjectModel/Utilities/EqtHash.cs#L26) in the VSTest codebase, but with the references between TFS/MsftDevOps Bugs and TFS/MsftDevOps workitems.  I am unsure what, if any, associations TFS/MsftDevOps places with test run results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -330,3 +330,5 @@ tools/
 # VisualStudioCode Patch
 # Ignore all local history of files
 .history
+/Dump
+/src/NUnitTestAdapterTests/Dump
--- a/src/NUnit.TestAdapter.Tests.Acceptance/TestSourceWithCustomNames.cs
+++ b/src/NUnit.TestAdapter.Tests.Acceptance/TestSourceWithCustomNames.cs
@@ -1,0 +1,130 @@
+﻿using System.Linq;
+using NUnit.Framework;
+using NUnit.VisualStudio.TestAdapter.Tests.Acceptance.WorkspaceTools;
+
+namespace NUnit.VisualStudio.TestAdapter.Tests.Acceptance
+{
+    public sealed class TestSourceWithCustomNames : AcceptanceTests
+    {
+        private static void AddTestsCs(IsolatedWorkspace workspace)
+        {
+            workspace.AddFile("Tests.cs", @"
+                using System;
+                using System.Collections;
+                using System.Collections.Generic;
+                using NUnit.Framework;
+
+                namespace Test
+                {
+                    public class Tests
+                    {
+                        [Test]
+                        [TestCaseSource(nameof(TestCaseSourceMethod))]
+                        public void PassingTestStr(object arg)
+                        {
+                            Assert.Pass();
+                        }
+
+
+                        private static IEnumerable<TestCaseData> TestCaseSourceMethod()
+                        {
+                            yield return new TestCaseData(""Name with mismatched parenthesis 'pp:-) :-)'"").SetName(""Name with mismatched parenthesis 'pp:-) :-)'"");
+                            yield return new TestCaseData(""Name with mismatched quote '\""c'"").SetName(""Name with mismatched quote '\""c'"");
+                            
+                            // Adding a parenthesis to the end of this test name will stop the exception from throwing (e.g. $""TestName(...)"")
+                            yield return new TestCaseData(1).SetName($""TestName(..."");
+
+                            // Cannot be duplicated without a second test included that ends with a ']'
+                            yield return new TestCaseData(2).SetName($""TestName(...)]"");
+                        }
+
+                        [Test]
+                        [TestCase(typeof(IEnumerable<(String oneValue, Int32 twoValue)>))]
+                        public void UnitTest_TestCaseWithTuple_TestIsNotExecuted(Type targetType)
+                        {
+                            Assert.That(targetType == typeof(IEnumerable<(String oneValue, Int32 twoValue)>), Is.True);
+                        }
+
+                        [Test, TestCaseSource(nameof(SourceA))]
+                        public void TestA((int a, int b) x, int y) { }
+                        public static IEnumerable SourceA => new[] {new TestCaseData((a: 1, b: 2), 5)};
+
+                        [Test, TestCaseSource(nameof(SourceB))]
+                        public void TestB(int y, (int a, int b) x) { }
+                        public static IEnumerable SourceB => new[] {new TestCaseData(5, (a: 1, b: 2))};
+
+                        [Test, TestCaseSource(nameof(SourceC))]
+                        public void TestC((int a, int b) x, int y) { }
+                        public static IEnumerable SourceC => new[] {new TestCaseData((a: 1, b: 2), 5).SetArgDisplayNames(""a+b"", ""y"")};
+
+                        [Test(), TestCaseSource(typeof(CaseTestData), nameof(CaseTestData.EqualsData))]
+                        public void EqualsTest(Case case1, Case case2)
+                        {
+                            Assert.AreEqual(case1, case2);
+                        }
+
+                        public class CaseTestData
+                        {
+                            public static IEnumerable EqualsData()
+                            {
+                                yield return new object[] { new Case { Name = ""case1"" }, new Case { Name = ""case1"" } };
+                            }
+                        }
+
+                        public class Case
+                        {
+                            public string Name;
+                            public override string ToString() => Name;
+                            public override bool Equals(object obj) => obj is Case other && other.Name == this.Name;
+                        }
+
+                    }
+                }");
+        }
+
+        [Test]
+        [TestCase("net47")] //test code requires ValueTuple support, so can't got to net35
+        [TestCase("netcoreapp2.1")]
+        public static void Single_target_csproj(string targetFramework)
+        {
+            var workspace = CreateWorkspace()
+                .AddProject("Test.csproj", $@"
+                    <Project Sdk='Microsoft.NET.Sdk'>
+
+                      <PropertyGroup>
+                        <TargetFramework>{targetFramework}</TargetFramework>
+                      </PropertyGroup>
+
+                      <ItemGroup>
+                        <PackageReference Include='Microsoft.NET.Test.Sdk' Version='*' />
+                        <PackageReference Include='NUnit' Version='*' />
+                        <PackageReference Include='NUnit3TestAdapter' Version='{NuGetPackageVersion}' />
+                      </ItemGroup>
+
+                    </Project>");
+
+            AddTestsCs(workspace);
+
+            workspace.MSBuild(restore: true);
+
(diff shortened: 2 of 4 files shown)


johnmwright:
Note: I believe this change may fix other open issues. I will associate them to this PR as I verify them.

johnmwright:
Note: I see the CI failures and will investigate them as soon as I can.

johnmwright:
This "Issue" (placeholder) is related to this PR: https://github.com/nunit/nunit3-vs-adapter/issues/505

johnjaylward:
@johnmwright This looks like a huge improvement. I love that all my tests now show under a single grouping as that's how I expected them to show in the first place. Great work, and a wonderful explanation in the PR!

It looks like some test cases may need to be updated though to match the changes: 
```
Expected: "NUnit.Tests.FixtureWithTestCases.MethodWithParameters(9,11)"
  But was:  "NUnit.Tests.FixtureWithTestCases.MethodWithParameters"
```

johnmwright:
Yeah, there's a few test failures that I missed in my local runs.  I'll need some time to address them.

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -159,20 +160,33 @@ public struct TestResultSet
...
+            var className = testNode.GetAttribute("classname");
+            var methodName = testNode.GetAttribute("methodname");
+           
+            // VS expected FullName to be the actual class+type name,optionally with parameter types
+            // in parenthesis, but they must fit the pattern of a value returned by object.GetType().
+            // It should _not_ include custom name or param values (just their types).
+            // However, the "fullname" from NUnit's file generation is the custom name of the test, so
+            // this code must convert from one to the other.
+            // Reference: https://github.com/microsoft/vstest-docs/blob/master/RFCs/0017-Managed-TestCase-Properties.md
+            var fullname = className + "." + methodName;
```
aitch-hat: Nit: It might be worth using a variable name other than `fullname` so that this isn't confused with `testNode.GetAttribute("fullname")`.

src/NUnit.TestAdapter.Tests.Acceptance/TestSourceWithCustomNames.cs
```diff
@@ -0,0 +1,130 @@
...
+                            
+                            // Adding a parenthesis to the end of this test name will stop the exception from throwing (e.g. $""TestName(...)"")
+                            yield return new TestCaseData(1).SetName($""TestName(..."");
+
+                            // Cannot be duplicated without a second test included that ends with a ']'
+                            yield return new TestCaseData(2).SetName($""TestName(...)]"");
+                        }
+
+                        [Test]
+                        [TestCase(typeof(IEnumerable<(String oneValue, Int32 twoValue)>))]
```
aitch-hat: What is the reason for using `String` and `Int32` here when `string` and `int` have been used in the rest of the class?

johnmwright: Oh, that was unintended and is an artifact of me copy&pasting it from another tool while I worked on it.  I'll update them in my next commit.

johnmwright:
I've updated the logic (and my description in "How I addressed the issue:") to be more accurate and handle more cases that were failing to parse the passed-in filters from VS (such as `ValueTuples` in the TestCase parameter signature, which were not being properly passed-in by VS).

johnmwright:
@johnjaylward Thanks for this! It's been fixed in my latest push.

> 
> It looks like some test cases may need to be updated though to match the changes:
> 
> ```
> Expected: "NUnit.Tests.FixtureWithTestCases.MethodWithParameters(9,11)"
>   But was:  "NUnit.Tests.FixtureWithTestCases.MethodWithParameters"
> ```

.gitignore
```diff
@@ -329,4 +329,6 @@ tools/
 
 # VisualStudioCode Patch
 # Ignore all local history of files
-.history
\ No newline at end of file
+.history
+/Dump
```
johnmwright: Ignoring the "dump" files created when troubleshooting builds using the [`DumpXml` utility](https://github.com/nunit/nunit3-vs-adapter/blob/bddee2c7739f9cc91a021aadf9b3a4f2ddba4c2c/src/NUnitTestAdapter/DumpXml.cs#L42).

johnmwright:
@peterwald thanks!  I realized I forgot to put a TL;DR at the top of the PR description, so I borrowed your "key highlights" for that purpose :)

.gitignore
OsirisTerje: Not sure why you need this one ?  The dump files [normally] goes below the bin folders, which is ignored.   No harm though.....

OsirisTerje:
@johnmwright  Thank you for this awesome work!   I believe this PR is the best one we have ever received.  It is straight to the point, and the documentation you have added is way above anything else I have seen!  

Not only is the documentation comprehensive and well structured,  the code simple and readable, tests have been added, but you have also taken the time to verify the resolution to a bunch of other similar issues!    

Really appreciate this work!

.gitignore
johnmwright: @OsirisTerje It creates one name `D_FileThatDoesntExist.dll.dump` in the top-level folder.  Not sure which test specifically created it, but it showed up in my git changes :)

PR #670, opened 2019-10-30
Update for net core

OsirisTerje:
Updated dumpxml to also work with .net core.
Added debug logging for the calling methods
Updated version to 3.16.0

### DIFF
--- a/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
+++ b/src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
@@ -61,7 +61,7 @@ namespace NUnit.VisualStudio.TestAdapter
             if (!Debugger.IsAttached)
                 Debugger.Launch();
 #endif
-            Initialize(discoveryContext,messageLogger);
+            Initialize(discoveryContext, messageLogger);
             TestLog.Info($"NUnit Adapter {AdapterVersion}: Test discovery starting");
 
             // Ensure any channels registered by other adapters are unregistered
@@ -83,17 +83,14 @@ namespace NUnit.VisualStudio.TestAdapter
                 if (Settings.DumpXmlTestDiscovery)
                 {
                     dumpXml = new DumpXml(sourceAssemblyPath);
-
                 }
 
                 try
                 {
                     runner = GetRunnerFor(sourceAssemblyPath, null);
-
-                    XmlNode topNode = runner.Explore(TestFilter.Empty);
-#if NET35
+                    var topNode = runner.Explore(TestFilter.Empty);
                     dumpXml?.AddString(topNode.AsString());
-#endif
+
                     // Currently, this will always be the case but it might change
                     if (topNode.Name == "test-run")
                         topNode = topNode.FirstChild;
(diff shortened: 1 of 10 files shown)


src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
```diff
@@ -61,10 +61,7 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
             if (!Debugger.IsAttached)
                 Debugger.Launch();
 #endif
-            Initialize(discoveryContext, messageLogger);
-
-
-
+            Initialize(discoveryContext,messageLogger);
```
jnm2: Normally there's a space after the comma. Do you think you'd like using an analyzer that gives a squiggle and Ctrl+. fix so that you can spend minimal energy on this stuff? (Or would you rather not care about stuff like this? I don't want to annoy you!)

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -322,9 +322,9 @@ private void RunAssembly(string assemblyPath, IGrouping<string, TestCase> testCa
             }
             finally
             {
-#if NET35
+//#if NET35
                 dumpXml?.Dump4Execution();
-#endif
+//#endif
```
jnm2: Do you think these lines will be needed in the future?

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
OsirisTerje: Shouldn't this be in the editorconfig file ?

OsirisTerje: It is the default, but there is no squiggly lines from that one, so yes, what analyzer do you suggest?

src/NUnitTestAdapter/NUnit3TestExecutor.cs
OsirisTerje: Gone and Done

src/NUnitTestAdapter/NUnit3TestExecutor.cs
```diff
@@ -79,21 +79,21 @@ public NUnit3TestExecutor()
         /// <param name="frameworkHandle">Test log to send results and messages through</param>
         public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrameworkHandle frameworkHandle)
         {
-#if LAUNCHDEBUGGER
+#if LAUNCHDEBUGGER 
             if (!Debugger.IsAttached)
                 Debugger.Launch();
 #endif
             Initialize(runContext, frameworkHandle);
+            TestLog.Debug("RunTests by IEnumerable<TestCase>");
```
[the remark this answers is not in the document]
OsirisTerje: THANKS!!!

src/NUnitTestAdapter/NUnit3TestDiscoverer.cs
jnm2: Yes, but it can only be set to true or false, not `true:error`. That means that it'll follow the preference when you type `;` at the end of the line and it autoformats, but it won't show a warning or a Ctrl+. fix option. What do you think about that?

https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#csharp_space_after_comma

OsirisTerje: Well, it works when you autoformat, but as you see, it (or I) doesn't always end up with the right result.  Squiggly lines are nice in those cases .

jnm2: I'd be happy to submit a PR adding an analyzer for this and you could see how you liked it. I use StyleCop and turn off at least half the rules. The remaining squiggles and code fixes make life easier for me.

OsirisTerje: Please do :-)

PR #673, opened 2019-10-31
Add StyleCop Analyzers

jnm2:
Followup from https://github.com/nunit/nunit3-vs-adapter/pull/670#discussion_r340717943.

@OsirisTerje This PR is in case you want to check out the branch locally and see if you like the experience of fixing things like missing spaces with `Ctrl+.`, `Enter`. If you don't like it, no worries. The "fix all in solution" option is so fast and easy that this didn't take much time.

If you do like the concept, you can skim the list of commit messages and let me know if there are any changes you *don't* prefer. This was just a guess. I can very easily drop those commits and suppress the rules. (My own opinion is that it should be up to your style. The goal is to make life easier.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -332,3 +332,82 @@ tools/
 .history
 /Dump
 /src/NUnitTestAdapterTests/Dump
+*.rsuser
+# Mono auto generated files
+mono_crash.*
+[Aa][Rr][Mm]/
+[Aa][Rr][Mm]64/
+[Ll]ogs/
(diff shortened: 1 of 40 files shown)


OsirisTerje:
Thanks @jnm2 !   I'll go through them,  right now I see 1614 and 1616 which I think will give too many warnings where I dont think we will do them.   Most of the rest seems ok, but give me a couple of days :-)

Issue #679, opened 2019-12-01
build.cake does not work in a pure visual studio preview installation


ovebastiansen:
I am trying to debug the adapter, and I am not able to build the nuget package in my setup. I have only Visual Studio 2019 preview Community installed.

PR coming with a fix

OsirisTerje:
Thanks !

PR #680, opened 2019-12-01
finding msbuild with a pure vs preview installation

ovebastiansen:
getting a path for msbuild in a pure vs preview installation now works
#679


OsirisTerje:
LGTM

PR #717, opened 2020-02-01
Directory.DeleteDirectory(recursive: true) is non-atomic

jnm2:
Re: https://nunit.visualstudio.com/NUnit/_build/results?buildId=2517&view=ms.vss-test-web.build-test-results-tab

I don't have any idea what causes this problem in general. Sometimes it's antivirus or Windows Explorer (obviously not in the case of CI). Maybe a file was locked?

I typically end up using this helper eventually anywhere I use Directory.DeleteDirectory. Another approach we could take is to make it warn if it fails since it does not indicate an error in the SUT, just leftovers in %TEMP% that weren't cleaned up after the test succeeded.

### DIFF
--- a/src/NUnitTestAdapterTests/TempDirectory.cs
+++ b/src/NUnitTestAdapterTests/TempDirectory.cs
@@ -39,7 +39,7 @@ namespace NUnit.VisualStudio.TestAdapter.Tests
 
         public void Dispose()
         {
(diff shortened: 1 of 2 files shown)


OsirisTerje:
Looks like the github build is not starting ?

jnm2:
> Looks like the github build is not starting ?

Maybe because this was a branch in my fork?

PR #741, opened 2020-04-24
Issue740

OsirisTerje:


src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -368,6 +358,23 @@ string ValidatedPath(string path, string purpose)
             }
         }
 
+        private void MapTestCategory(string vsTestCategoryType)
+        {
+            switch (vsTestCategoryType.ToLower())
```
jnm2, line 363: The idiomatic thing to do is `string.Equals(x, y, StringComparer.OrdinalIgnoreCase)`. ToLower has issues with certain culture settings and it produces an unnecessary string allocation.

jnm2: Recommended:

```cs
if (Enum.TryParse<VsTestCategoryType>(vsTestCategoryType, ignoreCase: true, out var value))
{
    VsTestCategoryType = value;
}
else
{
    _logger.Warning(
        $"Invalid value ({vsTestCategoryType}) for VsTestCategoryType. Valid values are NUnit and MSTest.");
}
```

src/NUnitTestAdapter/AdapterSettings.cs
```diff
@@ -488,6 +495,28 @@ private void Log<T>(string xpath, T res)
...
+                "passed" => TestOutcome.Passed,
+                "none" => TestOutcome.None,
+                _ => TestOutcome.NotFound,
+            };
+            if (testOutcome == TestOutcome.NotFound)
+            {
+                _logger.Warning(
+                    $"Invalid value ({outcome}) for MapWarningTo, should be either Skipped,Failed,Passed or None");
+                return TestOutcome.Skipped;
+            }
```
jnm2, line 516: Same recommendation, Enum.TryParse with ignoreCase: true.

src/NUnitTestAdapter/AdapterSettings.cs
OsirisTerje: Thanks!

src/NUnitTestAdapter/AdapterSettings.cs
OsirisTerje: And thanks again!

src/NUnitTestAdapter/AdapterSettings.cs
OsirisTerje: We're still on .net 3.5, where this doesn't exist.-......   We really need to get off 3.5

jnm2: Okay, now ToLower is still not equivalent to StringComparer.OrdinalIgnoreCase. `"NUNIT".ToLower()` is not `"nunit"` if the user's locale is set to Turkish.

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
```diff
@@ -199,8 +189,25 @@ public NUnitTestEventTestCase(XmlNode node) : base(node)
 
         public string ReasonMessage { get; }
 
-        public bool HasReason => string.IsNullOrEmpty(ReasonMessage);
+        public bool HasReason => !string.IsNullOrEmpty(ReasonMessage);
```
jnm2, line 192: Does this indicate a test hole?

src/NUnitTestAdapter/AdapterSettings.cs
jnm2: Options I'd recommend:

- Add a utility `TryParseEnum<T>` method by catching ArgumentException from Enum.Parse (I do this on .NET 3.5)

- Convert `switch` to `if` and use `nameof(VsTestCategoryType.NUnit).Equals(vsTestCategoryType, StringComparison.OrdinalIgnoreCase)`

- Use a private static readonly `new Dictionary<string, T>(StringComparer.OrdinalIgnoreCase)` filled at runtime from the enum

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
```diff
@@ -199,8 +189,25 @@ public NUnitTestEventTestCase(XmlNode node) : base(node)
...
         public bool HasFailure => Failure != null;
+
+        /// <summary>
+        /// Find stacktrace in assertion nodes if not defined (seems .netcore2.0 doesn't provide stack-trace for Assert.Fail("abc"))
```
jnm2, line 196: > seems .netcore2.0 doesn't provide stack-trace for Assert.Fail("abc")

This sounds like a potentially fixable framework bug.

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
```diff
@@ -199,8 +189,25 @@ public NUnitTestEventTestCase(XmlNode node) : base(node)
...
+        /// </summary>
+        public string StackTrace
+        {
+            get
+            {
+                string stackTrace = string.Empty;
+                foreach (XmlNode assertionStacktraceNode in Node.SelectNodes("assertions/assertion/stack-trace"))
+                {
+                    stackTrace += assertionStacktraceNode.InnerText;
+                }
```
jnm2, line 206: If there are multiple assertions, we should show separate stack traces rather than concatenating them as though there was one call inside all the other calls.

src/NUnitTestAdapter/NUnitEventListener.cs
```diff
@@ -163,7 +163,16 @@ public void TestFinished(NUnitTestEventTestCase resultNode)
...
             var result = _testConverter.GetVsTestResults(resultNode, outputNodes ?? EmptyNodes);
             if (_settings.ConsoleOut == 1 && !string.IsNullOrEmpty(result.ConsoleOutput))
+            {
                 _recorder.SendMessage(TestMessageLevel.Informational, result.ConsoleOutput);
+            }
+            if (_settings.ConsoleOut == 1 && !string.IsNullOrEmpty(resultNode.ReasonMessage))
+            {
+                var msg = resultNode.ReasonMessage;
+                var msg2 = $"{resultNode.Name}: {msg}";
+                _recorder.SendMessage(TestMessageLevel.Informational, msg2);
```
jnm2, line 173: Consider avoiding variable names suffixed with '2' (can we inline both the variables?)

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -112,26 +112,29 @@ public TestResultSet GetVsTestResults(NUnitTestEventTestCase resultNode, ICollec
...
-                        string stackTrace = string.Empty;
-                        foreach (XmlNode assertionStacktraceNode in resultNode.Node.SelectNodes("assertions/assertion/stack-trace"))
-                        {
-                            stackTrace += assertionStacktraceNode.InnerText;
-                        }
-                        testCaseResult.ErrorStackTrace = stackTrace;
+                        if (adapterSettings.ConsoleOut > 0 && !string.IsNullOrEmpty(resultNode.ReasonMessage))
+                            testCaseResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, resultNode.ReasonMessage));
+                        break;
                     }
```
jnm2, line 129: Is there a test that fails if this change is reverted?

src/NUnitTestAdapter/TestConverter.cs
```diff
@@ -112,26 +112,29 @@ public TestResultSet GetVsTestResults(NUnitTestEventTestCase resultNode, ICollec
...
+                        break;
+                    default:
                     {
-                        string stackTrace = string.Empty;
-                        foreach (XmlNode assertionStacktraceNode in resultNode.Node.SelectNodes("assertions/assertion/stack-trace"))
-                        {
-                            stackTrace += assertionStacktraceNode.InnerText;
-                        }
-                        testCaseResult.ErrorStackTrace = stackTrace;
+                        if (adapterSettings.ConsoleOut > 0 && !string.IsNullOrEmpty(resultNode.ReasonMessage))
```
jnm2, line 133: What do you think of making `adapterSettings.ConsoleOut` an enum in the future?

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
OsirisTerje: Yes :-(

src/NUnitTestAdapter/TestConverter.cs
OsirisTerje: Actually.....I have forgotten why it even is an integer.... it could perhaps just be a boolean.

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
OsirisTerje: Done

src/NUnitTestAdapter/AdapterSettings.cs
OsirisTerje: Done, but seems a waste - we should get over to 4.X asap

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
OsirisTerje: Should it be raised in the NUnit repo?

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
OsirisTerje: I am unsure how the engine creates these nodes, and what actually they represent.  Also need to figure out if we have that option in the Test Explorer interface.

src/NUnitTestAdapter/NUnitEventListener.cs
OsirisTerje: Done

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
jnm2: Yes please, if you know how to reproduce it.

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
jnm2: I believe this is tied to the multiple assertions feature in the framework:

```cs
Assert.Multiple(() =>
{
    Assert.Fail(); // Here's one stack trace

    Assert.That(1, Is.EqualTo(2)); // Here's another stack trace.
});
```

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
OsirisTerje: I can't repro with netcore 2.1.  And, with net core 2.0 I can't even execute the tests:  1>D:\repos\NUnit\FrameworkIssues\NetCore20NoStackTrace\NetCore20NoStackTrace.csproj : warning NU1701: Package 'NUnit3TestAdapter 3.16.1' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.

Have we lost the netcore 2.0 compatibility ?

OsirisTerje: If I try with earlier versions (than 3-.16.1 of the adapter) it says:  "Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Unable to find D:\repos\NUnit\FrameworkIssues\NetCore20NoStackTrace\bin\Debug\netcoreapp2.0\testhost.dll. Please publish your test project and retry."

jnm2: .NET Core 2.1 and 3.1 are the oldest two .NET Core *runtime* versions still supported by Microsoft. We only support back to 2.1:

https://github.com/nunit/nunit3-vs-adapter/blob/26aa1f0304a75fe2937cda95b4ebe7eb3ef19888/src/NUnitTestAdapter/NUnit.TestAdapter.csproj#L8

If 2.1 doesn't repro, then I'm not worried about it.

Test projects should have no reason to target 2.0 rather than 2.1 because 2.0 is a deprecated runtime.

OsirisTerje: Ok, the adapter is built with netcore2.1.  I assumed that could run 2.0, but obviously not.  No one has reported anything about this, so should we care?

jnm2: This is the flip side of not having `developmentDependency="true"`. I still think not having it is the right thing to do. It means the target framework and version now must match the project that it's installed into.

I don't think we should care. If someone asks, we say, "Move to 2.1 because not even Microsoft supports 2.0."

OsirisTerje: And trying a .net framework 3.5 gives this: "3>d:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2081,5): warning MSB3268: The primary reference "C:\Users\TerjeSandstrom\.nuget\packages\nunit\3.12.0\lib\net35\nunit.framework.dll" could not be resolved because it has an indirect dependency on the framework assembly "System.Web, "

jnm2: That's because the adapter requires System.Web for one interface which NUnit Engine uses. The answer is to make the test project target the normal full `net35`.

OsirisTerje: Version 3.16 stopped supporting net core 1.0 **and** net core 2.0.    Version 3.15.1 was the latest with support for those. And it needs the Test.Sdk 15, doesn't work with 16.  And - the stacktrace is there... so I think the comment is bogus....or outdated....

jnm2: Cool, then I'll mark this comment thread as resolved.

src/NUnitTestAdapter/TestConverter.cs
OsirisTerje: The code has been moved, not changed.....  There are 9 tests covering this area... I'll check if they handle all of this.

src/NUnitTestAdapter/NUnitEngine/NUnitTestEvent.cs
OsirisTerje: We're currently returning this a single test resultset, and there we only have a single stack trace property.  Believe we need to look more into how to represent this on the Test Explorer side.

src/NUnitTestAdapter/TestConverter.cs
jnm2: Ah, sorry, missed where the code in the `default` section came from.

PR #754, opened 2020-06-03
Update .editorconfig to support *.csproj

### LINKED ISSUE #753: EditorConfig for *.csproj Types
Similar to #746 when https://github.com/nunit/nunit3-vs-adapter/commit/87e0087daf8c35535c5c1399c45ce752c1d9934c occurred the configuration settings for `*.csproj` were reverted.

This should probably be put back in place, otherwise the next person to come in and run format on the csproj files (such as when upgrading the Engine as I did on PR #752 ) are going to run into needless diffs in formatting.

There was at least one other file type on that commit (`*.vsixmanifest`) however I do not believe that we ever edit that file, if we think we do I don't mind changing the PR and the issue to reflect all types.

aolszowka:
Closes #753 

We need to change CSPROJ Files when we add or remove dependencies, the .editorconfig should ensure that we have these files remain consistent

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,10 @@ root = true
 indent_style = space
 indent_size = 4
 
+[*.csproj]
+indent_style = space
+indent_size = 2
+
 [*.props]
 indent_style = space
 indent_size = 2
--- a/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
+++ b/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
@@ -4,7 +4,8 @@
     <AssemblyName>NUnit3.TestAdapter</AssemblyName>
     <!--<AssemblyName>NUnit3.TestAdapterDebug</AssemblyName>-->
     <RootNamespace>NUnit.VisualStudio.TestAdapter</RootNamespace>
-    <!--<TargetFramework>netcoreapp2.1</TargetFramework>--><!-- For testing and debugging-->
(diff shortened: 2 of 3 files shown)


aolszowka:
I ran the formatter on all *.csproj files, and we do get some changes, but I am not sure we want to apply them.

Specifically in `src\NUnit.Test.Adapter.csproj` and `src\NUnit.Test.Adapter.Tests.csproj` have this:

```xml
    <!--<TargetFramework>netcoreapp2.1</TargetFramework>--><!-- For testing and debugging-->
```

Running Visual Studios' `Format Document` command wants to put these on separate lines. To me I could go either way on if we care that these comments are on the same line, its more preference of the developer, I would defer to any one else's opinion on it.

OsirisTerje:
@aolszowka  This is rather harmless, so just go through the project files and update them.

OsirisTerje:
LGTM

aolszowka:
I am sorry for all the "noise" on this commit I am still trying to learn how to better "squash commits" is there a guide somewhere that can walk me though step by step?

I have tried git rebase master which does what I would expect it to do, but I am unable to commit this.

In my mind there should only show 1 commit with the 3 files changed, but I am still missing something...

OsirisTerje:
@aolszowka Don't worry about doing it manually, it is handled by github when I do the merge, see screenshot below :-)
image

aolszowka:
@OsirisTerje That is HUGE I did not realize it did that, thank you I will change off of WIP.

I committed the changes to the affected CSPROJ files as well.

PR #790, opened 2020-10-09
Remove unnecessary package dependencies.

### LINKED ISSUE #785: Seemingly redundant dependency on Microsoft.DotNet.InternalAbstractions
As of 3.17.0 there's a dependency on Microsoft.DotNet.InternalAbstractions package in .nuspec but it's unclear how it is useful - I cannot find any code in this codebase which would depend on Microsoft.DotNet.InternalAbstractions.

The problem is Microsoft.DotNet.InternalAbstractions is published in Nuget https://www.nuget.org/packages/Microsoft.DotNet.InternalAbstractions/ without mentioning any license. This causes a great deal of trouble when auditing the codebase which uses code from this codebase in a larger project - the dependency on Microsoft.DotNet.InternalAbstractions raises a lot of red flags because it is "unlicensed" and so this component raises a red flag too because it relies on an "unlicensed" component.

Since it looks like the dependency is redundant it'd be nice if it was removed and a new version of this component was published in Nuget.

teo-tsirpanis:
Fixes #785.
These dependencies are for .NET Standard 1.x dependencies and for .NET Core 2.1 they just bloat the package dependency stream for no reason.
And use an SPDX license expression to identify the license as MIT.

I would especially appreciate if this PR was marked with a `hacktoberfest-accepted` label.

### DIFF
--- a/nuget/NUnit3TestAdapter.nuspec
+++ b/nuget/NUnit3TestAdapter.nuspec
@@ -5,7 +5,7 @@
         <version>$version$</version>
         <title>NUnit 3 Test Adapter for Visual Studio and DotNet</title>
         <authors>Charlie Poole, Terje Sandstrom</authors>
-        <license type="file">LICENSE.txt</license>
+        <license type="expression">MIT</license>
         <projectUrl>https://github.com/nunit/docs/wiki/Visual-Studio-Test-Adapter</projectUrl>
         <repository type="git" url="https://github.com/nunit/nunit3-vs-adapter"/>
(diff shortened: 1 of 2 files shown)


OsirisTerje:
Hi @teo-tsirpanis !  
1. You're doing two unrelated things in this PR.  Not critical in this case, since they are both simple, but they are unrelated.  
2. A PR should refer to an issue.  In your case the first fix actually has an issue, but it is addressing just partially what you fix.  See #785

PR #714, opened 2020-01-25
Add Linux CI job

jnm2:

PR #803, opened 2020-11-22
Improve first-time experience opening the solution

jnm2:
When you open the solution without having a .NET Core SDK 3.1.x installed, Visual Studio shows warnings for all projects and can't build anything. This is due to the global.json file I added in https://github.com/nunit/nunit3-vs-adapter/pull/714/commits/85f09e771e474c46bfc86c129e73952ebbfa88ad which should have had `"rollforward": "major"` to allow working when you don't have 3.1.x but you do have 5.0.x. Rather than fixing this, I just deleted the global.json file because the CI images are updated to the point where they already have the right things installed.

This also cleans up all warnings in the error list and adds a CI check that prevents them from accumulating again.


src/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\NUnitAdapter.snk</AssemblyOriginatorKeyFile>
     <CodeAnalysisRuleSet>..\..\Osiris.Extended.ruleset</CodeAnalysisRuleSet>
+    <TreatWarningsAsErrors Condition="'$(Configuration)' == 'Release'">true</TreatWarningsAsErrors>
```
jnm2, line 8: The reason this is conditional is that it's too annoying to be in the middle of writing something in the IDE and try to run or debug a test and get a build failure over something that is only a warning. Often there is an unused variable or incomplete formatting. These things shouldn't make it into the master branch, but they also definitely shouldn't slow down inner-loop development.

Issue #811, opened 2021-01-11
"System.FormatException: The UTC representation of the date falls outside the year range 1-9999" from skipped test in Eastern European time zone


KalleOlaviNiemitalo:
I have a C# test project in which a test fixture class has NUnit.Framework.ExplicitAttribute. The project references NUnit3TestAdapter and NUnit from NuGet. I then run `vstest.console.exe "REDACTED.dll" /Parallel /Framework:".NETFramework,Version=4.5" /Platform:x64 /Logger:trx;LogFileName=".NETFramework,v4.5.trx" /ResultsDirectory:"Test\TestResults"` and get a FormatException for each test that was skipped because of the ExplicitAttribute:

```
Microsoft (R) Test Execution Command Line Tool Version 15.9.1
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
NUnit Adapter 3.17.0.0: Test execution started
Running all tests in REDACTED.dll
   NUnit3TestExecutor discovered 13 of 13 NUnit test cases

Error processing REDACTED event for REDACTED
System.FormatException: The UTC representation of the date falls outside the year range 1-9999.
   at System.DateTimeParse.Parse(String s, DateTimeFormatInfo dtfi, DateTimeStyles styles, TimeSpan& offset)
   at System.DateTimeOffset.Parse(String input, IFormatProvider formatProvider, DateTimeStyles styles)
   at NUnit.VisualStudio.TestAdapter.NUnitEngine.NUnitTestEvent.StartTime() in D:\repos\NUnit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnitEngine\NUnitTestEvent.cs:line 135
   at NUnit.VisualStudio.TestAdapter.TestConverter.GetBasicResult(INUnitTestEvent resultNode, IEnumerable`1 outputNodes) in D:\repos\NUnit\nunit3-vs-adapter\src\NUnitTestAdapter\TestConverter.cs:line 302
   at NUnit.VisualStudio.TestAdapter.TestConverter.GetVsTestResults(INUnitTestEventTestCase resultNode, ICollection`1 outputNodes) in D:\repos\NUnit\nunit3-vs-adapter\src\NUnitTestAdapter\TestConverter.cs:line 111
[...]

PR #812, opened 2021-01-12
Issue #811 , adapter fix to make it more resistant

OsirisTerje:
This fix, as suggested by @KalleOlaviNiemitalo  in #811 , will make the adapter more resilient to the current explicit time information coming from version 3.13 of the NUnit framework.  See issue for more details


OsirisTerje:
@KalleOlaviNiemitalo I don't see exactly how we ensure this now.  We might use Azure data centers at different locations, but I am not sure how they are handling their local time.   Note, however, we are open to receiving pull requests so if you have a suggestion, feel free to implement it :-)

PR #821, opened 2021-03-05
Optimize ProcessTestCaseProperties

pakrym:
The IsInternalProperty is on a very hot path:

image

This PR devirtualizes the call to `ShowInternalProperties` and replaces a list lookup with a hashset lookup.


OsirisTerje:
LGTM :-)

PR #845, opened 2021-04-13
[ProcessUtils] Simplify reading the process sdtout and stderr.

### LINKED ISSUE #844: [ProcessUtils] Simplify reading the process sdtout and stderr.
Not a bug more like a little optimization request, I looked at the code at https://github.com/nunit/nunit3-vs-adapter/blob/9b38318f46aac06326c455e17efc205741ca3482/src/NUnit.TestAdapter.Tests.Acceptance/WorkspaceTools/ProcessUtils.cs#L38-L68 which @jnm2 pointed out to me, I then realized the code could be simplified further yet still do the same thing (just with a few less lines of code to maintain).

I have successfully commited the change to a fork, I will open a draft pr for it too for review as well.

AraHaan:
@jnm2 pointed me to this code where later I realized in reading it that it could be simplified further so then the events do not need to allocate the actual objects.

The result is the exact same as where if an event is never executed (the process never outputs to stdout or stderr then the builder's ``ToString()`` method will return string.Empty.

I do not think this will be a breaking change since most people check if strings are null or empty using ``string.IsNullOrEmpty()`` anyway so.

Fixes #844.

### DIFF
--- a/src/NUnit.TestAdapter.Tests.Acceptance/WorkspaceTools/ProcessUtils.cs
+++ b/src/NUnit.TestAdapter.Tests.Acceptance/WorkspaceTools/ProcessUtils.cs
@@ -45,7 +45,9 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Acceptance.WorkspaceTools
                 {
                     if (e.Data is null) return;
 
-                    stdout.AppendLine();
+                    if (stdout.Length != 0)
+                        stdout.AppendLine();
+
                     stdout.Append(e.Data);
                 };
 
@@ -53,7 +55,9 @@ namespace NUnit.VisualStudio.TestAdapter.Tests.Acceptance.WorkspaceTools
                 {
                     if (e.Data is null) return;
 
-                    stderr.AppendLine();
+                    if (stderr.Length != 0)
+                        stderr.AppendLine();
+
                     stderr.Append(e.Data);
                 };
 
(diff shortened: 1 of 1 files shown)


OsirisTerje:
Non-draft is ok, but perhaps good if the acceptance tests are green :-)

jnm2 (commented):
This undoes the optimization I put in place to avoid allocating a string builder only to not use it. It looks like it also changes the behavior by inserting an extra newline at the beginning of each.

AraHaan:
Fixed that using the ``StringBuilder.Length`` property 😄.

jnm2:
@AraHaan I still wonder why it is good to make the overall change. I'm not opposed to it depending on what we're gaining, and I appreciate the enthusiasm, but what is the gain?

AraHaan:
Well for starters it's more readable, plus the way I read it it looked like the allocation would be made regardless, plus how much memory could an empty StringBuilder (provided there was no output or error outputs) or even made a big issue?

(LOL now the CI passes)

jnm2:
I can say for sure that the allocation wasn't happening regardless; if no stderr was produced, I don't think a StringBuilder was allocated for it; same with stdout. I would be very surprised if the JIT was capable of avoiding the allocation. Since constructors can have side effects, it would not be a valid optimization to skip newing these classes up unless the JIT was giving StringBuilder special treatment in this way which would also surprise me. Like I said though, I'm okay with not worrying about the StringBuilder allocation once per process run. That seems negligible.

I don't quite see what you mean about readability and maintainability. It's not a thing I would personally do but either way seems great to me. @OsirisTerje is the maintainer.

OsirisTerje:
Looking at this, I get confused.  Old Lines 74-75 are much more readable than new lines 70-71.   I would not use a typecast on old lines 38-39, but just called it out explicitly,  like 
```
StringBuilder stdout;
StringBuilder stderr;
```
which both would be null by default.

Checking length or checking the existence - IDK.....

jnm2:
On one hand I don't like turning down enthusiastic help because the desire is very much appreciated. On the other hand, my honest opinion on this subjective matter still leads me back to where I ended up when I originally weighed these things, so if it makes no difference to Terje, I'd come down on the side of keeping it as-is.

PR #920, opened 2021-11-26
Removing the vsix project

OsirisTerje:
Remove the vsix project, as the 4.x versions will not support vsix anymore.  The vsix is being deprecated by Microsoft for Visual Studio, and developers should move to use the nuget adapters instead.


Tampa:
For how deprecated it is meant to be its weirdly alive and well that whole ecosystem. It's almost as if nuget is a total mess rivaling that of npm. vsix still has a bunch of things in 2022 including articles on how to keep things updated, almost as if Microsoft says a lot of things in a day, most of which is complete bollocks.

-1 on this change, vsix has it's uses.

OsirisTerje:
Vsix for unit testing in Visual Studio is what is being deprecated.  Not vsix in general. 

Up to VS2022 it was easy to support vsix too, so it just went along with the rest, but since VS2022 is a x64 app, the vsix also have to change.  How much work that is, we haven't investigated, but you're free to do a try.  The source code is open and free under the MIT license, and you can build your own vsix if you like too.

PR #966, opened 2022-03-24
Issue 965: Special build using pre-release of nunit.engine 4.0.0

### LINKED ISSUE #965: Create (beta) version that can be used in NUnit4 development
Now that the engine is tested to work with nunit4 (https://github.com/nunit/nunit-console/issues/1139) can we create a pre-release version of the adapter using nunit.engine 4, which can be referenced in the nunit framework so we can test it again from within Visual Studio.
This would then fix https://github.com/nunit/nunit/issues/3867

manfred-brands:
Fixes #965 

I left this as a draft PR as I don't expect it to be merged into master.
The purpose is to get a special build that can be referenced in nunit framework nunit4 developments.

As I also not aware of the changes in the engine, I have deleted several lines in `TestAdapterUtils` as those classes no longer exist.

There was one other change. The name/fullname properties of the test-run node seem to no longer exist.

I testes the nuget package with nunit framework and it allows me to run tests in VS and with `dotnet test`

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -12,7 +12,7 @@ var configuration = Argument("configuration", "Release");
 // SET PACKAGE VERSION
 //////////////////////////////////////////////////////////////////////
 
-var version = "4.2.1";
+var version = "5.0.0";
 var modifier = "";
 
 var dbgSuffix = configuration.ToLower() == "debug" ? "-dbg" : "";
@@ -101,7 +101,7 @@ Task("Clean")
 Task("CleanPackages")
     .Does(()=>
     {
-    CleanDirectory(PACKAGE_DIR);
+        CleanDirectory(PACKAGE_DIR);
     });
 
 //////////////////////////////////////////////////////////////////////
--- a/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
+++ b/src/NUnitTestAdapter/NUnit.TestAdapter.csproj
@@ -22,9 +22,17 @@
     <DisableHandlePackageFileConflicts>false</DisableHandlePackageFileConflicts>
   </PropertyGroup>
 
+  <!-- We reference a specific pre-release version of the nunit.engine -->
+  <PropertyGroup>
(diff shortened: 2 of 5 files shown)


manfred-brands:
Not sure if you agree with the version number, but the adapter seems to be one ahead of the engine/framework.

manfred-brands:
The build seem to not be able to find the myget package when referenced in .csproj file. Trying to add to nuget.config

manfred-brands:
@CharliePoole The engine does find the test for netcore/net6, but not for net framework.
In the debugger is shows:
```
System.ComponentModel.Win32Exception : The system cannot find the file specified--Win32Exception
The system cannot find the file specified
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at NUnit.Engine.Services.TestAgency.GetAgent(TestPackage package)
   at NUnit.Engine.Runners.ProcessRunner.CreateAgentAndRunnerIfNeeded()
   at NUnit.Engine.Runners.ProcessRunner.Explore(TestFilter filter)
```
It seems to be looking for nunit-agent.exe binaries. I see them in the contentFiles directory.

@OsirisTerje build.cake --target=Test shows all green lines, even though it didn't find any tests.
```
No test is available in D:\3rd\NUnit\nunit3-vs-adapter\src\NUnitTestAdapterTests\bin\Debug\net46\NUnit.VisualStudio.TestAdapter.Tests.dll. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.
```

OsirisTerje (changes requested):
This is not a major version change, so 4.3.0 is sufficient.  Also, add pre-release for now, so e.g. 4.3.0-beta.1 should do it.   And you're right, the version numbers are **not** aligned across components, and that is "By design".  :-)

CharliePoole:
@manfred-brands My guess is that you are running into nunit/nunit-console#1023, which is a pretty big one!

I just assigned you an  issue, which is a small part of a bigger part of that issue. :-)

OsirisTerje:
I assume we don't need this anymore, as 4.5 does run nunit 4.

Issue #1021, opened 2022-10-20
update root folder links


Arhell:
http=>https

OsirisTerje:
Hi!   There is very little value in raising issues of this kind.  Would really appreciate if you just added a few more words around this, or, if you just fix a few things in the files, like the last PR, just skip the issue all together and just do the PR. :-)

OsirisTerje:
Thanks for the 1st PR!  And for the second, you could just do the delete of that file there - that would be ok too :-)

PR #1022, opened 2022-10-21
update root folder links

Arhell:
resolve https://github.com/nunit/nunit3-vs-adapter/issues/1021
cc @OsirisTerje
all links chacked


OsirisTerje:
This is a generated ndproj file from NDepend, used for an analysis a while back.  It should not be manually changed, but regenerated with a newer version of NDepend.  This version can just be deleted.

PR #1112, opened 2023-08-03
feat: enable custom test output result file

Lorilatschki:
When using the ``TestOutputXmlFileNameWithoutExtension`` run setting property, the client can define a custom result file name rather than the name of the assembly to be tested

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -119,6 +119,7 @@ namespace NUnit.VisualStudio.TestAdapter
         public string WorkDirectory { get; private set; }
         public string Where { get; private set; }
         public string TestOutputXml { get; private set; }
+        public string TestOutputXmlFileNameWithoutExtension { get; private set; }
         public bool UseTestOutputXml => !string.IsNullOrEmpty(TestOutputXml) || OutputXmlFolderMode == OutputXmlFolderMode.UseResultDirectory;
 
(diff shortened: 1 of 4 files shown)


Lorilatschki:
@dotnet-policy-service agree

Lorilatschki:
@OsirisTerje this will enable the custom result file definition. The docs need to be adapted due to the newly introduced property ``TestOutputXmlFileNameWithoutExtension``. After merging and adapting the docs, the [docs/issue/592](https://github.com/nunit/docs/issues/592) can be closed.

OsirisTerje:
Looks good, but `TestOutputXmlFileNameWithoutExtension` is a very long propertyname.  I think the `WithoutExtension` can be skipped.  It can be stated in the docs, and if they don't read it the worst that will happen is they will have a file named `whatever.xml.xml`.

OsirisTerje:
Moved the docs item over here, since it is adapter docs, at #1114

Lorilatschki:
> Looks good, but `TestOutputXmlFileNameWithoutExtension` is a very long propertyname. I think the `WithoutExtension` can be skipped. It can be stated in the docs, and if they don't read it the worst that will happen is they will have a file named `whatever.xml.xml`.

Will adjust it.

OsirisTerje (commented):
lgtm

OsirisTerje (approved):

OsirisTerje:
Fixes ##1114

OsirisTerje:
@Lorilatschki Fixed in 4.6,  beta here :  https://www.myget.org/feed/nunit/package/nuget/NUnit3TestAdapter

PR #1153, opened 2024-02-20
Initial support of Microsoft.Testing.Platform

### LINKED ISSUE #1152: Microsoft Testing Platform for NUnit
Dear NUnit team,

As per our previous discussions, we would like to help you onboard NUnit to the new self-contained testing platform that powers MSTest Runner.

We are currently working on the technical documentation for framework authors and extensions authors so sadly for now there isn't much available for you to browse.

The new platform is already OSS and available inside of MSTest repository (at the moment).

Related issue about support NUnit on MSTest repo: https://github.com/microsoft/testfx/issues/2164

MSTest runner blogpost announcement: https://devblogs.microsoft.com/dotnet/introducing-ms-test-runner/
MSTest runner doc: https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-runner-intro?tabs=dotnetcli
Platform code: https://github.com/microsoft/testfx/tree/main/src/Platform/Microsoft.Testing.Platform
Sample code: https://github.com/microsoft/testfx/tree/main/samples/mstest-runner

Evangelink:
Provide the base changes to support the new platform while keeping backward compatibility.

There are multiple points to discuss for the design, naming and update of your testing infra.

Fixes #1152


nuget.config
```diff
@@ -3,6 +3,6 @@
   <packageSources>
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
-    <!--<add key="Local" value="c:\nuget" />-->
+    <add key="Local" value="C:\src\nunit3-vs-adapter\package" />
```
Evangelink, line 6: Should not be kept, this is to allow working with the samples.

nuget/NUnit3TestAdapter.nuspec
```diff
@@ -24,6 +24,19 @@
     <tags>test visualstudio testadapter nunit nunit3 dotnet</tags>
 
     <developmentDependency>false</developmentDependency>
+
+    <dependencies>
+      <group targetFramework="net462">
+        <dependency id="Microsoft.Testing.Extensions.VSTestBridge" version="1.0.1" />
```
Evangelink, line 30: We recommend having explicit deps to the bridge rather than repackaging the dlls/props/targets... to ease and standardize versioning/update conflicts.

nuget/NUnit3TestAdapter.nuspec
```diff
@@ -24,6 +24,19 @@
...
+        <dependency id="Microsoft.Testing.Extensions.Telemetry" version="1.0.1" />
```
Evangelink, line 31: This is not a requirement but we wanted to avoid having telemetry on the core platform as it could be causing issues and would be leading to extra deps on the platform that we want to keep as light and no deps as possible.

nuget/NUnit3TestAdapter.nuspec
```diff
@@ -24,6 +24,19 @@
...
+        <dependency id="Microsoft.Testing.Platform.MSBuild" version="1.0.1" />
```
Evangelink, line 32: This package contains the main logic for handling MSBuild: producing the entry point, registering and loading extensions... This is not required but without it, the user would have to manually provide the entry point.

This is also handling the `dotnet test` logic and many more MSBuild related features.

nuget/net462/NUnit3TestAdapter.props
```diff
@@ -1,11 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <EnableNUnitRunner Condition=" '$(EnableNUnitRunner)' == '' ">false</EnableNUnitRunner>
```
Evangelink, line 5: Feel free to change to any name you want and to decide if you want to have it enabled by default. On MSTest side we decided to keep it disabled by default for now to avoid breaking changes and have incremental opting-in wave of users. We do plan to make it default in v4.

nuget/net462/NUnit3TestAdapter.props
```diff
@@ -1,11 +1,24 @@
...
+    <IsTestingPlatformApplication>$(EnableNUnitRunner)</IsTestingPlatformApplication>
```
Evangelink, line 6: Some equivalent to the "TestContainer" capability for the new platform.

nuget/net462/NUnit3TestAdapter.props
```diff
@@ -1,11 +1,24 @@
...
+  <ItemGroup>
+    <!--
+      !!! IMPORTANT !!!
+      DO NOT CHANGE THE GUID, IT'S A WELL KNOWN EXTENSION POINT AND THIS EXTENSION NEEDS TO BE REGISTERED AT THE END
+      WE HAVE CODE INSIDE THE TASK 'TestingPlatformEntryPoint' TO ENSURE THE ORDER OF THE REGISTRATION BASED ON THIS GUID
+    -->
+    <TestingPlatformBuilderHook Include="2E8E7F63-61DB-4EDB-A21E-5BF48279A7B8" Condition=" $(GenerateTestingPlatformEntryPoint) == 'true' " >
+      <DisplayName>NUnit</DisplayName>
+      <TypeFullName>NUnit.VisualStudio.TestAdapter.TestingPlatformAdapter.TestingPlatformBuilderHook</TypeFullName>
+    </TestingPlatformBuilderHook>
```
Evangelink, line 18: Register the hook called by `Microsoft.Testing.Platform.MSBuild` when auto-generating the entry point.

nuget/net462/NUnit3TestAdapter.props
```diff
@@ -1,11 +1,24 @@
...
+  </ItemGroup>
+    
   <ItemGroup>
-    <None Include="$(MSBuildThisFileDirectory)NUnit3.TestAdapter.dll">
```
Evangelink, line 4: Moved to the target, see there for explanation.

nuget/net462/NUnit3TestAdapter.targets
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Handle the coexistance between Microsoft.Testing.Platform and Microsoft.NET.Test.Sdk  -->
+  <PropertyGroup>
+    <GenerateTestingPlatformEntryPoint Condition=" '$(GenerateTestingPlatformEntryPoint)' == '' ">$(EnableNUnitRunner)</GenerateTestingPlatformEntryPoint>
+    <GenerateProgramFile Condition=" '$(EnableNUnitRunner)' == 'true' ">false</GenerateProgramFile>
```
Evangelink, line 7: The first line enables/disables the generation of the entry point for the new platform and the second line disables the VSTest entry point if the new platform is enabled.

nuget/net462/NUnit3TestAdapter.targets
```diff
@@ -0,0 +1,31 @@
...
+  </PropertyGroup>   
+    
+  <Choose>
+    <!-- Avoid false warning about missing reference (msbuild bug) -->
+    <!-- https://github.com/dotnet/msbuild/issues/9698#issuecomment-1945763467 -->
+    <When Condition=" '$(EnableNUnitRunner)' == 'true' ">
+      <ItemGroup>
+        <Reference Include="NUnit3.TestAdapter">
```
Evangelink, line 15: With the current layout, the adapter now needs to be added as a reference as it contains the `AddNUnit` extension point used by the hook. It's definitely possible to produce a different dll if you prefer.

src/NUnitTestAdapter/NUnit.TestAdapter.csproj
```diff
@@ -32,19 +32,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Testing.Extensions.VSTestBridge" Version="1.0.1" />
```
Evangelink, line 35: The bridge as a deps on object model (Microsoft.TestPlatform.ObjectModel >= 17.5.0). If that's a too big constraint for you, we need to do some manual tests here to see if we would break something when doing a forced downgrade here.

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitBridgedTestFramework.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            lock (InitializationLock)
+            {
+                if (initialized)
+                {
+                    return;
+                }
+
+                // TODO: Uncomment this line when InternalsVisibleTo is set up.
+                // var nopWriter = new InternalTraceWriter(new StreamWriter(Stream.Null));
```
Evangelink, line 87: I am not sure what's different but on our last test with NUnit we were able to access this type.

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitExtension.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Threading.Tasks;
+
+using Microsoft.Testing.Platform.Extensions;
+
+namespace NUnit.VisualStudio.TestAdapter.TestingPlatformAdapter
+{
+    internal sealed class NUnitExtension : IExtension
```
Evangelink, line 7: Extension points in `Microsoft.Testing.Platform` uses this base interface that provides some identity about the extension. This is used to ease diagnostic for end users.

src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Testing.Platform.Capabilities.TestFramework;
+
+namespace NUnit.VisualStudio.TestAdapter.TestingPlatformAdapter
+{
+    public static class TestApplicationBuilderExtensions
+    {
+        public static void AddNUnit(this ITestApplicationBuilder testApplicationBuilder, Func<IEnumerable<Assembly>> getTestAssemblies)
+        {
+            NUnitExtension extension = new();
+            testApplicationBuilder.AddRunSettingsService(extension);
```
Evangelink, line 17: Register support for runsettings (provided by the VSTest bridge)

src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+            testApplicationBuilder.AddTestCaseFilterService(extension);
```
Evangelink, line 18: Register support for `--filter` as it was handled by VSTest.

src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+            testApplicationBuilder.RegisterTestFramework(
```
Evangelink, line 19: Register NUnit as the test framework for this test application.

MarcoRossignoli:
:shipit:

nuget/NUnit3TestAdapter.nuspec
OsirisTerje: What does this telemetry package do?
If it sends information to some central server, we're not very keen of having that into the adapter.  In understand it is optional, but if any telemetry should be sent, it should be to the NUnit project, and we don't have any server for receiving telemetry.

SeanKilleen:
I'd personally like to see https://github.com/microsoft/testfx/issues/2298 resolved before we move forward with this, to ensure that Microsoft is actually embracing a Framework agnostic runner vs trying to extinguish other frameworks (the comment referenced in that issue is...not encouraging.)

nuget/net462/NUnit3TestAdapter.targets
OsirisTerje: Is it necessary to disable the VSTest entry point?  
It would be better if the user decides without having to use any parameters.  If he runs it using `dotnet test`, it would use that entry point, if he runs the executable, it would use that entry point.

nuget/NUnit3TestAdapter.nuspec
Evangelink: This is collecting "general" data about the platform usage (see https://learn.microsoft.com/dotnet/core/testing/unit-testing-mstest-runner-telemetry#data-points). This is something that is backed in VSTest, SDK... most MS tools but in our case we wanted to ensure that the core platform is not having any external deps so we had to move the telemetry into its own extension point.

We are suggesting to keep it, although I understand fear/concerns..., mainly because this is helping us justify priorization of work to management. Especially now that we have 2 competing platforms.

nuget/net462/NUnit3TestAdapter.targets
Evangelink: VSTest always creates its own entry point. Our new platform also wants to auto-generate it (to simplify user experience) so we need to have them mutually exclusive.

> If he runs it using dotnet test

Not really, as you could see in the example because `dotnet test` is pretty common gesture, we have a mode to hook into it skipping VSTest redirection and calling the runner instead.

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitBridgedTestFramework.cs
OsirisTerje: Yes, it is public. And has been so since 2013.

Evangelink:
@SeanKilleen I can already close the issue you are referencing. As you can see in the nuget packages added everything is already MSTest agnostic. The only thing that remains MSTest oriented is the main doc.

On the contrary to VSTest which is the main tool people use, the new platform is hidden and not something people are much aware of.

SeanKilleen:
@Evangelink great, thanks for confirming!

Evangelink:
@OsirisTerje @SeanKilleen How do you guys want me to move on this PR? Not sure if you prefer to discuss on the PR or ticket.

1. Are you happy with having the new platform has optional disabled by default on the same nuget package or would you like to have a different package with maybe the runner enabled by default?

1. If we move on with a flag to enable, how do you want to name it (current draft uses `EnableNUnitRunner`)?

Not for now but for some future, I'd be happy to start seeing the "VS" part going away (maybe if/when we are able to drop support from VSTest and focus only on the new platform). We are really trying to have the new platform felt more as a dotnet tooling/sdk concept rather than a VS centric feature.

OsirisTerje:
- Best to discuss things on the ticket, as that will kind of persist more than the PR - as we link more to issues than to PRs.

So I'll simply copy your questions here over to the ticket and start answering there.

PS: Are the relevant packages now published so we could drop the local feed and get the CI build green?

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitBridgedTestFramework.cs
Evangelink: @OsirisTerje This is not public, the class is internal
image

We need you to set the `InternalsVisibleTo` from nunit.engine.core to testadapter.

Evangelink: My bad, I just checked the old code and we were actually referencing `nunit` package in addition to `nunit.engine`.

Evangelink:
Here is my build and sample testing workflow:

1. build.cmd
2. build.cmd -t package
3. cd samples\1-nunit-runner
4. dotnet build
5. .\bin\Debug\net6.0\nunit-runner.exe
6. cd ..\2-nunit-runner-dotnet-test\
7. dotnet test

Could you confirm this is working for you?

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitBridgedTestFramework.cs
OsirisTerje: We're talking about two different InternalTraceWriter classes, I referred to the one in the framework:
image

Referencing the framework from the adapter might be troublesome.  We need to ensure people can use different versions  of nunit with the same adapter.  Possibly some clever use of wildcards on the versions can handle it. 

Changing the engine to make this public instead of internal is doable.

OsirisTerje:
I changed the nuget.config to point to a relative path
image

But as the packages are not there, the build fails.
image

I believe the best is to upload the packages to e.g. MyGet or as beta packages to nuget.

nuget.config
OsirisTerje: Should not be here:  
```suggestion
    <!--<add key="Local" value="C:\src\nunit3-vs-adapter\package" /> -->
```

OsirisTerje: Add a file  `nuget.config` in the samples directory:

```xml
<configuration>
  <packageSources>
    <clear />
    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
    <add key="Local" value="../package" />
  </packageSources>
</configuration>

```

Evangelink:
> I changed the nuget.config to point to a relative path

Thanks, I called out this was not workin and was only a temp solution to demo you something so we can start some discussion.

> Also, the samples folder should be moved into either the NUnit-*-Samples repo or the docs repo

Same as above, this is just a temp "solution" to demo. I expect quite some cleaning to do here.

Output from 4 is using vstest output (old dotnet test) while output from 7 uses the new default.

> No information here about tests however, except that "they" succeeded.

This is to reduce output contention by default. We know this is not the most loved solution and we have this issue https://github.com/microsoft/testfx/issues/2162 to track improving this output for "interactive" scenarios.

For the tests, I expect you may want to "improve" them to run maybe in the 2 modes (runner or vstest) for better safety.

OsirisTerje:
1.  I have copied the Samples folder into the adapter issues repo, at https://github.com/nunit/nunit3-vs-adapter.issues/tree/master/Issue1152.  

[ ] It should be removed from this PR.
[ ] Referring to the package can be done from a local location, e.g. C:\nuget, or pretty soon now, from Myget. 

2.  We depend on the objectmodel version  11. This is for backward compatibility down to VS 2012.   (We actually don't know anymore which VS version the adapter works for currently.)   And we don't have any metrics either, nor any issues raised wrt this.
I assume you want 4.6.2 to be able to run the new Testing.Platform, and that is fine. That should work enabling the later objectmodel, but when it runs in the "old" mode, it should still run version 11.  I believe it will make for some interesting configurations.

3. Please update this PR with the current master branch.  That should make the PR pass the Cake CI build. 
4. Then please redirect this PR to the branch named `vnext`.  It is currently pointing to the same commit in master.  We can then use this branch as an integration branch, and also deploy an alpha-version from that branch.  It will also make it easier for us to work on this.

nuget/net462/NUnit3TestAdapter.props
OsirisTerje: Happy with that name!

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitBridgedTestFramework.cs
Evangelink: I honestly don't remember well all details but we used this trick following a bug we were having. @nohwnd Do you recall?

Evangelink:
@OsirisTerje I have commented `ObjectModel` v11 in the test because that's not clear how you want to handle that case.

I have been discussing with @nohwnd and having dependency on newer version of ObjectModel should not break support for older VS.

OsirisTerje:
@Evangelink  If it doesn't crash the older ones, we can remove it.  We just have to trust you on that one.  :-) 

I no longer have any of those installed, and not sure if it will break on other things, like what framework we're actually now running under, so, anyway... Then we let that go.

Evangelink:
> @Evangelink If it doesn't crash the older ones, we can remove it. We just have to trust you on that one. :-)

It shouldn't said our VSTest expert :)

> The CI breaks due to the nuget local src, used for the samples. That should be removed. I have already moved the samples themselves, so when this is merged (and pushed to myget using the Publish build), the samples should grab the new version from myget.org

Yep sorry, will fix that now.

Evangelink:
Working on a fix for the tests.

src/NUnitTestAdapter/TestingPlatformAdapter/NUnitBridgedTestFramework.cs
nohwnd: I don't remember, but all the details of the fix, including link to the issue it is fixing are in the code in form of comments. 🙂 If we need to change our dependencies and use class that is not public, we instantiate it via reflection as well, since we are already patching it via reflection. But looks like you reverted the `// new ...` so I guess it is okay as is.

PR #1172, opened 2024-05-06
Fix NUnit extension version

Evangelink:
Use AssemblyInformationalVersionAttribute version or fallback to
AssemblyVersion instead of using hardcoded version.


OsirisTerje (approved):
lgtm

PR #1249, opened 2025-03-08
Fixes #1248, Allowing to exclude stacktrace in output.

### LINKED ISSUE #1248: How to completely suppress the Stack Trace in Test Results?
Hello,

I'm trying to find a way to suppress the Stack Trace from showing up in test results as it is irrelevant in my specific case and cluttering up the results view in VS and AzDevops.

### Custom Exception

Throwing a custom exception directly which overrides the StackTrace property works:

```cs
    public class MyAssertionException : AssertionException
    {
        public MyAssertionException(string message) : base(message)
        {
        }

        public override string StackTrace => string.Empty;
    }
```

...but then I cannot use `Assert.That(....),` which is producing nice result messages.

### Global Exception Handler

```cs
    AppDomain.CurrentDomain.FirstChanceException += CurrentDomain_FirstChanceException;
```

In the handler, I'm clearing the stack trace fields via Reflection, but then I still get a stack trace with a single line:

Image

### StackFilter

Then I found StackFilter, which I could set to match everything:

```cs
    StackFilter.DefaultFilter = new StackFilter("^.*$", "^.*$");
```

Now, the line content is gone, but the section is still there:

Image

I wonder whether there's any better way for suppressing the stack trace entirely.

Thanks for any hints or insights!

sw

OsirisTerje:
Fixes #1248, Allowing to exclude stacktrace in output.

Adding runsettings IncludeStackTrace, default true.

### DIFF
--- a/src/NUnitTestAdapter/AdapterSettings.cs
+++ b/src/NUnitTestAdapter/AdapterSettings.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2014-2021 Charlie Poole, 2014-2022 Terje Sandstrom
+// Copyright (c) 2014-2021 Charlie Poole, 2014-2025 Terje Sandstrom
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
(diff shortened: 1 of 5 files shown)


manfred-brands (commented):
Looks good

PR #1253, opened 2025-03-27
Fix incorrect condition on TestingPlatformBuilderHook

Youssef1313:
@OsirisTerje This is currently blocking some users from having their own entry point.


OsirisTerje (approved):
Is there something else that stop this Hook for being active ?  
Or, doesnt it matter ?

PR #1279, opened 2025-06-10
Use FQN as TestNodeUid in MTP

Youssef1313:
@OsirisTerje Can you please validate if this fixes the issue for Retry extension? https://github.com/nunit/nunit3-vs-adapter/issues/1261

PR #1283, opened 2025-07-11
Update MTP to 1.7.3

Youssef1313:


Youssef1313:
I'll include in https://github.com/nunit/nunit3-vs-adapter/pull/1279

Issue #1292, opened 2025-08-14
(Apparently) inconsistent version requirements of Microsoft.Testing.Extensions.VSTestBridge


lennartb-:
Hi,

I'm currently troubleshooting an issue we have since updating to NUnit3TestAdapter to 5.1.0. We're getting the build error:

```
Assembly 'NUnit3.TestAdapter' with identity 'NUnit3.TestAdapter, Version=5.1.0.0, Culture=neutral, PublicKeyToken=4cb40d35494691ac' uses 'Microsoft.Testing.Platform, Version=1.8.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'Microsoft.Testing.Platform' with identity 'Microsoft.Testing.Platform, Version=1.7.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```

Which seemed strange to me, because NuGet says that the package requires version >=1.7.3, so where does 1.8.1.0 come from?

It seems that in [nuget/NUnit3TestAdapter.nuspec](https://github.com/nunit/nunit3-vs-adapter/blob/9893f2cbdb878ba474f42f8ff63be6ec69549fd7/nuget/NUnit3TestAdapter.nuspec#L30), the requirements for NuGet are set to 1.7.3, while in [Directory.Packages.props](https://github.com/nunit/nunit3-vs-adapter/blob/9893f2cbdb878ba474f42f8ff63be6ec69549fd7/Directory.Packages.props#L20) it's 1.8.1.

I can fix this locally by just explicitly installing 1.8.0, but I wondered whether the difference in the package is okay and/or intended.

manfred-brands:
@OsirisTerje I have copied and adapted the `NuspecDependenciesTests` test from nunit where we had the same problem.
It will likely need some tweaking to suite this project.

It did find the reported inconsistency:
```
Package Microsoft.Testing.Extensions.VSTestBridge in .csproj should have version '1.8.1' in .nuspec
Assert.That(nuspecVersion, Is.EqualTo(pair.Version))
  String lengths are both 5. Strings differ at index 2.
  Expected: "1.8.1"
  But was:  "1.7.3"
  -------------^
```

moh-hassan:
I update my test project for the package `NUnit3TestAdapter`  from v 5.0.0 to v 5.1.0 , then get compilation error:
[...]

PR #1293, opened 2025-08-15
Add test to compare package version in nuspec vs .csproj

manfred-brands:
Partly addresses #1292 

It finds the mismatched version:
```
  2)   Package Microsoft.Testing.Extensions.VSTestBridge in .csproj should have version '1.8.1' in .nuspec
Assert.That(nuspecVersion, Is.EqualTo(pair.Version))
  String lengths are both 5. Strings differ at index 2.
  Expected: "1.8.1"
  But was:  "1.7.3"
  -------------^
```

It also find missing packages, but I don't know if these are valid.
If the adapter depends on the engine should this also be reflected in the nuspec or is the engine packaged in?

```
  1)   Missing packages in framework 'net462' in .nuspec: SourceLink.Create.CommandLine, nunit.engine, TestCentric.Metadata
Assert.That(missingPackages, Is.Empty)
  Expected: <empty>
  But was:  < "SourceLink.Create.CommandLine", "nunit.engine", "TestCentric.Metadata" >
```

It also found a package listed in the .nuspec, which is not a dependency of the .csproj:
```
     Packages in .nuspec that are not in .csproj and should be deleted from nuspec: Microsoft.Testing.Platform.MSBuild
Assert.That(missingPackages, Is.Empty)
  Expected: <empty>
  But was:  < "Microsoft.Testing.Platform.MSBuild" >
```

### DIFF
--- a/src/NUnitTestAdapterTests/NuspecDependenciesTests.cs
+++ b/src/NUnitTestAdapterTests/NuspecDependenciesTests.cs
@@ -1,0 +1,312 @@
+﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using NUnit.Framework;
+
+#nullable enable
+
+namespace NUnit.VisualStudio.TestAdapter.Tests
+{
+    /// <summary>
+    /// Checks that the csproj of the framework projects are reflected correctly in the nuspec files.
+    /// </summary>
+    [TestFixture("NUnit3TestAdapter.nuspec", "NUnitTestAdapter/NUnit.TestAdapter.csproj")]
+    internal sealed class NuspecDependenciesTests
+    {
+        private readonly string _nuspecPath;
+        private readonly string _csprojPath;
+        private readonly string _propsPath;
+
+        private Dictionary<string, List<PackageWithVersion>> _nuspecPackages;
+        private Dictionary<string, List<PackageWithVersion>> _csprojPackages;
+        private Dictionary<string, string> _csprojPackageVersions;
+
+        private const string Root = "../../../../../";
+        private const string NotSpecified = nameof(NotSpecified);
+
+        public NuspecDependenciesTests(string nuspecPath, string csProjPath)
+        {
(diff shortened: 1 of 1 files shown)


OsirisTerje:
@manfred-brands This test is awesome.  

@Evangelink @Youssef1313 Can you comment on the last one, this is part of your MTP stuff

About the engine, that is packed/embedded in, and we don't list that as a dependency.  You see it in the file block.  

The TestCentric.Metadata I need to check out, if I update it to the next patch version, it breaks the build, so there are something there I have not have time to figure out yet, so for now, it  should be ignored.

About SourceLink.Create.CommandLine, this is only for development, so should not be part of anything deployed.

src/NUnitTestAdapterTests/NuspecDependenciesTests.cs
```diff
@@ -0,0 +1,312 @@
...
+        private readonly string _csprojPath;
+        private readonly string _propsPath;
+
+        private Dictionary<string, List<PackageWithVersion>> _nuspecPackages;
+        private Dictionary<string, List<PackageWithVersion>> _csprojPackages;
+        private Dictionary<string, string> _csprojPackageVersions;
+
+        private const string Root = "../../../../../";
+        private const string NotSpecified = nameof(NotSpecified);
+
```
OsirisTerje, line 30: Suggest adding a list with the packages that are development usage only, and those we should ignore due to they being embedded.

Youssef1313:
`Microsoft.Testing.Platform.MSBuild` is expected to be packaged even if the project doesn't need a PackageReference to it.

Youssef1313:
Generally, I think it's much better to just move away completely from nuspec packaging :)

OsirisTerje:
> Generally, I think it's much better to just move away completely from nuspec packaging :)

Agree, but the adapter is not that easy to move over.

OsirisTerje:
> Microsoft.Testing.Platform.MSBuild is expected to be packaged even if the project doesn't need a PackageReference to it.

What exactly do you mean here?  Is it a development dependency or should it be part of the resulting package ?  Currently it is not being referenced in the csproj and not included **in** the package, but **by** the package as a dependency.

And if needed, any reason why it cant be included by the TestBridge itself?

Youssef1313:
It's a development dependency for **users** of NUnit. So it should be a dependency of NUnit3TestAdapter, but doesn't need to be referenced in your own csproj. The way it's done currently is good.

PR #1305, opened 2025-09-02
MTP v2

Youssef1313:
@OsirisTerje 

FYI @nohwnd

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,7 +4,7 @@
     <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
   </PropertyGroup>
   <ItemGroup>
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="nunit" Version="4.3.2" />
     <PackageVersion Include="nunit.engine" Version="3.18.1" />
--- a/src/NUnit.TestAdapter.Tests.Acceptance/AcceptanceTests.cs
+++ b/src/NUnit.TestAdapter.Tests.Acceptance/AcceptanceTests.cs
@@ -19,7 +19,7 @@ public class Frameworks
 public abstract class AcceptanceTests
 {
     public static string NuGetPackageId => "NUnit3TestAdapter";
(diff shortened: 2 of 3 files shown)


OsirisTerje:
@Youssef1313 The build fails :-)

Youssef1313:
@OsirisTerje As VSTest 17.14 dropped support for old unsupported TFMs, are you fine following that for NUnit as well?

OsirisTerje:
@Youssef1313 
For NUnit 4.X we only support from .net6.   We will drop net6 soon, but then it will be version 5.X. 
For the adapter, it runs net framework 462 and netcore 3.1.  Will probably add .net 8 to that mix, and fine by not letting netcore 3.1 support MTP then.  But there are a lot of people running legacy versions, so as long as I can maintain the adapter for different frameworks without making more documentation on which versions to use for what, aka less surprises, I will do that.

OsirisTerje:
@Youssef1313 Have you looked at why it fails in tests for net8 ?  The main branch doesn't, so it must be related to your changes somehow.

AHH

Just saw your changes.  That will not work, net 8 has to be added, but you cant remove netcore3.1 just like that, if that is needed for MTP.  We're testing using all reasonable frameworks though.   There are a lot of small stuff which needs to be aligned to make it work.

Youssef1313:
I haven't had much chance to look.

But VSTest 17.14 dropped support for netcoreapp3.1-net7.0 range. And as VSTest/MTP are packed into the same package where we are adding a dependency on VSTest ObjectModel 17.14, dropping netcoreapp3.1 might be needed.

netcoreapp3.1 is way too old and is out of support. Any reason you want to still support this in NUnit?

Users needing netcoreapp3.1 can simply just stay on old version of NUnit adapter.

The next release that supports MTP v2 can also be a major version, to indicate a breaking change for old TFMs.

Youssef1313:
Alternatively, we can go with a little bit more complicated approach where the package depends on VSTest 17.13 for netcoreapp3.1 build (and don't include dependency on MTP at all in this case), and bring MTP only for net8.0+.

But I suggest simply dropping netcoreapp3.1. NUnit users relying on it can simply just stay on the old version of NUnit.

Youssef1313:
@OsirisTerje It's green now. Let me know if this is how you want to proceed.

OsirisTerje (approved):

Youssef1313:
@OsirisTerje Are you going to publish this soon? Potentially as 6.0.0?
People are asking in testfx as they are not able to use MTP v2 with NUnit.

PR #1417, opened 2026-02-24
Removing default timeout from .net 6 and up. No longer sent to the fr…

OsirisTerje:
…amework.


OsirisTerje:
It is only the adapter that reads the .runsettings, but I believe the nunit.console also can set it its own way.  Correct @CharliePoole ?

Issue #1455, opened 2026-06-16
Microsoft.Testing.Platform: emit public TestMethodIdentifierProperty and stop relying on the vstestProvider capability


Evangelink:
This adapter opts into the MTP `vstestProvider` capability here: `src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs`.

## Background

When a test app runs in **Microsoft.Testing.Platform (MTP) server mode** (how Visual Studio Test Explorer talks to it), the VSTest bridge can expose test metadata to the IDE in two ways:

1. **Legacy** `vstest.TestCase.*` properties (e.g. `vstest.TestCase.Id`, `vstest.TestCase.FullyQualifiedName`), serialized via an **internal** `SerializableKeyValuePairStringProperty` key/value bag.
2. **Public, structured** property types — in particular [`TestMethodIdentifierProperty`](https://github.com/microsoft/testfx/blob/7984ff9be0d31e5e3b859e93b448eb7c2b0b5e46/src/Platform/Microsoft.Testing.Platform/Messages/TestMethodIdentifierProperty.cs), which serializes to the well-known `location.type` / `location.method` / `location.method-arity` wire properties.

VS picks which path to consume based purely on the server-advertised `vstestProvider` capability. **This adapter currently opts into `vstestProvider`** (by registering [`VSTestBridgeExtensionBaseCapabilities`](https://github.com/microsoft/testfx/blob/7984ff9be0d31e5e3b859e93b448eb7c2b0b5e46/src/Platform/Microsoft.Testing.Extensions.VSTestBridge/Capabilities/VSTestBridgeExtensionBaseCapabilities.cs#L16-L32), whose `INamedFeatureCapability.IsSupported("vstestProvider")` returns `true`), which forces VS onto path **#1** — the legacy internal KVP.

The testfx / VS Test Explorer team is working to **deprecate and remove that internal KVP** and have all IDE metadata flow through **public** property types. MSTest has already moved off the `vstestProvider` path; NUnit and Expecto are the only two adapters still using it.

## How MSTest does it (the model to follow)

MSTest does **two** things:

[...]

Issue #1351, opened 2025-11-04
Test class name missing in TRX report for parametrized fixture when running through MTP


radim-bernatik-veeam:
In the following example is situation, when class name is not fully included in resulting TRX report. Only the namespace of the class is included, but the class name is missing. There are newest packages for MTP (2.0.1) and NUnit3TestAdapter (6.0.0-beta.1). But the problem was firstly recognized by me on MTP/1.8.1 and NUnit3TestAdapter/5.0.1. I also checked NET 8 and NET 10 rc1 frameworks, issue is reproducible on both.

In the TRX report for both instances of the `Test1` test is only `className="One"`, when running using MTP.
The TRX report fragment:
```
    <UnitTest name="Test1" storage="d:\git\mtp\one\bin\debug\net10.0\one.dll" id="f250b33b-8a35-b41a-e7bf-8e9547f93a26">
      <Execution id="983facad-df20-4239-a725-afd3d6150424" />
      <TestMethod codeBase="D:\Git\MTP\One\bin\Debug\net10.0\One.dll" adapterTypeName="executor://NUnitExtension/1.0.0+c7d97261880dc4b9bcf3f48c2cabc5293016adba" className="One" name="Test1" />
    </UnitTest>
```
Execution command: `dotnet exec One\bin\Debug\net10.0\One.dll --results-directory TestResults --report-trx`

On the other side, when same example is executed by `vstest.console.exe` command, the TRX report contains proper `className="One.Tests(One)"` and `className="One.Tests(Two)"`.
The TRX report fragment:
```
  <TestDefinitions>
    <UnitTest name="Test1" storage="d:\git\mtp\one\bin\debug\net10.0\one.dll" id="31e09d51-24f9-f1bf-3546-ea0d27e9d1e2">
      <Execution id="205cb3bb-29b2-47ff-8f7a-16721f04c50a" />
      <TestMethod codeBase="D:\Git\MTP\One\bin\Debug\net10.0\One.dll" adapterTypeName="executor://nunit3testexecutor/" className="One.Tests(One)" name="Test1" />
    </UnitTest>
    <UnitTest name="Test1" storage="d:\git\mtp\one\bin\debug\net10.0\one.dll" id="b28b10d5-d1e1-b8e3-f896-8c4751d57b96">
      <Execution id="8cd2ab2d-6490-4b08-8438-500a777931a3" />
[...]

PR #1456, opened 2026-06-16
MTP: stop advertising vstestProvider so the IDE consumes public properties

Evangelink:
> ⚠️ **Draft / proposal** for discussion — see #1455 for the full background.

## What

Stop advertising the Microsoft.Testing.Platform `vstestProvider` capability from the NUnit adapter, so Visual Studio Test Explorer consumes the **public** structured node properties instead of the legacy `vstest.TestCase.*` key/value-pair properties.

## Why

When the adapter advertises `vstestProvider`, VS routes NUnit through its legacy parser that reads `vstest.TestCase.*` properties, which are serialized via Microsoft.Testing.Platform's **internal** `SerializableKeyValuePairStringProperty` key/value bag. The testfx team is deprecating that internal type in favor of public property types.

The good news: this adapter **already emits the public `TestMethodIdentifierProperty`** in `NUnitBridgedTestFramework.AddAdditionalProperties` (serialized as `location.type` / `location.method` / `location.method-arity`), and `location.file` / `location.line-start` come from the bridge. So the only thing still pinning VS to the legacy path is the `vstestProvider` capability advertised by `VSTestBridgeExtensionBaseCapabilities`.

## How

Replace `VSTestBridgeExtensionBaseCapabilities` (which implements both `IInternalVSTestBridgeTrxReportCapability` **and** `INamedFeatureCapability("vstestProvider")`) with a small capability that implements only the public `ITrxReportCapability`. This mirrors MSTest's [`MSTestCapabilities`](https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs#L21-L31), which already dropped `vstestProvider`.

(MSTest implements the *internal* `IInternalVSTestBridgeTrxReportCapability` because `MSTest.TestAdapter` is on the bridge's `InternalsVisibleTo` list. External adapters can't, so this uses the public `ITrxReportCapability`. Note: the platform's `TrxDataConsumer` checks `IsSupported` then calls `Enable()`, so trx enrichment is effectively always-on with the public interface; the trx-only node properties are ignored by the server-mode (IDE) serializer, so there is no wire-format impact.)

## Notes / things to validate before this leaves draft

- **Sequencing:** the VS Test Explorer side must be able to recover everything from the public properties before this ships, otherwise NUnit navigation/grouping could regress. The non-`vstestProvider` path in VS already reconstructs the FQN from `location.type`/`location.method`, but please coordinate timing.
- **Parameterized tests:** VSTest's FQN encodes argument *values* (`Ns.Cls.Method(x: 1)`) which `TestMethodIdentifierProperty` (method name + arity, no values) can't represent. Per-row uniqueness should rely on the public `TestNode.Uid` (already unique per case). Worth a focused test pass on `TestCaseSource` / fixture-parameterized cases.
- **TRX reports:** `--report-trx` still works because the capability reports `IsSupported = true`.

Builds clean for `net462` + `net8.0`.

Refs #1455

---

### Additional consideration: `vstest.original-executor-uri` / framework identity

Dropping `vstestProvider` also stops the testfx bridge from emitting `vstest.original-executor-uri` (it is only sent from `CopyVSTestProviderProperties`, which runs under `vstestProvider`). In VS that property feeds:

- FQN-normalization kind — now moot, because this adapter emits managed names so the declaring-syntax kind is `ManagedMethod` on both parser paths.
- **Framework identity** (the `TestFramework` enum behind "Associate to Test Case"), which keys off the originating adapter URI.

**Parity note:** the bridge already only emits `original-executor-uri` under `vstestProvider`, so **MSTest — which already dropped `vstestProvider` — does not send it on the public path either.** This change therefore brings this adapter to exact parity with MSTest-via-MTP. If framework-identity features still work for MSTest under MTP they will work here too; if they depend on `original-executor-uri`, that is a **pre-existing gap shared with MSTest** and should be solved in testfx with a *public* framework/adapter-identity property — not in this adapter.

Everything required for navigation / grouping / source mapping is covered by public properties: `location.type` / `location.method` (this PR) and `location.file` / `location.line-start` (emitted unconditionally by the bridge).

---

### ⚠️ Correction / important caveat: parameterized-method identity is NOT free

An earlier framing of this change suggested dropping `vstestProvider` is essentially free once `TestMethodIdentifierProperty` is emitted. That is **not accurate**, and reviewers should weigh this:

- `TestMethodIdentifierProperty` / `location.method` carries the method name + parameter **types**, not the argument **values**.
- **Parameterized fixtures** (`TestFixture(...)`) encode their args in the *type* name (`Tests(One)`), which the builder preserves — these survive on the public path (this is what #1351 fixed).
- **Parameterized methods** (`[TestCase]` / `TestCaseSource`) encode their args in the FQN (`Tests.Test1(1)`). On the public path `location.method` becomes `Test1(System.Int32)` — **identical for every data row** — so all rows reconstruct to a single FQN and are distinguished only by the public `TestNode.Uid` + `DisplayName`.

Today, under `vstestProvider`, VS receives the per-row `vstest.TestCase.FullyQualifiedName` (`Tests.Test1(1)`). Dropping `vstestProvider` **collapses** that per-row FQN. Navigation and display still work via `Uid`/`DisplayName`, but the per-row **FQN hash** used by Azure DevOps "Associate to Test Case" would no longer be unique per row.

**Implication:** this migration should be sequenced with the VS Test Explorer side keying per-row identity off the public `Uid` (or a public property carrying the full parameterized name) before `vstestProvider` is fully dropped, otherwise method-parameterized tests regress. (Note: `TestMethodIdentifierProperty` here was originally added for the TRX class-name fix in #1351/#1449, independently of this capability change.)

### DIFF
--- a/src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
+++ b/src/NUnitTestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Reflection;
 
-using Microsoft.Testing.Extensions.VSTestBridge.Capabilities;
+using Microsoft.Testing.Extensions.TrxReport.Abstractions;
 using Microsoft.Testing.Extensions.VSTestBridge.Helpers;
 using Microsoft.Testing.Platform.Builder;
 using Microsoft.Testing.Platform.Capabilities.TestFramework;
@@ -11,6 +11,33 @@ namespace NUnit.VisualStudio.TestAdapter.TestingPlatformAdapter
 {
     public static class TestApplicationBuilderExtensions
     {
+        // We intentionally do NOT use VSTestBridgeExtensionBaseCapabilities because that type also implements
+        // INamedFeatureCapability and advertises the "vstestProvider" capability. When a server (this adapter)
+        // advertises vstestProvider, Visual Studio Test Explorer consumes the legacy "vstest.TestCase.*"
+        // properties (serialized through Microsoft.Testing.Platform's internal SerializableKeyValuePairStringProperty
+        // key/value bag) instead of the public, structured properties.
+        //
+        // NUnitBridgedTestFramework.AddAdditionalProperties already emits the public TestMethodIdentifierProperty
+        // (serialized as location.type / location.method / location.method-arity), and location.file /
+        // location.line-start are emitted by the bridge from the TestCase. By not advertising vstestProvider we let
+        // the IDE consume those public properties and we stop depending on the internal key/value-pair shape.
+        //
+        // This mirrors what MSTest does with its own MSTestCapabilities type:
+        // https://github.com/microsoft/testfx/blob/main/src/Adapter/MSTest.TestAdapter/TestingPlatformAdapter/TestApplicationBuilderExtensions.cs
+        // MSTest can implement the internal IInternalVSTestBridgeTrxReportCapability (it is on the bridge's
+        // InternalsVisibleTo list); external adapters cannot, so we implement the public ITrxReportCapability here.
+        private sealed class NUnitTestFrameworkCapabilities : ITrxReportCapability
+        {
+            // Returning true advertises that NUnit can enrich nodes with the trx-report properties; the platform's
+            // TrxDataConsumer checks IsSupported and then calls Enable() only when a trx report was requested.
+            bool ITrxReportCapability.IsSupported => true;
+
+            void ITrxReportCapability.Enable()
+            {
+                // No-op: NUnitBridgedTestFramework always populates the trx-report properties for discovered/run nodes.
+            }
+        }
+
         public static void AddNUnit(this ITestApplicationBuilder testApplicationBuilder, Func<IEnumerable<Assembly>> getTestAssemblies)
         {
             NUnitExtension extension = new();
@@ -21,7 +48,7 @@ namespace NUnit.VisualStudio.TestAdapter.TestingPlatformAdapter
             testApplicationBuilder.AddRunSettingsEnvironmentVariableProvider(extension);
 
             testApplicationBuilder.RegisterTestFramework(
-                _ => new TestFrameworkCapabilities(new VSTestBridgeExtensionBaseCapabilities()),
+                _ => new TestFrameworkCapabilities(new NUnitTestFrameworkCapabilities()),
                 (capabilities, serviceProvider) => new NUnitBridgedTestFramework(extension, getTestAssemblies, serviceProvider, capabilities));
         }
     }


OsirisTerje:
1. Do I read you right that this change only affects when MTP is selected.  It does not change anything for non-MTP projects ?
2. When you say it has to be "sequenced with Test Explorer", does that mean it will only work with the latest (or upcoming) version of VS 2026?  If so, what will happen when people use an earlier version ?

OsirisTerje:
@Evangelink  Should we try to push this into 6.3 ?

Evangelink:
I validated this change against the current testfx/MTP source (microsoft/testfx `main`). **The approach is correct and rides a path testfx explicitly supports.** Summary for reviewers:

### Validated against testfx source
- **`vstestProvider` is the only thing pinning VS to the legacy properties.** `CopyVSTestProviderProperties` (which emits `vstest.TestCase.*` and `vstest.original-executor-uri`) is gated by `ShouldAddVSTestProviderProperties`, which requires `namedFeatureCapability?.IsSupported("vstestProvider") == true` (`ObjectModelConverters.cs`). Dropping the `INamedFeatureCapability` removes all of that.
- **No other side effect from dropping `INamedFeatureCapability`.** Within the VSTestBridge, that capability is queried *only* for `"vstestProvider"` — nowhere else.
- **TRX still works through the public interface — by design.** `VSTestBridgedTestFrameworkBase.IsTrxEnabled` has an explicit public-interface fallback:
  ```csharp
  => _isTrxEnabled ??= _trxReportCapability is IInternalVSTestBridgeTrxReportCapability internalCapability
      ? internalCapability.IsTrxEnabled
      : _trxReportCapability is ITrxReportCapability { IsSupported: true };
  ```
  The second branch was added precisely so external adapters that can only implement the public `ITrxReportCapability` still get trx enrichment. This PR rides that branch.
- **Mirrors MSTest exactly.** `MSTestCapabilities` likewise implements only the trx capability and no `INamedFeatureCapability`.

### One nuance worth calling out (not a blocker)
Because the public branch keys off `IsSupported: true` (not off `Enable()` being called), **`IsTrxEnabled` becomes always-true for NUnit** — unlike MSTest, whose *internal* capability keeps it false until `--report-trx` is requested. Two implications:

1. **Always-on trx enrichment.** Every discover/run adds `Trx*` node properties even without `--report-trx`. Negligible cost, and the server-mode (IDE) serializer ignores trx-only properties, so no wire impact. (The no-op `Enable()` here is consistent with that.)
2. **A latent throw path becomes reachable on every run.** In `TestResult.ToTestNode`, the `isTrxEnabled` block throws `InvalidOperationException("Unable to parse fully qualified type name…")` when there is **no** `TestMethodIdentifierProperty` **and** the FQN can't be parsed into a type. NUnit's `AddAdditionalProperties` early-returns (adds no property) when the `NUnitClassName`/`NUnitMethodName` seed values are null. For normal `Ns.Class.Method` FQNs this is fine, but it would be worth a focused test on any case where those seed values can be absent.

### "Nothing else needed" — testfx side
Confirmed: **no testfx change is required**; the design already supports this. The remaining items are consumer/sequencing concerns already captured in the description — VS Test Explorer keying per-row identity off the public `TestNode.Uid` before `vstestProvider` is dropped (parameterized methods), and `original-executor-uri`/framework-identity being a pre-existing gap shared with MSTest that should be solved in testfx with a *public* identity property.

Net: ✅ correct change, safe on the testfx side; the only follow-up is the consumer-side sequencing already noted plus a focused test on the always-on-trx edge above.

OsirisTerje:
So the answers to my two questions above are positive ?
Any reason why this is then still a draft ?

Evangelink:
Sorry :)

> Do I read you right that this change only affects when MTP is selected. It does not change anything for non-MTP projects ?

Yes

> When you say it has to be "sequenced with Test Explorer", does that mean it will only work with the latest (or upcoming) version of VS 2026? If so, what will happen when people use an earlier version ?

All versions of 2026 are supporting that. For earlier versions of VS, I am not sure, I'll try to ask around.

> Should we try to push this into 6.3 ?

Would be awesome!

> Any reason why this is then still a draft ?

In case you wanted to guard with adding the tests for the TRX case I mentioned. I'll mark the PR as ready and let you decide.

OsirisTerje:
## Analysis

**Q1 (non-MTP impact):** Confirmed safe. This change is entirely within the `TestingPlatformAdapter` namespace. The classic VSTest path is completely separate — nothing changes for non-MTP projects.

**Q2 (VS version sequencing):** @Evangelink answered VS 2026 is fine, but left VS 2022 as "will ask around." This is the **remaining open risk** before merging.

However, there is a strong argument VS 2022 is also fine: **MSTest already dropped `vstestProvider`** and MSTest works on VS 2022. VS 2022 therefore must already handle the non-`vstestProvider` public-properties path. NUnit would ride the same path.

## Remaining risks before merging

1. **VS 2022 confirmation** — still pending @Evangelink's follow-up, but likely fine given MSTest parity
2. **Parameterized methods (`[TestCase]`)** — per-row FQN collapses to `Method(System.Int32)` instead of `Method(1)`. Tests still run and display correctly via `Uid`/`DisplayName`, but **Azure DevOps "Associate to Test Case"** per-row hashing breaks. This is a known, accepted limitation — not a crash.
3. **Always-on TRX latent throw** — `IsSupported: true` makes TRX enrichment always-on, and there is a code path in `TestResult.ToTestNode` that throws if `TestMethodIdentifierProperty` is absent and the FQN cannot be parsed. @Evangelink flagged this needs a focused test on cases where `NUnitClassName`/`NUnitMethodName` are null.

## Merge decision

Not merging yet — but close. The one blocker is the VS 2022 question. @Evangelink, can you confirm VS 2022 behavior? Once that is answered, the only remaining ask is the TRX edge-case test, which could be accepted as a follow-up issue rather than a hard gate for 6.3.

PS:  This is an analysis done by Claude, not my words, just saying.

OsirisTerje:
Since VS2022 supports this change, we should also be safe.  I don't see any reason not to merge, and not sure we need the two test cases, if I read this correctly.

OsirisTerje (approved):

PR #1464, opened 2026-07-09
Preserve escaped '|' in FullyQualifiedName filter values

azat-msft:
SplitOnFullyQualifiedName now consumes escaped pairs so an escaped '\|' inside a test-argument value is not treated as a clause separator.


OsirisTerje (approved):
Ok, make sense.

PR #1466, opened 2026-07-12
Updating VSTestBridge from version 2.2.3 to version 2.3.1

OsirisTerje:
Tests are failing

@Evangelink Can you have a look here?   Seems this is a breaking change from 2.2.3 to 2.3.1.

This bridge version (2.3.1) fails on loading Microsoft.Testing.Extensions.VSTestBridge.ObjectModel.FilterExpressionWrapper  from Microsoft.Testing.Extensions.VSTestBridge

image

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -18,6 +18,11 @@
     <PackageVersion Include="Castle.Core" Version="5.2.1" />
     <PackageVersion Include="Microsoft.Testing.Extensions.VSTestBridge" Version="2.3.1" />
     <PackageVersion Include="Microsoft.Testing.Extensions.TrxReport" Version="2.3.1" />
+    <!-- Source-only package providing VSTest's filter-expression types (FilterExpressionWrapper /
+         TestCaseFilterExpression), compiled directly into the test assembly. This is the same package the
+         VSTestBridge uses internally, so the tests build the filter expression exactly as the product
+         does. Only available on the dnceng test-tools feed (see nuget.config). -->
+    <PackageVersion Include="Microsoft.TestPlatform.Filter.Source" Version="18.8.0-release-26277-01" />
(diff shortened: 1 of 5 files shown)


Evangelink:
Hey @OsirisTerje - had a look. Good news: this is **not a breaking change to any public VSTestBridge API**, and the shipped adapter itself is fine. The failure is only in the test helper `FilteringTestUtils.CreateVSTestFilterExpression`, which reflects into **internal** bridge types by hard-coded name.

What changed in 2.3.0: we swapped the bridge's hand-maintained filter classes for the source-only `Microsoft.TestPlatform.Filter.Source` package. That did two things to those internal types:

1. **Namespace moved** `Microsoft.Testing.Extensions.VSTestBridge.ObjectModel.FilterExpressionWrapper` -> `Microsoft.VisualStudio.TestPlatform.Common.Filtering.FilterExpressionWrapper`. Your `Type.GetType(..., throwOnError: true)` then throws `TypeLoadException` (the error in your screenshot).
2. **Interface dropped** the new `TestCaseFilterExpression` no longer implements the public `ITestCaseFilterExpression` (the bridge wraps it in an internal adapter now), so even after fixing the namespace the direct cast throws `InvalidCastException`.

Both types were always `internal`, so nothing in the supported surface broke - the test just depended on implementation details that moved.

I pushed a fix: #1467 (targets this PR's branch, so it merges straight in). It probes both namespaces and adapts the expression to `ITestCaseFilterExpression` via reflection, mirroring what the bridge does internally, so it works on both 2.2.x and 2.3.x. Verified locally: `--filter FullyQualifiedName~Filtering` -> 43 passed, 0 failed against 2.3.1.

Heads-up that this helper is inherently fragile since it pokes at bridge internals; longer term it'd be more robust to build the filter expression without reflecting into them.

Evangelink:
18.8 stable should be published this week FYI

PR #1467, opened 2026-07-13
Fix filtering tests for VSTestBridge 2.3.x using the source filter package

Evangelink:
Fixes the failing tests in #1466 (the VSTestBridge 2.2.3 -> 2.3.1 bump). Targets that PR's `minornugetupdate` branch so it can be merged straight in.

### Root cause
This is **not** a breaking change to any public `Microsoft.Testing.Extensions.VSTestBridge` API, and the shipped adapter is unaffected. The failure was entirely in the test helper `FilteringTestUtils.CreateVSTestFilterExpression`, which reached into **internal** bridge types by fully-qualified name via reflection and cast the result to `ITestCaseFilterExpression`.

In bridge **2.3.0** the bridge switched to the source-only `Microsoft.TestPlatform.Filter.Source` package, which changed those internal types in two ways:

1. **Namespace moved** `Microsoft.Testing.Extensions.VSTestBridge.ObjectModel` -> `Microsoft.VisualStudio.TestPlatform.Common.Filtering`, so the old `Type.GetType(..., throwOnError: true)` threw `TypeLoadException` (the error in the screenshot on #1466).
2. **Interface dropped** the new `TestCaseFilterExpression` no longer implements the public `ITestCaseFilterExpression` (the bridge wraps it internally), so even after fixing the namespace the cast threw `InvalidCastException`.

Both types were always `internal`, so nothing in the supported surface broke.

### Fix
Instead of reflecting into bridge internals, reference the **same source-only `Microsoft.TestPlatform.Filter.Source` package the bridge uses** and build the filter expression directly, then wrap it in a small `ITestCaseFilterExpression` adapter - mirroring the bridge's own `BridgeFilterExpression` (and MSTest's `MSTestFilterExpression`). No more reflection.

Changes:
- Add the dnceng public **test-tools** feed to `nuget.config` (this source package isn't published to nuget.org) and pin it in `Directory.Packages.props`.
- Reference the package (`PrivateAssets=all`) from the test project.
- Rewrite `FilteringTestUtils` to construct `new TestCaseFilterExpression(new FilterExpressionWrapper(filter))` and wrap it.
- Suppress `CS0436` on `net8.0`, where the source-compiled types also exist in the referenced `Microsoft.VisualStudio.TestPlatform.Common` binary; the source-compiled copy is the one used.

### Verified
`dotnet test -c Release --filter FullyQualifiedName~Filtering` -> **43 passed, 0 failed on both net462 and net8.0** (was 5 failed before). Full Release build is clean (warnings-as-errors on).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -18,6 +18,11 @@
     <PackageVersion Include="Castle.Core" Version="5.2.1" />
     <PackageVersion Include="Microsoft.Testing.Extensions.VSTestBridge" Version="2.3.1" />
     <PackageVersion Include="Microsoft.Testing.Extensions.TrxReport" Version="2.3.1" />
+    <!-- Source-only package providing VSTest's filter-expression types (FilterExpressionWrapper /
+         TestCaseFilterExpression), compiled directly into the test assembly. This is the same package the
+         VSTestBridge uses internally, so the tests build the filter expression exactly as the product
+         does. Only available on the dnceng test-tools feed (see nuget.config). -->
+    <PackageVersion Include="Microsoft.TestPlatform.Filter.Source" Version="18.8.0-release-26277-01" />
     <PackageVersion Include="Microsoft.Extensions.DependencyModel" Version="10.0.9" />
   </ItemGroup>
   <ItemGroup>
--- a/nuget.config
+++ b/nuget.config
@@ -3,5 +3,9 @@
   <packageSources>
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <!-- VSTest source-only packages (e.g. Microsoft.TestPlatform.Filter.Source) are published to this
+         public dnceng feed rather than nuget.org. It is the same feed the
+         Microsoft.Testing.Extensions.VSTestBridge package consumes the filter sources from. -->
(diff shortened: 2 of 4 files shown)


OsirisTerje:
Awesome!   Thanks @Evangelink !

> Happy to switch to this if you'd prefer taking the preview-feed dependency.

This sems like it is more future proof.  We can accept that preview package feed.

Evangelink:
Updated per the maintainer's preference: switched from the reflection workaround to referencing the source-only `Microsoft.TestPlatform.Filter.Source` package directly (the same one the VSTestBridge uses), so the tests build the filter expression the same way the product does - no reflection.

Notes:
- Adds the dnceng public **test-tools** feed to `nuget.config` (the package isn't on nuget.org).
- Needs a `CS0436` suppression on net8.0, since those types are also in the referenced `Microsoft.VisualStudio.TestPlatform.Common` binary there (the source-compiled copy wins, which is what we want).
- Still keeps a tiny `BridgeFilterExpression` wrapper because, outside the vstest repo, the source-compiled `TestCaseFilterExpression` is internal and doesn't implement `ITestCaseFilterExpression` - same pattern the bridge uses internally.

Verified: 43 passed / 0 failed on both net462 and net8.0, clean Release build.

OsirisTerje (approved):

PR #1489, opened 2026-09-02
Unescape all VSTest filter escape sequences, not just parentheses

### LINKED ISSUE #1405: Test Cases With Strings That Contain the Pipe Character are Not Run
# To Reproduce

* Create a new project that targets .NET 10. Specifically, ensure the project file has this setup:

```xml
<Project Sdk="Microsoft.NET.Sdk">
	<PropertyGroup>
		<EnableNUnitRunner>true</EnableNUnitRunner>
		<LangVersion>latest</LangVersion>
		<TargetFramework>net10.0</TargetFramework>
		<TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
		<ImplicitUsings>enable</ImplicitUsings>
		<Nullable>enable</Nullable>
	</PropertyGroup>
	<ItemGroup>
		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
		<PackageReference Include="NUnit" Version="4.4.0" />
		<PackageReference Include="NUnit.Analyzers" Version="4.11.2">
			<PrivateAssets>all</PrivateAssets>
			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
		</PackageReference>
		<PackageReference Include="NUnit3TestAdapter" Version="6.1.0" />
	</ItemGroup>
</Project>
```

* Add a test to the project:

```c#
using NUnit.Framework;

namespace NUnitIssue;

public static class TestIssue
{
    [TestCase("This | That", false)]
    public static void TestLength(string input, bool expected) =>
        Assert.That(input.Length % 2 == 0, Is.EqualTo(expected));
}
```

# Expected

The test runs and passes.

# Actual

The test doesn't run at all.

# Commentary

If I remove the `|` character, it runs. This is similar to [another issue I reported](https://github.com/nunit/nunit3-vs-adapter/issues/1347), which is fixed, but a pipe character still seems to cause the runner to ignore it.

### LINKED ISSUE #1488: `FullyQualifiedName` filters containing escaped operators (`\=`, `\&`, `\|`, `\!`, `\~`, `\\`) are discovered but never executed
### Environment

| | |
|---|---|
| NUnit3TestAdapter | 6.2.0 (also reproduces on `main` @ fe7287f) |
| NUnit | 4.6.1 |
| Microsoft.NET.Test.Sdk | 17.10.0 |
| VSTest | 18.7.0 (x64) |
| Target framework | net8.0 |
| .NET SDK | 10.0.400 |
| OS | Windows 11 x64 |

### Summary

`TestFilterParser.UnEscape` only reverses `\(` and `\)`. Every other VSTest escape sequence keeps its backslash and is passed through into the NUnit filter, so the generated `<test>` element can never match a real test name.

### Steps to reproduce

```csharp
using NUnit.Framework;

namespace Repro;

public class FilterRepro
{
    [TestCaseSource(nameof(Cases))]
    public void EscapedOperator(string value) => Assert.That(value, Is.Not.Null);

    private static IEnumerable<TestCaseData> Cases()
    {
        yield return new TestCaseData("a").SetName("{m}(Case 1: X = Y)");
    }
}
```

The resulting fully qualified name is:

```
Repro.FilterRepro.EscapedOperator(Case 1: X = Y)
```

**Discovery finds it.** `FilterHelper.Unescape` is applied correctly here:

```powershell
PS > dotnet vstest Repro.dll /ListFullyQualifiedTests /ListTestsTargetPath:out.txt "/TestCaseFilter:FullyQualifiedName=Repro.FilterRepro.EscapedOperator\(Case 1: X \= Y\)" ; Get-Content out.txt
VSTest version 18.7.0 (x64)

Repro.FilterRepro.EscapedOperator(Case 1: X = Y)
```

**Execution does not:**

```powershell
PS > dotnet test Repro.dll --filter "FullyQualifiedName=Repro.FilterRepro.EscapedOperator\(Case 1: X \= Y\)"
VSTest version 18.7.0 (x64)

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
No test matches the given testcase filter `FullyQualifiedName=Repro.FilterRepro.EscapedOperator\(Case 1: X \= Y\)` in Repro.dll
```

**And the escape cannot be omitted**:

```powershell
PS > dotnet test Repro.dll --filter "FullyQualifiedName=Repro.FilterRepro.EscapedOperator\(Case 1: X = Y\)"
VSTest version 18.7.0 (x64)

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
An exception occurred while invoking executor 'executor://nunit3testexecutor/': Incorrect format for TestCaseFilter Error: Invalid Condition 'FullyQualifiedName=Repro.FilterRepro.EscapedOperator\(Case 1: X = Y\)'. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.
Stack trace:
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Discovery.DiscoveryContext.GetTestCaseFilter(IEnumerable`1 supportedProperties, Func`2 propertyProvider) in /_/src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs:line 44
   at ...
```

### Expected behavior

The test runs. Discovery and execution should agree on the meaning of a filter string, and both should follow the escaping rules documented for [selective unit tests](https://learn.microsoft.com/dotnet/core/testing/selective-unit-tests?pivots=nunit).

### Root cause

The following method does not follow the escaping rules documented for [selective unit tests](https://learn.microsoft.com/dotnet/core/testing/selective-unit-tests?pivots=nunit).

`src/NUnitTestAdapter/TestFilterConverter/TestFilterParser.cs`:

```csharp
private string UnEscape(string rhs)
{
    return rhs.Replace(@"\(", "(").Replace(@"\)", ")");
}
```

This affects any test name containing `=`, `&`, `|`, `!`, `~` or `\` — which in
practice means most `TestCaseData.SetName` output that embeds a record/object
`ToString()`.

PanzerFowst:
TestFilterParser.UnEscape only reversed \( and \), so a filter containing \= (which VSTest requires, since a bare = makes Condition.Parse throw) kept its backslash and matched nothing. VSTest discovers the test; the adapter cannot run it.

Fixes #1488 
Fixes #1405


PanzerFowst:
@dotnet-policy-service agree

PanzerFowst:
@OsirisTerje, is there anything I am missing for this PR?  Not in a huge rush, but just wanted to know if there's anything on my side that needs to happen.

OsirisTerje:
@PanzerFowst 

Sorry for the delay.  I have been too busy lately.  

I need to go through your PRs, there are two of them, they seem related.  I did a test on your repro of #1488, and found you are correct there.

You are a new contributor, and if you read our [Contributing](https://github.com/nunit/nunit3-vs-adapter/blob/main/CONTRIBUTING.md) document, you'll see that we would like some more information.  We are very careful with AI generated code, we absolutely allow it, but we need to ensure it has been used very carefully.  

As we write in the doc:

AI-assisted contributions are permitted, but our contribution guidelines require the contributor to understand, verify, and take responsibility for everything submitted. We also ask contributors to discuss the proposed approach and receive confirmation from a maintainer before starting a pull request.

Also, it is a good practice to state that you have used AI assistance.

Your profile also raise some flags, as there is nothing there telling us anything about you, and your activity is hidden.  No location, no information whatsoever.  
Even your profile account name, is well - a little bit too much. 

Working with open source require trust, and trust require open-ness - which is actually the point of open source, right?

Based on you having repos dating back to 2021, we see you are not a new github account, which is good. 

The fact that you actually are requesting feedback, is also very good.  That means you are actually a human! 

However, looking at your contributions here, we see that:

Within about one hour, you submitted:

- Two detailed bug investigations
- Two minimal reproductions
- Two root-cause analyses
- Two pull requests
- Multiple unit tests
- An additional fix for an unbalanced-parenthesis infinite loop inside PR 1491, including extra unit tests for this.

That strongly resembles an agent being asked to inspect the filtering code, find defects, create issues, and submit fixes.

The extra fixed is even not discussed or mentioned.

-> So, are these issues and PRs a result of you hitting some issues in your own work ?


To be constructive:

Could you therefore:

1. Confirm you have read the contribution document.
2. Confirm that you personally ran and verified both reproductions.
3. Confirm you have checked the other issues about the same fundamentals, and explain why you think these issues hits the target better.
4. Explain why using FilterHelper.Unescape is correct for every supported filter operator and whether it has any compatibility implications.
5. Explain the intended tokenizer behavior when whitespace occurs outside versus inside an argument list.
6. Separate the unbalanced-parenthesis change from #1490, since that appears to be a different defect.
7. Add acceptance tests demonstrating the original discovery/execution failures, rather than testing only the modified parser implementation.

Also please add a little information about yourself to your GitHub profile or introduce yourself here, as requested by our contribution guidelines.

OsirisTerje:
If you want to chat privately, you can join our Slack , invite link here:  https://join.slack.com/t/nunit/shared_invite/zt-49hz5l9h0-rh3e5vPt~8E9XF4Jufs2Ag

-----------

Updated Issue #505, which is the parent issue of FQN issues. See there.  #1488 seems to be a duplicate of #1405

PanzerFowst:
No worries!  Thanks so much for responding!  I feel bad for pining you now...I just wanted to make sure my PR didn't get lost in the ether before I forgot about it myself.

I did read the docs hence the creation of the issues before the PRs--I may have misread / ignored the the part about creating the PR before a discussion in the issue was completed...  Sorry.  I think my reasoning was that the two issues are so small and that the PRs should (hopefully?) be easy enough that they aren't too big a deal.  Sorry if that messes up the process.

I did use AI to find the offending line of code, but I came up with the fix myself since my company also has a custom project that creates NUnit filters for the tests to run in parallel where I was using the `FilterHelper.Escape()` method.  I probably should have used AI to verify my added test cases, but I just added them manually myself based on the existing patterns and verified they failed before and passed after my changes.

Yeah, this is my only GitHub account--I thought I created it in high school for FTC, but I guess that must have been a different account.
¯\_(ツ)_/¯

I totally get the trust and AI slop / vulnerabilities--been watching too many LowLevel YT videos--but I also am super wary of what info I put out on myself in public.  I will message you some details privately though if that works for you?  Or if you don't trust me, I am fine with that as long as my issues get resolved--I just get too excited and want to fix things--junior dev energy.  😬

1. I have.
2. I have run it and verified it fixes the issues in my own test runner code by swapping out the `.dll`s.
3. I only spent 10 minutes doing keyword searches through the other issues and none of them seemed to be very specific / triaged, so I created the issue since I knew the offending lines of code.
4. The `FilterHelper` class comes from MS and is the programmatic approach to test filtering according to the MSDN [here](https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=nunit#character-escaping). The problem I was running into was that using `dotnet test --list-tests` would discover the tests and list them correctly but removing the `--list-tests` caused NUnit to crash (maybe I didn't capture that well in my issue write ups?).
5. I think if something works with `--list-tests`, it should work without it too.  That's my main motivation.  I guess the other "requirement" should be that NUnit follows that MSDN link about selecting tests?
6. No, the unbalanced-parenthesis is a related change I'd argue because it relates to whether there is one space or many spaces before the `(...)` params block of a test name.  This was another case of "works with `--list-tests` but not without it"...
7. Could you say more here?  I don't really understand what you're asking?

PanzerFowst:
> #1488 seems to be a duplicate of #1405

Yeah, that is definitely the same problem of unescaped characters messing up the filter.

PanzerFowst:
> However, looking at your contributions here, we see that:
> Within about one hour, you submitted:
> ...
> That strongly resembles an agent being asked to inspect the filtering code, find defects, create issues, and submit fixes.
> The extra fixed is even not discussed or mentioned.
> -> So, are these issues and PRs a result of you hitting some issues in your own work ?

Oh, forgot to answer these...  I submitted it all at once after I verified it...  Didn't want to write up a half-done issue / PR without testing all of them first.

And yes, these are issues I ran into--there is custom attribute as a workaround which just converts all characters NUnit doesn't escape correctly into underscores so that it doesn't mess up the tests in the pipeline, but that workaround is super ugly and messes with VS's Test Explorer view.

I was testing out Opus in VS Code and was using it to track down the bug in the code from that "works with `--list-tests` but not without" and it was super cool to watch it download the NUnit source files to look for the issues and then point the issue out.  Once it showed me that replacement only replacing the escaped parenthesis, I knew the fix.  I also used it to write up the markdown for the issues, but the only thing I honestly kept from that was the table of versions at the top of the issues and then rewrote the rest of it because it was adding too much fluff.

OsirisTerje:
> Add acceptance tests demonstrating the original discovery/execution failures, rather than testing only the modified parser implementation.
> Could you say more here? I don't really understand what you're asking?

If you look in the project, you'll see an acceptance test project.  That projects simulates calling the adapter from externally, like `dotnet test`.  Adding it there would show that it works when you pass the FQN parameters the way you would do in real, with the code being your repro code.  

It might be a bit hard to understand the harness, so this is not mandatory, but a very nice extra.  You can have a look.

PanzerFowst:
Alright, I'll take a look--that sounds like the best place for the tests I want.  Should I revert the other tests then?

OsirisTerje:
No, dont revert them, they can stay,

src/NUnitTestAdapterTests/TestFilterConverterTests/TestFilterParserTests.cs
```diff
@@ -89,6 +89,20 @@ public void CreateParser()
...
+    [TestCase(
+        @"FullyQualifiedName~My.Test.Fixture.Method(Case 1: X \= Y)",
+        @"<test re='1'>My\.Test\.Fixture\.Method\(Case 1: X = Y\)</test>")]
+    [TestCase(
+        @"FullyQualifiedName=My.Test.Fixture.Method(a \& b \| c \! d \~ e)",
+        "<test>My.Test.Fixture.Method(a &amp; b | c ! d ~ e)</test>")]
+    [TestCase(
+        @"FullyQualifiedName=My.Test.Fixture.Method(C:\\Temp)",
+        @"<test>My.Test.Fixture.Method(C:\Temp)</test>")]
+
```
OsirisTerje, line 105: FilterHelper.Unescape now also unescapes \" (quote). Confirm that's harmless for the FQN string-argument path — the existing ("xyz") test still passes, but a [TestCase] row with an escaped quote inside the argument (Method("a\"b")) would document it. One extra test row.

OsirisTerje: Would also be nice to add a line for the #1405 issue, with a pipe character, to ensure #1405 is actually fixed by this.

OsirisTerje (changes requested):

anthony-starkey:
Ah, sorry, life has been crazy.  I am going to get to working on this today and tomorrow--I definitely want to get these bugs fixed.  😅

PanzerFowst:
Ah, sorry, life has been crazy. I am going to get to working on this today and tomorrow--I definitely want to get these bugs fixed. 😅

src/NUnitTestAdapterTests/TestFilterConverterTests/TestFilterParserTests.cs
PanzerFowst: Where do you see the `"` getting unescaped?  I checked the [source code for FilterHelper.Unescape](https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Filter.Source/FilterHelper.cs#L77) and the list of special characters is `['\\', '(', ')', '&', '|', '=', '!', '~']`...  🤔

PanzerFowst: From my testing, `FilterHelper.Unescape("\\\"")` actually will [throw](https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Filter.Source/FilterHelper.cs#L97) because `"` is a character that shouldn't be escaped since it isn't in that list I just shared.

PanzerFowst: > Would also be nice to add a line for the #1405 issue, with a pipe character, to ensure #1405 is actually fixed by this.

There is a pipe on line 100 of this file...  Is that enough or would you like a separate test case?

src/NUnit.TestAdapter.Tests.Acceptance/ParanthesisTests.cs
PanzerFowst, line 6: I added the acceptance tests to this file since it is covering the same logic of escaped VS Test filter characters, but I don't like that it is still called `ParanthesisTests` and would like to change it to `FilterSpecialCharacterTests` (or other suggestions?)--I only kept it for the diff to be easier to read.

OsirisTerje: Ok, do it.

src/NUnitTestAdapterTests/TestFilterConverterTests/TestFilterParserTests.cs
OsirisTerje: Its good

OsirisTerje (approved):

PR #1491, opened 2026-09-03
Fix tokenizer: allow whitespace before '(' and guard unbalanced parens

### LINKED ISSUE #1490: Test names with a space before the argument list cannot be executed: `Unexpected FQN ... in selection expression`
### Environment

| | |
|---|---|
| NUnit3TestAdapter | 6.2.0 (also reproduces on `main` @ fe7287f) |
| NUnit | 4.6.1 |
| Microsoft.NET.Test.Sdk | 17.10.0 |
| VSTest | 18.7.0 (x64) |
| Target framework | net8.0 |
| .NET SDK | 10.0.400 |
| OS | Windows 11 x64 |

### Summary

`Tokenizer` breaks a word token on whitespace, and `GetWordOrFqn` only absorbs a
`(...)` group when `(` *immediately* follows the word characters. A test name such
as `Fixture.Method (Case 1)` therefore tokenizes as two tokens instead of one, and
`TestFilterParser` throws.

VSTest discovers and filters the same name without complaint, so the test is
visible but unrunnable from the command line.

### Steps to reproduce

```csharp
using NUnit.Framework;

namespace Repro;

public class FilterRepro
{
    [TestCaseSource(nameof(Cases))]
    public void SpaceBeforeParenthesis(string value) => Assert.That(value, Is.Not.Null);

    private static IEnumerable<TestCaseData> Cases()
    {
        // Note the space before '(' - NUnit accepts it and Test Explorer displays it fine.
        yield return new TestCaseData("a").SetName("{m} (Case 1)");
    }
}
```

Fully qualified name:

```
Repro.FilterRepro.SpaceBeforeParenthesis (Case 1)
```

**Discovery matches:**

```powershell
PS > dotnet vstest Repro.dll /ListFullyQualifiedTests /ListTestsTargetPath:out.txt "/TestCaseFilter:FullyQualifiedName=Repro.FilterRepro.SpaceBeforeParenthesis \(Case 1\)" ; Get-Content out.txt  
VSTest version 18.7.0 (x64)

Repro.FilterRepro.SpaceBeforeParenthesis (Case 1)
```

**Execution throws:**

```powershell
PS > dotnet test Repro.dll --filter "FullyQualifiedName=Repro.FilterRepro.SpaceBeforeParenthesis \(Case 1\)"
VSTest version 18.7.0 (x64)

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
An exception occurred while invoking executor 'executor://nunit3testexecutor/': Unexpected FQN '\(Case 1\)' at position 60 in selection expression.
Stack trace:
   at NUnit.VisualStudio.TestAdapter.TestFilterConverter.TestFilterParser.Expect(Token[] valid) in D:\a\nunit3-vs-adapter\nunit3-vs-adapter\src\NUnitTestAdapter\TestFilterConverter\TestFilterParser.cs:line 260
   at ...
```

### Expected behavior

The a test that can be discovered can be run as it does when the same name has no space before `(`.

### Root cause

`src/NUnitTestAdapter/TestFilterConverter/Tokenizer.cs`:

```csharp
private const string WORD_BREAK_CHARS = "=~!()&|";

private bool IsWordChar(char c)
{
    if (char.IsWhiteSpace(c) || c == EOF_CHAR)
        return false;
    return WORD_BREAK_CHARS.IndexOf(c) < 0;
}

private Token GetWordOrFqn()
{
    var sb = new StringBuilder();
    int pos = index;

    CollectWordChars(sb);

    if (NextChar != '(')
        return new Token(TokenKind.Word, sb.ToString()) { Pos = pos };   // <-- must be immediate

    CollectBalancedParentheticalExpression(sb);
    ...
    return new Token(TokenKind.FQN, sb.ToString()) { Pos = pos };
}
```

For `...SpaceBeforeParenthesis \(Case 1\)`:

1. `CollectWordChars` stops at the space, and `NextChar` is `' '` rather than `'('`,
   so a **`Word`** token is returned.
2. `ParseFilterCondition` accepts that as the RHS and returns.
3. `ParseFilterExpression` then looks for `&`, `|`, `)` or EOF. `SkipBlanks` eats the
   space and `GetWordOrFqn` runs again, this time producing an **`FQN`** token
   `\(Case 1\)`.
4. `Expect(EOF)` fails → `Unexpected FQN '\(Case 1\)' at position 78`.

Note that everything *inside* the parentheses is already free-form — spaces, quotes,
commas and colons are all collected verbatim by `CollectBalancedParentheticalExpression`.
Only the gap between the name and the `(` is fatal, and there is no way to escape it:
`IsWordChar` rejects whitespace unconditionally.

### Impact

Any name produced by `TestCaseData.SetName` that separates the method name from the
argument text — e.g. `SetName("{m}" + $" ({description})")`, a fairly natural
formatting choice — becomes unrunnable via `dotnet test --filter` or
`RunConfiguration/TestCaseFilter`, while still appearing in Test Explorer and in
`--list-tests`.

PanzerFowst:
- GetWordOrFqn now peeks past whitespace between a name and its argument list instead of requiring '(' to immediately follow word characters, so names like "Method (Case 1)" tokenize as a single FQN instead of a Word followed by a stray FQN token.
- CollectBalancedParentheticalExpression now stops at EOF even when depth never reaches 0, preventing an infinite loop on a truncated filter with an unbalanced '('.
- Add regression tests for both: whitespace-before-'(' FQN parsing, and a bounded-timeout test proving unbalanced parens no longer hang.

Fixes #1490

### DIFF
--- a/src/NUnitTestAdapter/TestFilterConverter/Tokenizer.cs
+++ b/src/NUnitTestAdapter/TestFilterConverter/Tokenizer.cs
@@ -234,9 +234,18 @@ public class Tokenizer
 
         CollectWordChars(sb);
 
-        if (NextChar != '(')
-            return new Token(TokenKind.Word, sb.ToString()) { Pos = pos };
+        int preWhitespaceIndex = index;
+        var whitespace = new StringBuilder();
+        while (char.IsWhiteSpace(NextChar))
+            whitespace.Append(GetChar());
 
+        if (NextChar != '(')
+        {
+            index = preWhitespaceIndex;
+            return new Token(TokenKind.Word, sb.ToString()) { Pos = pos };
+        }
+
+        sb.Append(whitespace);
         CollectBalancedParentheticalExpression(sb);
 
         while (NextChar == '+' || NextChar == '.')
@@ -272,7 +281,7 @@ public class Tokenizer
                 else if (c == '"')
                     CollectQuotedString(sb);
             }
-            while (depth > 0);
+            while (depth > 0 && NextChar != EOF_CHAR);
         }
     }
 
@@ -309,4 +318,3 @@ public class Tokenizer
             index++;
     }
 }
-
--- a/src/NUnitTestAdapterTests/TestFilterConverterTests/TestFilterParserTests.cs
+++ b/src/NUnitTestAdapterTests/TestFilterConverterTests/TestFilterParserTests.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Threading.Tasks;
 using System.Xml;
 using NUnit.Framework;
 using NUnit.VisualStudio.TestAdapter.TestFilterConverter;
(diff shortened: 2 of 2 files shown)


OsirisTerje:
@PanzerFowst 

Conflicts needs to be resolved.

Change from Draft

PanzerFowst:
After reviewing more of the adapter filter parsing code, I don't think like my changes made in this PR as they don't address the full problem (which is not enforcing the escaped special characters in test names) as I have started discussing in #1501...

OsirisTerje:
@PanzerFowst You're right.

